PR #18, opened 2018-10-16
change the syntax of createSelector to better fit our unit testing st…

bthaile:
…rategy, will change unit tests in other PR

removed unneeded selectors and remove `export` on methods that aren't called externally. Change syntax of  createSelector to return a method then export const that calls that method.


justinbarry (approved):
LGTM

PR #27, opened 2018-10-23
Split constants from augur service

justinbarry:
Break constants export out to another file to allow us to use jest automocking.


antazoey (commented):
What about putting the them both in an augurjs directory so that we know that the constants are associated with augurjs?

bthaile (approved):
lgtm

bthaile:
splitting 'em out does that mean all repos would use the these constants? Like if augur-node needs a constant and augur-ui uses the same constant then they move to this constants file?

justinbarry:
@bthaile I left the original code in place. So both of the following are valid:
```javascript
import Augur from "augur.js";

const { constants } = Augur;
const { constants } = new Augur();
```

PR #84, opened 2018-11-20
Fix tests run from root

justinbarry:
The test now fail due to contract API changes but the failures are for real. We will need to discuss how to move forward.


bthaile (approved):
changes look good for running tests.

PR #89, opened 2018-11-20
Fix tests and align version numbers

justinbarry:


packages/augur-ui/jest.config.js
```diff
@@ -67,11 +67,7 @@ module.exports = {
 
   // A map from regular expressions to module names that allow to stub out resources with a single module
   moduleNameMapper: {
-    "^src/(.*)$": "<rootDir>/src/$1",
-    "^(config|modules|utils|services)/(.*)$": "<rootDir>/src/$1/$2",
-    "^assertions/(.*)$": "<rootDir>/test/assertions/$1",
-    "^test/(.*)$": "<rootDir>/test/$1",
-    "\\.(style|styles)$": "<rootDir>/test/mockStyle.js"
+    "\\.styles?$": "<rootDir>/test/mockStyle.js"
```
bthaile: I think there might be include files that end with `.style` would that be an issue?

packages/augur.js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "augur.js",
-  "version": "6.0.1",
+  "version": "6.0.0-0",
```
bthaile, line 3: I thought this got changed to 6.0.1 ?

packages/augur-ui/jest.config.js
justinbarry: That is just shorthand for both style and styles. It was done on another branch and shouldn't haven't show up here.

packages/augur.js/package.json
justinbarry: The version needs to match what is required in the other modules for proper linking.

packages/augur-ui/jest.config.js
justinbarry: Will rebase it out.

packages/augur.js/package.json
justinbarry: It breaks this: https://github.com/AugurProject/augur/pull/96#issuecomment-440855225

PR #164, opened 2018-11-29
Update augur-node

justinbarry:


nuevoalex (approved):
LGTM but I'm also just skimming since this is just a merge

PR #220, opened 2018-11-29
Remove code that was added upstream but unneeded here

justinbarry:


packages/augur-node/src/types.ts
```diff
@@ -48,7 +48,7 @@ export class ConnectOptions {
   ws?: string = "ws://localhost:8546";
   ipc?: string;
   propagationDelayWaitMillis?: number;
-  maxRetries?: number;
+  maxRetries?: number = 3;
```
justinbarry:
```suggestion
// maxRetries is the maximum number of retries for retryable Ethereum
// RPC requests. maxRetries is passed to augur.js's augur.connect() and
// then to ethrpc library.connect(), and is used internally by ethrpc
// for both HTTP and WS transports. When an ethrpc request errors, a
// subset of errors are statically configured as retryable, in which case
// ethrpc will opaquely re-insert the RPC request at its internal queue
// head, such that augur.js (and augur-node) are ignorant of requests
// that eventually succeed after N retries (where N < maxRetries).
  maxRetries?: number = 3;
```

PR #623, opened 2018-12-05
fix merge conflicts

bthaile:
resolve conflicts.

We need to port back all test and code changes from monorepo/packages/augur-ui to augur-ui


justinbarry (approved):
This *looks* good but the upstream contains changes that need to come in.

PR #628, opened 2018-12-05
Update flash timestamp

bthaile:
Get flash working to push timestamp


justinbarry:
@bthaile Can you add example usage/docs to docker.md?

phoebemirman (approved):

PR #634, opened 2018-12-05
Juliya/init augur tests false pos

antazoey:


bthaile:
We'll bring over the update jest tests from augur-ui v1

unparalleled-js:
the one from

PR #647, opened 2018-12-08
Create bug report template

adrake33:
This will create a new button on the Issues page for filing a bug report and should help ensure we have more relevant info when a bug gets filed.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,48 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: Bug
+assignees: adrake33
+
+---
+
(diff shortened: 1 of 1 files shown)


bthaile (commented):
Mention iser has to login to see version information. 

Thoughts on button for feature request.

adrake33:
Ah good point about having to be logged in. Yeah, I was also planning to create one for Feature Requests but hadn't gotten around to it yet.

bthaile (approved):
Lgtm

adrake33:
Actually, since we're using ZenHub now, I don't think there's really a need to redirect all v1 tickets to the augur repo anymore. ZenHub can pull in issues from multiple repos, so we should use this template for v1 bugs and create a different template for v2 bugs later. Will close this for now and move it to the augur-app repo.

PR #655, opened 2018-12-11
Achapman/ch17993/reduce or eliminate the use of the controlled

nuevoalex:
A rather large PR that removes the Controller contract as well as the Controlled contract.

Rather than using a whitelist we're just looking for specific callers in most cases now. The registry component was moved to the Augur contract.


packages/augur-core/source/contracts/LegacyReputationToken.sol
```diff
@@ -8,7 +8,7 @@ contract LegacyReputationToken is OldLegacyReputationToken {
     using ContractExists for address;
     event FundedAccount(address indexed _universe, address indexed _sender, uint256 _repBalance, uint256 _timestamp);
 
-    address private constant FOUNDATION_REP_ADDRESS = address(0x1985365e9f78359a9B6AD760e32412f4a445E862);
+    address private constant FOUNDATION_REP_ADDRESS = address(0xE94327D07Fc17907b4DB788E5aDf2ed424adDff6);
```
epheph: Why change this?

nuevoalex: Hm yea actually this isn't really needed. Will revert

packages/augur-core/source/contracts/Augur.sol
```diff
@@ -60,10 +60,48 @@ contract Augur is Controlled, IAugur {
...
+            time = ITime(_address);
+        }
+        return true;
+    }
+
+    function lookup(bytes32 _key) public view returns (address) {
+        return registry[_key];
+    }
+
+    function finishDeployment() public returns (bool) {
```
epheph, line 101: Anyone can call this, potentially breaking an upload process before it is finished?

nuevoalex: fair point! Will restrict

epheph (approved):
🦉

justinbarry (approved):
LGTM

PR #657, opened 2018-12-11
update the docker build all script to use yarn workspaces, update bui…

bthaile:
…ld.sh to get newest image_id just in case there are multiple versions, default docker.json content just in case it is cleared out


nuevoalex (approved):
yay!

PR #707, opened 2018-12-14
update flash scripts to pass in new flag -asPrice

bthaile:
reused branch since I had it around.

`asPrice` flag allows the user to pass in outcome as price instead of index. For scalar markets if 0 is passed in it will be interpreted as outcome 0 (invalid) if the -asPrice flag is used then outcome of 0 will be interpreted as price 0.


justinbarry (approved):
LGTM

PR #767, opened 2018-12-31
Treat Cash as standard ERC20, remove market creator mailbox

epheph:
Lots of test fixes, a few solidity changes. ignoreShares behavior was changed.


nuevoalex (approved):
🐘

PR #774, opened 2019-01-02
changes augur-core CI so that all tests are run during build time

petong:


epheph (approved):
Works great. For posterity, the `set -x` was causing the `##vso` raw command to be output AND the actual command output to be output. Depending on the order these were seen by the azure service, it would either have seen this:
```
+ echo "##vso[task.setVariable variable=AUGUR_CORE]true"
```
or this
```
##vso[task.setVariable variable=AUGUR_CORE]true
```
last, meaning the value would either be `true"` or `true`. And `true"` != `true`. All comes down to stderr/stdout buffering.

PR #828, opened 2019-01-11
Erc777

nuevoalex:
Now with additional spec workarounds for our internal usage and for external users.

Note also we already disallow a market creator or designate reporter of address(0) and ownership cannot be transferred to address(0)


packages/augur-core/source/contracts/libraries/token/ERC777BaseToken.sol
```diff
@@ -0,0 +1,140 @@
...
+        if (_callHooks) {
+            callRecipient(_operator, _from, _to, _amount, _data, _operatorData, _preventLocking);
+        }
+
+        emit Sent(_operator, _from, _to, _amount, _data, _operatorData);
+        return true;
+    }
+
+    function doBurn(address _operator, address _tokenHolder, uint256 _amount, bytes32 _data, bytes32 _operatorData) internal returns (bool) {
+        callSender(_operator, _tokenHolder, 0x0, _amount, _data, _operatorData);
```
epheph, line 111: Will we need a burnNoHooks took too? I don't think we do, because you are never burning another user's tokens in the context of your own transaction, just wanted to bring it up, though

nuevoalex: Yea, shouldn't be needed. Technically I think even for the sendNoHooks it _could_ just skip the receiver hook for our case but that seems really weird as opposed to just not calling the 820 methods at all

epheph: I believe that is true; only receive hooks are a vulnerability, but the way it is done looks good, not calling either.

packages/augur-core/source/contracts/legacy_reputation/OldLegacyRepToken.sol
```diff
@@ -108,27 +108,32 @@ contract OldLegacyReputationToken is DelegationTarget, ITyped, Initializable, Va
...
     function trustedUniverseTransfer(address _source, address _destination, uint256 _attotokens) public whenNotMigratingFromLegacy afterInitialized returns (bool) {
         require(IUniverse(msg.sender) == universe);
-        return internalTransfer(_source, _destination, _attotokens);
+        return internalTransfer(_source, _destination, _attotokens, true);
     }
 
     function trustedMarketTransfer(address _source, address _destination, uint256 _attotokens) public whenNotMigratingFromLegacy afterInitialized returns (bool) {
         require(universe.isContainerForMarket(IMarket(msg.sender)));
-        return internalTransfer(_source, _destination, _attotokens);
+        return internalTransfer(_source, _destination, _attotokens, true);
```
epheph, line 126: Could probably save gas to make this false? But there are cases where the owner might want to be notified their tokens are being sent AND the owner is the one originating the tx so it isn't a risk to others? This implementation is probably the most correct, but does cost extra gas.

No change necessary, just noting it.

packages/augur-core/source/contracts/libraries/token/StandardToken.sol
```diff
@@ -85,9 +60,29 @@ contract StandardToken is ERC20, BasicToken {
         return true;
     }
 
-    function approveInternal(address _owner, address _spender, uint256 _value) internal returns (bool) {
-        allowed[_owner][_spender] = _value;
-        emit Approval(_owner, _spender, _value);
+    function approveInternal(address _owner, address _spender, uint256 _allowance) public returns (bool) {
```
epheph: !! This got set to public!

nuevoalex: 😨 nice catch

epheph (approved):
Looks good. I'd be curious of gas efficiency, but i imagine it's balanced out a bit by better delegation/no delegation?

pgebheim:
@epheph --- Why the conflicts?

pgebheim:
@epheph -- So many conflicts :-(

epheph:
I think we punted on actually merging this, right @nuevoalex ? Too many risks? I can't remember if we decided to punt or to move forward with the internal-dont-perform-callbacks

pgebheim:
Oh I see, gonna add [WIP] then,

epheph:
Merge conflict resolved, ready to merge?

PR #907, opened 2019-01-29
adds missing packages to augur-tools

petong:


pgebheim:
@justinbarry -- I think these deps would have come from the old augur-core module... I have a feeling augur-tools will need to be re-worked with the new contract interfaces once we get alex's patch in

pgebheim (approved):
I do think this is right!

@justinbarry Adding you as reviewer too in case you wanted to handle it differently.

justinbarry (approved):
LGTM

PR #922, opened 2019-01-30
Fixes missing types

petong:
a previous commit 9f95ca94d2f6b4bf234fc152df2bd2b457f44487 deleted these types which caused augur-tools and augur-node to stop working. Adding back for now.


pgebheim (approved):
Ohh ok. So just restoring these ABIs!

This may work.... does augur-core of ethjs* as a devDep or something so its installed so the tool can use it?

PR #949, opened 2019-02-02
Blockstream test tool

epheph:
Lots of improvements possible, but want to make sure we use it before putting more engineering effort into it.


epheph:
the `ethers` adapter will not work properly un until https://github.com/ethers-io/ethers.js/pull/412 merged.

It __seems__ like it works, but only incidentally, if you happen to be query logs for the block that is latest. ethers currently discards blockHash in the getLogs request, re-targetting the log query from the requested block to the latest block

justinbarry (approved):
LGTM

Issue #756, opened 2018-12-20
Spike: Create PouchDB schemas


adrake33:

adrake33:
Closing. This was merged into the Augur-668 branch, which is the main branch for Augur State changes until everything is ready to be merged.

PR #979, opened 2019-02-06
Initialize PouchDB schema

adrake33:
Fixes https://github.com/AugurProject/augur/issues/756.

Makes the following changes:
- Adds the ability to specify a networkId & tracked users when creating/initializing DB
- Creates SyncableDBs for generic events (name format: networkId-eventName)
- Creates UserSyncableDBs for user-specific events (name format: networkId-eventName-userAddress)
- Creates MetaDB for associating a block number with sequenceIds for each Syncable/UserSyncableDB. (Currently, the format for the _id field is `highestSyncedBlockNumber-timestamp`, but I may break out the block number into a separate field as I work on the rollback functionality to make querying simpler.)


packages/augur-state/src/db/DB.ts
```diff
@@ -2,50 +2,219 @@ import { SyncableDB } from './SyncableDB';
...
+  name: string;
+  numAdditionalTopics: number;
+  userTopicIndex: number;
+}
 
-  public syncStatus: SyncStatus;
-  public trackedUsers: TrackedUsers;
-  public marketCreated: SyncableDB<TBigNumber>;
-  public marketFinalized: SyncableDB<TBigNumber>;
+const genericEventNames: Array<string> = [
```
pgebheim, line 16: I wonder if we can get this from the generated contract interfaces so we don't need to duplicate it?

pgebheim (approved):

packages/augur-state/src/db/DB.ts
adrake33: I was initially doing that but then realized we only needed to get the ones that were generic events. We could add a property to the ones in that list to denote which ones are generic, though. I'll do that before merging this.

adrake33: Actually I'll just add a TODO for this and add it with the rollback PR.

PR #990, opened 2019-02-09
Upgrade ethers.js to version that supports getLogs by blockHash

epheph:


epheph:
re: https://github.com/ethers-io/ethers.js/pull/412

pgebheim (approved):

Issue #952, opened 2019-02-02
AN in browser MVP: Add rollback functionality


adrake33:
Once all event types are synced to their respective PouchDB documents, we need to be able to handle reorgs by being able to revert to a previous revision of the document.

PR #996, opened 2019-02-11
Add rollback functionality to Augur State

adrake33:
Fixes https://github.com/AugurProject/augur/issues/952. Also updates ethers version to 4.0.24.

### DIFF
--- a/packages/augur-state/src/Controller.ts
+++ b/packages/augur-state/src/Controller.ts
@@ -1,22 +1,23 @@
 import { DB } from './db/DB';
 import { Augur } from 'augur-api';
-const uploadBlockNumbers = require('augur-artifacts/upload-block-numbers.json');
 
 export class Controller<TBigNumber> {
   private dbController: DB<TBigNumber>;
   private augur: Augur<TBigNumber>;
   private networkId: number;
+  private defaultStartSyncBlockNumber: number;
   private trackedUsers: Array<string>;
 
-  public constructor (augur: Augur<TBigNumber>, networkId: number, trackedUsers: Array<string>) {
+  public constructor (augur: Augur<TBigNumber>, networkId: number, defaultStartSyncBlockNumber: number, trackedUsers: Array<string>) {
     this.augur = augur;
     this.networkId = networkId;
+    this.defaultStartSyncBlockNumber;
     this.trackedUsers = trackedUsers;
   }
 
   public async run(): Promise<void> {
-    this.dbController = await DB.createAndInitializeDB(this.networkId, this.trackedUsers);
-    await this.dbController.sync(this.augur, 100000, 5, uploadBlockNumbers[this.networkId]);
+    this.dbController = await DB.createAndInitializeDB(this.networkId, this.defaultStartSyncBlockNumber, this.trackedUsers);
+    await this.dbController.sync(this.augur, 100000, 5);
 
     // TODO Move this function into separate test
     await this.testRollback();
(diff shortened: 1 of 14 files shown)


packages/augur-state/src/Controller.ts
```diff
@@ -1,5 +1,6 @@
 import { DB } from './db/DB';
 import { Augur } from 'augur-api';
+const uploadBlockNumbers = require('augur-artifacts/upload-block-numbers.json');
```
robert-e-davidson3: What are upload block numbers? I'm worried that this will make testing difficult unless it's added to the Controller class via its constructor.

adrake33: These are the block numbers on which the Augur contracts were uploaded. We start scanning the blockchain for event logs at this block number because there cannot be any that exist before that block.

packages/augur-state/src/db/DB.ts
```diff
@@ -146,76 +150,177 @@ export class DB<TBigNumber> {
...
+   * @param {number} chunkSize Number of blocks to retrieve at a time when syncing logs
+   * @param {number} blockstreamDelay Number of blocks by which blockstream is behind the blockchain
+   * @param {number} uploadBlockNumber Block number at which Augur contracts were originally uploaded to the blockchain
    */
   public async sync(augur: Augur<TBigNumber>, chunkSize: number, blockstreamDelay: number, uploadBlockNumber: number): Promise<void> {
-    let sequenceIds: { [eventName: string]: string } = {};
+    // Sync generic event types & user-specific event types
+    let dbSyncPromises = [];
+    for (let eventName of genericEventNames) {
+      let dbName = `${this.networkId}-${eventName}`;
```
robert-e-davidson3: I recommend creating a separate function for creating these kinds of values, to minimize the risk of later typos causing confusing bugs.

pgebheim:
This is the first block that contains our contracts, used to help know where to start scanning.

Might be nice to dependency inject that so you don’t need to rely on a require / proxyquire to get them

Paul Gebheim
paul@augur.net

packages/augur-state/src/db/DB.ts
adrake33: Yeah, good idea. I'll add that.

packages/augur-state/src/Controller.ts
robert-e-davidson3: Ah makes sense. I see that there are network IDs that start at 0, 1, and some other low values so this won't affect testing 👍.

adrake33:
@pgebheim Sure, I can make the uploadBlockNumbers a param that gets passed into the DBController.

packages/augur-state/src/db/DB.ts
```diff
@@ -146,76 +150,177 @@ export class DB<TBigNumber> {
...
-          sequenceIds[`${this.networkId}-${userSpecificEvents[eventIndex].name}-${trackedUser}`] = updateSeq;
+        let dbName = `${this.networkId}-${userSpecificEvents[eventIndex].name}-${trackedUser}`;
+        let previousSequenceId = parseBlockSequenceIds[dbName];
+        let dbInfo = await this.userSyncableDatabases[dbName].getInfo();
+        if (dbInfo.update_seq > previousSequenceId) {
+          dbRollbackPromises.push(this.userSyncableDatabases[dbName].rollback(blockNumber, previousSequenceId));
         }
       }
     }
+    await Promise.all(dbRollbackPromises);
```
justinbarry: What happens if any given `dbRollbackPromises` fails? Will it put us in a bad state?

pgebheim: That's a good question... we should probably fail and see what failed?

adrake33: Agreed. So if any of these promises gets rejected, we should output the failure and stop running at that point? Or do we want to re-try the rollback a few times?

PR #1005, opened 2019-02-12
Generate mock solidity contracts for use in unit tests

robert-e-davidson3:
Generates mock solidity contracts for use in unit tests.
Also some minor miscellaneous changes.
Note that the unit tests do not yet work because contract registration needs to be overhauled.


pgebheim:
Adding this to Belligerent Banana 'cause I think its state needs to be handled.

robert-e-davidson3:
Split off the remaining work into #1119.

nuevoalex (approved):
LGTM for a start!

PR #1027, opened 2019-02-12
WIP: Update augur-core to use types added to TS generator

justinbarry:


justinbarry:
We can use this as a base for the linked task.

PR #1073, opened 2019-02-18
Tests use in-memory adapter for PouchDB. Can now mock PouchDB.

robert-e-davidson3:
Added a test in `DB.test.ts` that causes the PouchDB database to throw an error.

FYI: There are a smattering of edits that my IDE made. AFAICT it's removing trailing whitespace.


packages/augur-state/src/db/DB.test.ts
```diff
@@ -141,7 +141,7 @@ test("database failure during trackedUsers.getUsers() call", async () => {
 
     const db = await DB.createAndInitializeDB(
       TEST_NETWORK_ID,
-      BLOCKSTREAM_DELAY,
```
robert-e-davidson3: Should also delete this variable's definition earlier in the file.

adrake33 (approved):
Adding settings.json file. The other changes look good to me.

PR #1102, opened 2019-02-21
Rollback test

adrake33:
Creates a test for adding blocks & performing a rollback. Also moves makeMock function to a separate file under the `utils` directory.


robert-e-davidson3 (approved):
Looks good to me.

PR #1103, opened 2019-02-21
DB sync testing

robert-e-davidson3:


robert-e-davidson3:
I marked only the one test as runnable because the tests interfere with each other, likely because `wipeDB` only wipes the database I'm aware of.

It looks like if the first sync is successful then all subsequent syncs are successful, unless specifically forced to fail with `mock.failNext()` and similar.

However if the first sync fails then all subsequent syncs fail.

robert-e-davidson3:
`wipeDB` now actually works.

`DB.sync` calls don't get into a bad state. We still don't know why they get into a bad state when only one database call fails but a more likely situation is _all_ database calls failing for a little while so that's now what the tests do.

adrake33 (approved):
LGTM

PR #1115, opened 2019-02-23
using common mock for DB.test.ts

robert-e-davidson3:
Also fixes DB.test.ts since one of the imports was messed up.


adrake33 (approved):
👍

PR #1237, opened 2019-03-05
compile and deploy Augur contracts to ganache node running in the tests

robert-e-davidson3:


packages/ethersjs-provider/src/EthersJsonRpcProvider.ts
```diff
@@ -0,0 +1,61 @@
+import { NetworkId } from "@augurproject/artifacts";
```
robert-e-davidson3, line 1: This file won't be committed. It's temporary to make the rest of the code work.

packages/ethersjs-provider/src/EthersWeb3Provider.ts
```diff
@@ -0,0 +1,61 @@
+import { NetworkId } from "@augurproject/artifacts";
```
robert-e-davidson3, line 1: This file won't be committed. It's temporary to make the rest of the code work.

packages/ethersjs-provider/src/index.ts
```diff
@@ -1 +1,2 @@
-export { EthersProvider } from "./EthersProvider";
```
robert-e-davidson3, line 1: This file won't be committed. It's temporary to make the rest of the code work.

packages/augur-state/src/db/DB.test.ts
```diff
@@ -1,103 +1,25 @@
 import { TrackedUsers } from "./TrackedUsers";
 import { DB, UserSpecificEvent } from "./DB";
-import { EthersProvider } from "ethers-provider";
+import { EthersWeb3Provider } from "@augurproject/ethers-provider";
```
robert-e-davidson3, line 3: This line will probably be different as it's based on temporary code I used while waiting for the proper implementation to be completed. I'm not sure which ticket that has though. @adrake33?

adrake33: Yeah, this line should be `import { EthersProvider, Web3AsyncSendable } from "ethers-provider";`

PR #1252, opened 2019-03-06
Make bulk sync use async queue

### LINKED ISSUE #1219: Create web3 wrapper
We need to create a web3 wrapper class in Augur State that will asynchonously queue connections when syncing and allow a limit of simultaneous connections to be specified.

adrake33:
Fixes https://github.com/AugurProject/augur/issues/1219. Jest tests will be broken due to document update conflict & timeout errors (which will be fixed in a later PR once the bulk sync has been updated to fetch multiple events in a single call).


packages/augur-state/src/db/AbstractDB.ts
```diff
@@ -86,5 +87,9 @@ export abstract class AbstractDB {
 
 export type PouchDBFactoryType = (dbName: string) => PouchDB.Database;
 export function PouchDBFactory(dbArgs : object) {
-  return (dbName: string) => new PouchDB(`db/${dbName}`, dbArgs);
+  const dbDir = "db";
+  if (!fs.existsSync(dbDir)){
```
robert-e-davidson3, line 91: This is unnecessary if pouchdb is using the memory adapter.

packages/augur-state/src/db/AbstractDB.ts
```diff
@@ -86,5 +87,9 @@ export abstract class AbstractDB {
...
+    fs.mkdirSync(dbDir);
+  }
+  return (dbName: string) => new PouchDB(dbDir + `/${dbName}`, dbArgs);
```
robert-e-davidson3:
```suggestion
  return (dbName: string) => new PouchDB(`${dbDir}/${dbName}`, dbArgs);
```

packages/ethersjs-provider/src/EthersProvider.ts
```diff
@@ -5,14 +5,17 @@ import { EthersProvider as EProvider } from "contract-dependencies-ethers";
...
 
 interface ContractMapping {
     [contractName: string]: ethers.utils.Interface;
 }
 
-export class EthersProvider extends ethers.providers.JsonRpcProvider implements Provider, EProvider {
+export class EthersProvider extends ethers.providers.Web3Provider implements Provider, EProvider {
     private contractMapping: ContractMapping = {};
 
+    constructor(web3AsyncSendable: Web3AsyncSendable) { super(web3AsyncSendable); }
```
robert-e-davidson3: Is this redundant?

packages/ethersjs-provider/src/Web3AsyncSendable.ts
```diff
@@ -0,0 +1,38 @@
...
+    private asyncQueue: any;
+
+    constructor(
+        private ethNodeUrl: string,
+        times: number, 
+        interval: number, 
+        concurrency: number
+    ) 
+    {
+        const send: any = function(task: any, callback: any) {
```
robert-e-davidson3: Should this callback have the same type as the one in the public send's parameters?

packages/augur-state/src/db/AbstractDB.ts
adrake33: I was unable to run `yarn start` if the db directory didn't exist. I had to create the db directory, so I figured I'd just add this as a precaution.

packages/ethersjs-provider/src/EthersProvider.ts
adrake33: Oops, yeah. This constructor was originally different, but got changed later. I'll remove it.

packages/augur-state/src/db/AbstractDB.ts
robert-e-davidson3: Submitted https://github.com/AugurProject/augur/issues/1254 to fix this later since it's an unknown amount of work, low priority, and this PR is high priority.

packages/ethersjs-provider/src/Web3AsyncSendable.ts
adrake33: Yep! Will update.

robert-e-davidson3 (approved):

Issue #1259, opened 2019-03-06
Make bulk sync more efficient


adrake33:
~~We should be making single calls to the Ethereum node that return logs for all event types, rather than multiple calls for each event type. This will require changing the sync logic as well as the logic for parsing the logs and sending the data to each DB.~~

Move async queue to EthersProvider.

pgebheim:
@adrake33 -- Is this issue still supposed to be in Review/QA?

adrake33:
@pg I created a PR that gets multiple events in a single request for each chunk. However, Alchemy takes a significantly long time to send a response when requesting multiple events if the number of results is high (e.g. 1500+). I'd also like to get the async queue issue figure out before attempting to merge this since the tests take a long time to run. Will move this to the next sprint.

PR #1282, opened 2019-03-09
Optimize log fetching

adrake33:
Fixes https://github.com/AugurProject/augur/issues/1259. Also does some minor code cleanup. For the time being, I've bumped the timeout duration of the tests so that they'll pass. Hopefully we can work with Alchemy to reduce these times.


packages/augur-api/src/api/Events.ts
```diff
@@ -5,39 +5,48 @@ import {abi} from "@augurproject/artifacts";
...
+    let topics: Array<string | Array<string>> = [events];
+    if (additionalTopics) {
+      topics = topics.concat(additionalTopics);
     }
+
+    return this.provider.getLogs({fromBlock, toBlock, topics, address: this.augurAddress});
+  }
+
+  public parseLogs(logs: Array<Log>): Array<ParsedLog> {
+    return _.map(logs, (log) => {
```
brianosaurus, line 36: I believe the TypeScript Array type has a map method.

packages/augur-state/src/Controller.ts
```diff
@@ -1,131 +1,28 @@
...
+import settings from "./settings.json";
+import { PouchDBFactory, PouchDBFactoryType } from "./db/AbstractDB";
+import { DB } from "./db/DB";
+import FlexSearch = require("flexsearch"); // because flexsearch is a UMD type lib
 
 // Need this interface to access these items on the documents in a SyncableDB
 interface SyncableMarketDataDoc extends PouchDB.Core.ExistingDocument<PouchDB.Core.AllDocsMeta> {
-  extraInfo: string
-  description: string
+  extraInfo: string;
```
brianosaurus, line 9: Thanks! I didn't have my tslinter on yet when I wrote that.

packages/augur-state/src/Controller.ts
```diff
@@ -135,26 +32,26 @@ export class Controller<TBigNumber> {
...
-        this.augur, 
-        settings.chunkSize, 
-        settings.blockstreamDelay
+        this.augur.provider,
+        this.augur.events,
+        settings.chunkSize
       );
 
-      const previousDocumentEntries = await this.dbController.getSyncableDatabase(this.dbController.getDatabaseName("MarketCreated")).allDocs();
+      const previousDocumentEntries = await this.dbController.getSyncableDatabase("MarketCreated").allDocs();
```
brianosaurus, line 54: Thanks again.

packages/augur-state/src/db/DB.ts
```diff
@@ -103,43 +120,51 @@ export class DB<TBigNumber> {
...
-        )
-      );
-    }
+  public async sync(ethereumProvider: Provider, events: Events, chunkSize: number): Promise<void> {
+    const startSyncBlockNumber = await this.getSyncStartingBlock();
+    const endSyncBlockNumber = await ethereumProvider.getBlockNumber() - this.blockstreamDelay;
+    let firstBlockNumberInChunk = startSyncBlockNumber;
 
-    for (let trackedUser of await this.trackedUsers.getUsers()) {
+    while (firstBlockNumberInChunk < endSyncBlockNumber) {
```
brianosaurus, line 134: Should that be <= ?

packages/augur-state/src/db/DB.ts
```diff
@@ -103,43 +120,51 @@ export class DB<TBigNumber> {
...
+      const lastBlockNumberInChunk = Math.min(firstBlockNumberInChunk + chunkSize, endSyncBlockNumber);
+      const rawLogsPromises:Array<Promise<Array<Log>>> = [];
+
+      // Get all generic event logs in chunk
+      console.log(`SYNCING all generic events from blocks ${firstBlockNumberInChunk} to ${lastBlockNumberInChunk}`);
```
brianosaurus, line 139: Do we have a debug log?

packages/augur-state/src/db/DB.ts
```diff
@@ -103,43 +120,51 @@ export class DB<TBigNumber> {
...
-            blockstreamDelay,
-            highestAvailableBlockNumber
-          )
-        );
-      }
-    }
+        const additionalTopics: Array<string | Array<string>> = [];
+        additionalTopics.fill("", userSpecificEvent.numAdditionalTopics);
+        const trackedUsers = await this.trackedUsers.getUsers();
+        additionalTopics[userSpecificEvent.userTopicIndex] = trackedUsers.map(trackedUser => `0x000000000000000000000000${trackedUser.substr(2)}`);
```
brianosaurus, line 147: Feels like 0x00000000000000000000000 should be in a method or as a constant or something ... to prevent mistakes.

packages/augur-state/src/db/DB.ts
```diff
@@ -288,13 +294,79 @@ export class DB<TBigNumber> {
...
+      saveLogsPromises.push(this.getSyncableDatabase(genericEventName).saveLogs(parsedFilteredLogs, lastBlockNumberInChunk));
+      console.log(`Saved ${filteredLogs.length} ${genericEventName} log(s)`);
+    }
+
+    // Save user-specific events to their respective DBs
+    const trackedUsers = await this.trackedUsers.getUsers();
+    const encodedTrackedUsers = trackedUsers.map(trackedUser => `0x000000000000000000000000${trackedUser.substr(2)}`)
+    for (let userSpecificEvent of this.userSpecificEvents) {
+      const eventFilteredLogs = rawLogs.filter((log) => log.topics.includes(ethereumProvider.getEventTopic("Augur", userSpecificEvent.name)));
+        for (let trackedUserIndex = 0; trackedUserIndex < trackedUsers.length; trackedUserIndex++) {
```
brianosaurus, line 357: I recall tslint complaining about for(;;;) loops in favor of for (let i in list) which assigns I to the index.

packages/augur-state/src/db/SyncableDB.ts
```diff
@@ -1,100 +1,84 @@
...
+        sort: [{blockNumber: 'desc'}],
+      });
+      if (blocksToRemove.docs.length > 0) {
+        console.log("\n\nDeleting the following blocks from " + this.dbName);
+        console.log(blocksToRemove.docs);
+        for (let doc of blocksToRemove.docs) {
+          // Remove block number from event DB
+          await this.db.remove(doc._id, doc._rev);
+          // Update highest sync block with decremented block number
+          await this.syncStatus.setHighestSyncBlock(this.dbName, --highestSyncBlock);
```
brianosaurus, line 68: Is blocksToRemove always going to have a block listed even if a certain block had no data? I'm unclear on what db.find returns with the selector and fields. This question is mainly for my own edification.

brianosaurus (approved):
PR looks good. I have a bunch of questions but that's because I'm mainly curious about this or that.

I like the idea of async sending requests for logs about various events. Smart and must be a lot faster.

packages/augur-api/src/api/Events.ts
adrake33: Oh, you're right! We're using use lodash's map function in a few other places in the augur-state code, and we don't need to be. Will change this.

packages/augur-state/src/db/DB.ts
adrake33: Yeah, good catch. I think it should be. Will change it.

packages/augur-state/src/db/DB.ts
adrake33: Agreed, there are a few other places in the code where we're doing this, so it'd probably be worth turning this into a function. I'll add that.

packages/augur-state/src/db/SyncableDB.ts
adrake33: Ah, good point. We should still be decrementing the highest sync block even if db.find doesn't return any results. There could certainly be cases where the query doesn't return any documents, but we still need to decrement (in case of a re-org).

adrake33:
Closing since https://github.com/AugurProject/augur/pull/1296 uses the sync method we want to use. I've also copied over the requested changes to that PR.

PR #1320, opened 2019-03-14
Re-add queue

adrake33:
Re-adds queue to EthersProvider. (This was removed due to a misunderstanding on my part when talking to Paul.)


pgebheim (approved):
YES!

This looks exactly like I was thinking

PR #1387, opened 2019-03-19
Created augur-test and expanded augur-api's contract offerings

robert-e-davidson3:


jest.config.js
```diff
@@ -1,4 +1,8 @@
 module.exports = {
   reporters: ["default", "jest-junit"],
-  projects: ["<rootDir>/packages/augur-node", "<rootDir>/packages/augur-state", "<rootDir>/packages/augur-ui"]
+  projects: [
+    "<rootDir>/packages/augur-state",
+    // "<rootDir>/packages/augur-ui",
```
bthaile, line 5: this works for me

packages/augur-api/src/api/Contracts.ts
```diff
@@ -1,12 +1,29 @@
...
+  private readonly dependencies: GenericAugurInterfaces.Dependencies<TBigNumber>;
 
   public constructor (addresses: ContractAddresses, dependencies: GenericAugurInterfaces.Dependencies<TBigNumber>) {
+    this.dependencies = dependencies;
     this.augur = new GenericAugurInterfaces.Augur<TBigNumber>(dependencies, addresses.Augur);
     this.universe = new GenericAugurInterfaces.Universe<TBigNumber>(dependencies, addresses.Universe);
+    this.cash = new GenericAugurInterfaces.Cash<TBigNumber>(dependencies, addresses.Cash);
+  }
+
+  public async setReputationToken(testnet: boolean = false) {
```
robert-e-davidson3: I don't call this in the `Augur.ts` file's `create(...)` method. I'm not sure what the best approach would be:
1. Always create the production rep token but call `setReputationToken(true)` later for tests or
2. Add `testnet: bool` argument to `Augur.create(...)`.

packages/augur-state/tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "baseUrl": "./"
   },
   "exclude": [
-    "./src/**/*.test.ts"
+//    "./src/**/*.test.ts"
```
adrake33: Can this be removed?

packages/augur-test/libs/LocalAugur.ts
```diff
@@ -37,7 +37,14 @@ function makeDeployerConfiguration() {
   return new DeployerConfiguration(contractInputRoot, artifactOutputRoot, augurAddress, createGenesisUniverse, isProduction, useNormalTime, legacyRepAddress);
 }
 
-export async function makeTestAugur(accounts: AccountList): Promise<Augur<any>> {
+interface UsefulContractObjects {
+  provider: EthersProvider;
+  signer: EthersFastSubmitWallet;
+  dependencies: ContractDependenciesEthers;
+  compiledContracts: CompilerOutput;
+  addresses: any;
```
adrake33: Shouldn't this be changed to match the type returned by `ContractDeployer.deploy`?

packages/augur-api/src/api/Contracts.ts
nuevoalex: You can base this off of the network id. Any network id other than 1 is a testnet.

For the mainnet test deployment we actually wont have the faucet, but the REP contract will point to a fake, previously deployed legacy REP contract so this will work in that case as well.

adrake33 (approved):
LGTM

packages/augur-api/src/api/Contracts.ts
robert-e-davidson3: Thanks!

packages/augur-state/tsconfig.json
robert-e-davidson3: I think doing so will add the test files to the final build. I'm not actually sure though.

Justin did some changes to the tsconfig files that I should make sure I don't mess up here.

packages/augur-test/libs/LocalAugur.ts
robert-e-davidson3: Yeah it should.

jest.config.js
justinbarry: FWIW, these tests will be back online after I get the UI building properly in the monorepo.

packages/augur-state/tsconfig.json
justinbarry: This change will cause a merge conflict but we can deal with that when it happens.

robert-e-davidson3: OH. I didn't mean to include this change. I made it so that my IDE would behave correctly. I can undo this.

PR #1426, opened 2019-03-21
changes ADD to COPY so we cache properly

petong:


pgebheim (approved):
Oh my... That is a hard one to see

petong:
going to comment out the dev-pop-geth image job until it is fixed

PR #1450, opened 2019-03-22
remove paranoia code. Easily confirmable post launch (and needs to be)

nuevoalex:
This is all we really need to do to support mainnet testing as it turns out.


adrake33 (approved):
Lgtm

PR #1454, opened 2019-03-22
fixes augur-build to not use multistage build

petong:
speeds up augur-build creation.

fixes node-gyp dir typo


justinbarry (approved):
LGTM

PR #1480, opened 2019-03-26
Various pre audit issues

nuevoalex:
eth -> cash (dai but really a stand in for nay stablecoin so calling it cash)
use the correct bond in market creation
add convenience function for turning in multiple market winning shares at once


adrake33 (approved):
💵

PR #1566, opened 2019-04-03
1541 augur state in browser

### LINKED ISSUE #1541: Make augur-state run in the browser
Use the webpack-dev-server to get things running in a browser.

brianosaurus:
Changes to allow augur-state to be runnable in a browser.

Uses webpack-dev-server.

Closes #1541


packages/augur-state/package.json
```diff
@@ -34,20 +34,24 @@
...
     "lodash": "4.17.11",
     "postinstall-build": "5.0.3",
-    "pouchdb": "7.0.0",
-    "pouchdb-find": "7.0.0"
+    "pouchdb": "^7.0.0",
+    "pouchdb-find": "7.0.0",
+    "ts-loader": "^5.3.3",
+    "webpack": "^4.29.6",
+    "webpack-cli": "^3.3.0",
+    "yarn": "^1.15.2"
```
adrake33: Recommend removing the "^" from these. We've had package issues in the past that were caused by not using very specific version numbers, so we've been trying to avoid using the "^" notation in the package.json files.

packages/augur-state/package.json
```diff
@@ -56,6 +60,7 @@
     "pouchdb-adapter-memory": "7.0.0",
     "ts-jest": "23.10.5",
     "ts-node": "7.0.1",
-    "tsconfig-paths": "3.7.0"
+    "tsconfig-paths": "3.7.0",
+    "webpack-dev-server": "^3.2.1"
```
adrake33: Ditto.

packages/augur-ui/package.json
```diff
@@ -101,7 +101,8 @@
     "shelljs": "0.8.3",
     "shifty": "2.6.1",
     "speedomatic": "2.1.4",
-    "trezor-connect": "5.0.34"
+    "trezor-connect": "5.0.34",
+    "yarn": "^1.15.2"
```
adrake33: One more.

adrake33 (approved):
Looks good overall! Just had one minor comment about versions in the package.json files. There is a decent argument to be made for using "^", though, so we can talk about this more during today's standup.

PR #1571, opened 2019-04-04
fix fts: going back to setting method as a property

robert-e-davidson3:


packages/augur-api/src/api/Events.ts
```diff
@@ -23,9 +21,9 @@ export class Events {
         return this.parseLogs(logs);
     }
 
-    public getEventTopics(eventName: string) {
+    public getEventTopics = (eventName: string) => {
```
brianosaurus, line 24: @robert-e-davidson3 This is super weird.

Any idea why this fixes it? @adrake33 any thoughts?

brianosaurus (approved):

PR #1586, opened 2019-04-04
Fixed types in tests and added a test

robert-e-davidson3:
This PR is mostly about fixing the types used in the tests. I added a couple of tests toward that end but marked one as skipped because I'm not sure how to get it to work. The tests are for `UserSyncableDB`.

Alex has encountered a problem with log processing with ganache that might also be blocking the test I skipped. EDIT: That's still true; the issue has been identified as augur-artifacts being out of date.

### DIFF
--- a/packages/augur-test/tests/state/UserSyncableDB.test.ts
+++ b/packages/augur-test/tests/state/UserSyncableDB.test.ts
@@ -1,0 +1,27 @@
+import { BigNumber } from "ethers/utils/bignumber";
+import { ACCOUNTS, makeDbMock, makeTestAugur } from "../../libs";
+import { UserSyncableDB } from "@augurproject/state/src/db/UserSyncableDB";
+
+const mock = makeDbMock();
+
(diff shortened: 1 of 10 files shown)


packages/augur-test/tests/state/UserSyncableDB.test.ts
```diff
@@ -0,0 +1,88 @@
...
+  await db.sync(augur, mock.constants.chunkSize, mock.constants.blockstreamDelay, highestAvailableBlockNumber);
+
+  console.log(Object.keys(mock.getDatabases()));
+  console.log(db.dbName);
+
+  const tokensTransferredDB = mock.getDatabases()[`db/${db.dbName}`];
+  const docs = await tokensTransferredDB.allDocs();
+  console.log(docs);
+
+  expect(docs).toBe(42);  // TODO
```
nuevoalex, line 73: heh

nuevoalex (approved):
👍

PR #1603, opened 2019-04-05
added EventLogDBRouter.test.ts

robert-e-davidson3:
I worry this test is confusing even with a bit of commentary. So please LMK if it can be made clearer. Also if there are any more cases that stand out as good-to-test.


justinbarry (approved):
LGTM

PR #1604, opened 2019-04-05
added SyncStatus.test.ts

robert-e-davidson3:
Rebuilding gave me a new yarn.lock but I'm not sure if we want that or not.


adrake33 (approved):
LGTM

Issue #1446, opened 2019-03-22
Add more logs to Augur contracts


adrake33:
Some of the logs we will want to add include:
* ~~Fees when claiming trading proceeds~~
* ~~Fees when selling a complete set~~
* ~~Start & end dates for market creation events (needed for full-text search)~~
* ~~Forking market to `UniverseForked`~~
* ~~price to `OrderFilled`~~
* ~~Event logging for when a market has begun normal dispute pacing (possibly as a new param in DisputeCrowdsourcerCompleted)~~
* No-show bond amount
* Market info: ~~finalizationTime~~, ~~consensus~~, ~~endTime~~, ~~numTicks~~: https://gist.github.com/adrake33/c4b671181a1b4fb350b5dc6f18d11570
* Market price history (should use finalization payout amount): https://gist.github.com/adrake33/f7e71594182020a585ccf71306b83e31
* Market outcome price change (payout for that outcome should be used): https://gist.github.com/adrake33/d192651a004684cfcd8d347bc5a9bff6
* Market creator has unclaimed fees & market creator claimed fees: https://gist.github.com/adrake33/10c16da8ac291ef392cc6844a741b888

And more to come

joeykrug:
@nuevoalex status?

nuevoalex:
Closing this in favor of just using new issues to handle any additional logs we may need.

PR #1637, opened 2019-04-11
First cut of getMarketsInfo

### LINKED ISSUE #1134: getMarketsInfo getter
Create queries for getMarketsInfo to get a sense of what writing queries will entail.

adrake33:
Fixes https://github.com/AugurProject/augur/issues/1134. More changes will be needed for this getter once new logs are added (https://github.com/AugurProject/augur/issues/1446). This is a checkpoint with a few hard-coded values until those changes can be made.


packages/augur-state/src/db/DB.ts
```diff
@@ -300,25 +314,47 @@ export class DB<TBigNumber> {
...
    * @param {PouchDB.Find.FindRequest<{}>} request Query object
-   * @returns {Promise<Array<OrderCreatedLog>>}
+   * @returns {Promise<Array<CompleteSetsSoldLog>>}
    */
-  public async findOrderCreatedLogs(request: PouchDB.Find.FindRequest<{}>): Promise<Array<OrderCreatedLog>> {
-    const results = await this.findInSyncableDB(this.getDatabaseName("OrderCreated"), request);
-    return results.docs as unknown as Array<OrderCreatedLog>;
+  public async findCompleteSetsSoldLogs(request: PouchDB.Find.FindRequest<{}>): Promise<Array<CompleteSetsSoldLog>> {
+    const results = await this.findInSyncableDB(this.getDatabaseName("CompleteSetsSold"), request);
+    return results.docs as unknown as Array<CompleteSetsSoldLog>;
```
robert-e-davidson3, line 336: @justinbarry Are there good alternatives to use here instead of `as unknown`?

packages/augur-test/tests/state/api/Markets.test.ts
```diff
@@ -0,0 +1,193 @@
...
+const mock = makeDbMock();
+
+beforeEach(async () => {
+  mock.cancelFail();
+});
+
+let db: DB<any>;
+let api: API<any>;
+let john: ContractAPI;
+let mary: ContractAPI;
```
robert-e-davidson3, line 22: You can delete mary since you don't use her.

packages/augur-state/src/api/Markets.ts
```diff
@@ -1,33 +1,247 @@
...
 
   public async getMarkets(params: GetMarketsParams): Promise<void> {
     // TODO
   }
 
-  public async getMarketsInfo(params: MarketsInfoParams): Promise<void> {
-    // TODO
+  // TODO Add decorator code
+  // TODO Improve performance by using Promise.all for DB queries
+  public async getMarketsInfo(params: GetMarketsInfoParams): Promise<Array<MarketInfo>> {
```
robert-e-davidson3, line 87: I recommend cutting this up into smaller functions that you can then unit test. You've already organized this function into paragraphs; I think most of those paragraphs could be their own function.

packages/augur-test/tests/state/api/Markets.test.ts
adrake33: I'm planning to use mary eventually for trading to test volume/OI, so that's why I left her in this test.

packages/augur-state/src/api/Markets.ts
adrake33: Agreed, I was thinking the same thing. I was just planning to do that when I updated the getter to make use of the new log info.

packages/augur-test/tests/state/api/Markets.test.ts
```diff
@@ -0,0 +1,193 @@
...
+
+beforeAll(async () => {
+  const {provider, addresses} = await compileAndDeployToGanache(ACCOUNTS);
+
+  john = await ContractAPI.userWrapper(ACCOUNTS, 0, provider, addresses);
+  mary = await ContractAPI.userWrapper(ACCOUNTS, 0, provider, addresses);
```
robert-e-davidson3: Since you're keeping mary, make sure to use ACCOUNTS index `1` since john is already using `0`.

```suggestion
  mary = await ContractAPI.userWrapper(ACCOUNTS, 1, provider, addresses);
```

adrake33: Oops, thanks. I copied from the line above and then forgot to update the second param.

robert-e-davidson3 (approved):

PR #1681, opened 2019-04-15
1332 Feeds requests to apropos getter with io-ts

brianosaurus:
Ads a decorator pattern to augur state that routes requests to the apropos getter.

Uses io-ts to validate inputs.


packages/augur-test/tests/state/api/Router.test.ts
```diff
@@ -0,0 +1,49 @@
...
+import { Augur } from "@augurproject/api";
+import { DB } from "@augurproject/state/src/db/DB";
+import { Router } from "@augurproject/state/src/api/Router";
+import { ethers } from "ethers";
+import { makeTestAugur, ACCOUNTS, makeDbMock } from "../../../libs";
+
+const mock = makeDbMock();
+
+beforeEach(async () => {
+  mock.cancelFail();
```
robert-e-davidson3: I've been putting this in every test but it's not actually needed if you aren't telling the mockdb to fail.

robert-e-davidson3 (approved):

packages/augur-test/tests/state/api/Router.test.ts
brianosaurus: Removed. Thanks.

PR #1693, opened 2019-04-16
Users getters continued

nuevoalex:
Changes needed to make the user specific tables function.
Support for a table that has a custom id field to support just having the _latest_ log for a certain field composite.
More log data


adrake33 (approved):
lgtm

PR #1713, opened 2019-04-18
Adds http and https endpoints to augur-state

brianosaurus:
Creates a server that listens on http and https ports with a JSON-RPC interface.

Connects that to our API routing framework.


packages/augur-state/package.json
```diff
@@ -36,8 +36,12 @@
     "@augurproject/core": "2.0.0-0",
     "@augurproject/ethersjs-provider": "1.0.0-0",
     "@types/async": "2.4.1",
+    "@types/cors": "^2.8.4",
+    "@types/express": "^4.16.1",
+    "@types/helmet": "^0.0.43",
     "@types/web3": "1.0.18",
     "async": "2.6.2",
+    "cors": "^2.8.5",
```
adrake33: Just a reminder: these should have the caret removed.

packages/augur-test/package.json
```diff
@@ -31,17 +31,22 @@
...
+    "@augurproject/state": "1.0.0-0",
     "ganache-core": "^2.4.0"
   },
   "devDependencies": {
     "@types/jest": "24.0.4",
     "@types/node": "10.12.29",
+    "@types/supertest": "^2.0.7",
     "jest": "22.4.4",
     "pouchdb-adapter-memory": "7.0.0",
+    "supertest": "^4.0.2",
```
adrake33: Here too.

adrake33 (approved):
LGTM

packages/augur-state/src/IsJsonRpcRequest.ts
```diff
@@ -0,0 +1,6 @@
+export function isJsonRpcRequest(message: any): boolean {
+  if (message.jsonrpc !== "2.0") return false;
```
robert-e-davidson3, line 3: You may also want to test that `message` isn't `undefined` or 'null' since then `message.jsonrpc` will give an error.

packages/augur-state/src/index.ts
```diff
@@ -1,27 +1,21 @@
-export {API} from "./api/API";
-export {ServerController} from "./server/ServerController";
-export {Controller} from "./Controller";
+import RunWorker from "./Sync.worker";
 
-import RunWorker from "./run.worker";
+// start the web worker and move console.log over to paint in the browser
+console.log = (msg: string) => {
```
robert-e-davidson3: Does this mess with `console.log` calls made elsewhere?

brianosaurus: I left that there thinking it would be removed when augur-ui gets hooked up. However, you make a good point and I should just remove it.

packages/augur-test/tests/state/HTTPEndpoint.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  augur = await makeTestAugur(ACCOUNTS);
+  db = await mock.makeDB(augur, ACCOUNTS);
+}, 120000);
+
+test("HTTPEndpoint :: Responds to ping Json RPC Request ", async () => {
+  const api = new API(augur, db);
+  const app = HTTPEndpoint.createApp(api);  // fire up the endpoint
+
+  const result = await request(app).post("/").send({
+    jsonrpc: " 2.0",
```
robert-e-davidson3: There's a space in the `" 2.0"`. Is that intentional?

robert-e-davidson3 (approved):
This is really good code.

packages/augur-state/src/IsJsonRpcRequest.ts
brianosaurus: Added (even though github isn't showing that right now).

PR #1714, opened 2019-04-18
Adds methodIdentifiers to compilede output.

MicahZoltu:
These method identifiers are useful in a number of situations including signer tooling being able to present a useful method name and block explorers being able to decode method calls.  https://solidity.readthedocs.io/en/v0.5.7/using-the-compiler.html#output-description for where I got that particular path.

Note: I have _not_ tested this in any way whatsoever.


justinbarry (approved):
LGTM

PR #1738, opened 2019-04-18
Adds a websocket endpoint to augur-state

brianosaurus:
Adds a websocket interface to augur-state when run as a server.

Closes #1738

### DIFF
--- a/packages/augur-state/src/Server.ts
+++ b/packages/augur-state/src/Server.ts
@@ -84,8 +84,8 @@ export async function run() {
 
   // Sync.start();
   console.log("Starting websocket and http endpoints");
-  // WebsocketEndpoint.run(api, endpointSettings);
-  await HTTPEndpoint.run(api, endpointSettings);
+  HTTPEndpoint.run(api, endpointSettings);
+  await WebsocketEndpoint.run(api, endpointSettings);
 }
 
 run();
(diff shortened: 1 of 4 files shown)


packages/augur-state/src/Server.ts
```diff
@@ -26,66 +26,66 @@ export async function run() {
...
-    endpointSettings.httpPort = Number(process.env["HTTP_PORT".toString()]) ||
+    endpointSettings.httpPort = Number(process.env.HTTP_PORT) ||
       settings.endpointSettings.http.port;
   } catch {
     endpointSettings.httpPort = 9004;
   }
 
   try {
-    endpointSettings.startHTTPS = Boolean(process.env["STATRT_HTTPS".toString()]) ||
+    endpointSettings.startHTTPS = Boolean(process.env.STATRT_HTTPS) ||
```
pgebheim:
```suggestion
    endpointSettings.startHTTPS = Boolean(process.env.START_HTTPS) ||
```

pgebheim (changes requested):
Looks generally good, just need a typo fixed!

I also think we need to manually time out clients that don't do anything. Though I suppose they could get around that just by Ping/Ponging -- but something to stop people from just eating up public connections IF this thing is run publicly would be a nice to have.

pgebheim (approved):
LGTM

PR #1791, opened 2019-04-22
removes the force-recreate flag when running integrations tests

petong:
removes the flag that causes augur-core docker image to rebuild from scratch when running integration tests


robert-e-davidson3 (approved):
lgtm

PR #1830, opened 2019-04-23
Update OrderFilled log & add MarketCreated2 log

adrake33:


nuevoalex (changes requested):
Talked in discord and decided we prefer the tactic of grouping log params by type over multiple logs

PR #1864, opened 2019-04-25
Update logs

### LINKED ISSUE #1831: Update OrderFilled & MarketCreated logs for use by `getMarkets` getter
This also entails updating the solidity-typescript generator, the FTS code & numerous tests.

### LINKED ISSUE #1882: Update FTS code to work with `description` being added to `extraInfo`
Since logMarketCreated has too many params, we need to add description to extraInfo, which will require tweaking the FTS code & related tests.

### LINKED ISSUE #1883: Add support for int256[] in Solidity Typescript Generator
This will require forking & publishing as @augurproject/solidity-typescript-generator.

adrake33:
Fixes https://github.com/AugurProject/augur/issues/1831.
Fixes https://github.com/AugurProject/augur/issues/1882.
Fixes https://github.com/AugurProject/augur/issues/1883.

A number of the tests in augur-test are failing, but they were also failing in master when I ran them.


packages/augur-core/source/contracts/Augur.sol
```diff
@@ -229,22 +229,22 @@ contract Augur is IAugur {
...
         recordMarketShareTokens(_market);
         markets[address(_market)] = true;
-        emit MarketCreated(_universe, _endTime, _topic, _description, _extraInfo, _market, _marketCreator, _minPrice, _maxPrice, _marketType, 10000, _outcomes);
+        emit MarketCreated(_universe, _endTime, _topic, _extraInfo, _market,_marketCreator, _designatedReporter, _feeDivisor, _prices, _marketType, 10000, _outcomes);
         return true;
     }
 
     // This signature is intended for yesNo and scalar market creation. See function comment above for explanation.
-    function logMarketCreated(uint256 _endTime, bytes32 _topic, string memory _description, string memory _extraInfo, IMarket _market, address _marketCreator,int256 _minPrice, int256 _maxPrice, IMarket.MarketType _marketType, uint256 _numTicks) public returns (bool) {
+    function logMarketCreated(uint256 _endTime, bytes32 _topic, string memory _extraInfo, IMarket _market, address _marketCreator, address _designatedReporter, uint256 _feeDivisor, int256[] memory _prices, IMarket.MarketType _marketType, uint256 _numTicks) public returns (bool) {
```
nuevoalex, line 242: Would these function signatures still work (not produce stack depth errors) if we kept them as _minPrice and _maxPrice?

If so I think that's a bit nicer of a function signature. We only really need to use the array format for the log if we can get away with that.

nuevoalex (approved):
👍 

Have one recommendation if it's doable

packages/augur-core/source/contracts/Augur.sol
adrake33: I don't think so since MarketCreated already has 12 params, but I can double-check.

adrake33: Yep, I got a stack-depth error when changing these signatures to accept _minPrice & _maxPrice as separate params. I think we can output them as separate params in the actual MarketCreated log, though. Do you think it would be better to make that change?

nuevoalex: Oh woops I commented in the wrong function. this was intended to be about `Universe.createScalarMarket`

adrake33: ~~That's the one I changed the signature for, but it still gave the stack-depth error.~~
Never mind, I changed the signature for both of these. I'll try with only changing Universe.createScalarMarket.

adrake33: No luck. :-/ Changing _prices to _minPrice & _maxPrice in createScalarMarket's signature and then passing them into augur.logMarketCreated as _prices still throws the error.

PR #1941, opened 2019-04-29
1939 remove mailbox and fixups

### LINKED ISSUE #1939: Remove mailbox + fixups for v2 augur-node
* New contracts no longer have mailboxes, removed those.
* Fixups to get augur-node to run 

Issues remaining

* decrease token balance fails
* the create-market log processor fails when run outside of the debugger

brianosaurus:
Closes #1939 

* New contracts no longer have mailboxes, removed those.
* Fixups to get augur-node to run

Issues remaining

* decrease token balance fails
* the create-market log processor fails when run outside of the debugger


packages/augur-artifacts/src/docker.json
```diff
@@ -1,5 +1,5 @@
 {
- "augurproject/dev-pop-geth-v2": "88642924a317",
```
justinbarry: This file was removed in master.

justinbarry (approved):
Lgtm

PR #1944, opened 2019-04-29
remove signed trades

nuevoalex:
Removing this since there are third party mechanisms that can achieve this functionality and better


adrake33 (approved):
Lgtm

PR #1956, opened 2019-04-30
Begin to remove lerna

justinbarry:


pgebheim:
@petong --- Do you agree that we can dump Lerna?

If so I think we should just merge this quick.

petong (approved):
LGBTM!

PR #1996, opened 2019-05-03
changes block times from 1 second to 5 second

### LINKED ISSUE #1971: Dev Docker Images
all docker images have 5 second block times, removing 1 second images for v2

3 images that I can think of: 

 * fake time, with canned data
 * normal time, with canned data
 * unpopulated

petong:
Fixes #1971


bthaile (approved):
Lgtm

PR #2004, opened 2019-05-03
Order event refactor

nuevoalex:


packages/augur-core/source/contracts/Augur.sol
```diff
@@ -48,11 +56,23 @@ contract Augur is IAugur {
...
     event UniverseCreated(address indexed parentUniverse, address indexed childUniverse, uint256[] payoutNumerators);
-    event OrderCanceled(address indexed universe, address indexed shareToken, address indexed sender, IMarket market, bytes32 orderId, Order.Types orderType, uint256 tokenRefund, uint256 sharesRefund);
-    event OrderPriceChanged(address indexed universe, bytes32 orderId, uint256 outcome, uint256 price);
-    // The ordering here is to match functions higher in the call chain to avoid stack depth issues
-    event OrderCreated(Order.Types orderType, uint256 amount, uint256 price, address indexed creator, bytes32 tradeGroupId, bytes32 orderId, IUniverse indexed universe, IMarket indexed market, ERC20Token kycToken, uint256 outcome);
-    event OrderFilled(address indexed universe, address filler, address creator, IMarket market, bytes32 orderId, uint256 price, uint256 outcome, uint256 fees, uint256 amountFilled, bytes32 tradeGroupId, bool orderIsCompletelyFilled, uint256 timestamp);
+
+    //  addressData
+    //  0:  kycToken
+    //  1:  orderCreator (Fill)
```
adrake33: Looks like orderCreator is also logged by logOrderCanceled, but it was removed from OrderCreated? Do we not need it for OrderCreated?

I suppose shareToken isn't really needed for OrderCanceled, so that's probably ok to remove.

packages/augur-state/src/api/Trading.ts
```diff
@@ -74,26 +75,26 @@ export class Trading {
       if (!orderDoc) return trades;
       const marketDoc = markets[orderFilledDoc.market];
       if (!marketDoc) return trades;
-      const isMaker: boolean | null = params.account == null ? false : params.account === orderFilledDoc.creator;
+      const isMaker: boolean | null = params.account == null ? false : params.account === orderFilledDoc.addressData[OrderEventAddressValue.orderCreator];
       const orderType = orderDoc.orderType === 0 ? "buy" : "sell";
-      const fees = new BigNumber(orderFilledDoc.fees);
-      const minPrice = new BigNumber(marketDoc.prices[0]._hex);
-      const maxPrice = new BigNumber(marketDoc.prices[1]._hex);
+      const fees = new BigNumber(orderFilledDoc.uint256Data[5]);
```
adrake33: Shouldn't this be OrderEventUint256Value.feeDivisor instead of 5?

adrake33 (approved):
Looks good overall. Just a few minor comments.

packages/augur-core/source/contracts/Augur.sol
nuevoalex: Ah that a mis-lable in the comment really. orderCreator will be on all OrderEvents

adrake33: Ah right, it is getting set in `getOrderDataForLogs`. Ok, never mind 👍

PR #2010, opened 2019-05-03
Fixes reviver

brianosaurus:
No longer downcases addresses in the reviver.
Instead it uses ethers to check the address to make sure it is well formatted.
Changes are in augur-node and augur-state.


bthaile (approved):
lgtm

PR #2030, opened 2019-05-05
Make log processors work with OrderEvent

pgebheim:


nuevoalex (approved):
Glorious

bthaile (approved):
fixed processing logs, just need to start AN with `yarn start` not `yarn clean-start`

PR #2049, opened 2019-05-06
Ignore unimplemented augur-node methods in augur.js

justinbarry:


bthaile:
closing, augur-node will get stubbed out.

PR #2079, opened 2019-05-07
waits for geth to die before killing docker

petong:


brianosaurus (approved):
LGTM - I can see what wait is doing. Kind of strange that wait needed to happen however. I guess I don't understand it fully.

PR #2099, opened 2019-05-08
Add augur-api to ui

justinbarry:


bthaile (approved):
lgtm, good first stab at calling contracts with augur api.

PR #2102, opened 2019-05-08
[WIP] Augur core docker fix 20190503

petong:
* deletes venv dir if it exists, which would cause an error building 
* removes docker pull from run.sh so that if the image exists locally already use that, if not, fetch from docker hub
* add debug info to pop-geth image, showing last block after deploy
* tags augur-core docker image with contracts md5
* gets rid of python2 in docker image
* removes old node from augur.js matrix


bthaile (approved):
lgtm, lets do this

packages/augur-tools/docker-builder/run.sh
```diff
@@ -4,6 +4,6 @@ IMAGE=$1
 TAG=$(node ./scripts/get-contract-hashes.js)
 ./scripts/copy-docker-files.sh $IMAGE:$TAG
 echo "Running Geth with contracts deployed from ${IMAGE}:${TAG}"
-docker pull $IMAGE:$TAG
+#docker pull $IMAGE:$TAG
```
justinbarry, line 7: Is this debug code?

petong: no, we don't need that, running the image will pull from docker hub, but if you have it locally the pull will overwrite it. We should just delete the whole line

pgebheim: We actually just decided NOT to do that.

@petong lets just remove?

```suggestion
```

pgebheim:
```suggestion
```

Shall we remove it @petong

petong:
I made a new PR with the CI optimizations, and will include the python2 -> python3 changes separately

PR #2136, opened 2019-05-10
pts in redeem. handle invalid for affiliates

nuevoalex:
Small fix to affiliate fees (Send to dispute window on invalid)
Addition of PT redemption back to the universe redeem function


adrake33 (approved):
Lgtm 👍

Issue #2110, opened 2019-05-09
Simplify repetitive code in actions


JohnDanz:
Reduce repetitive code in favor of a less verbose option. A good example is orders/actions/cancel-order.js. It currently looks like this:
```
export const cancelOrder = (
  { orderId, marketId, outcome, orderTypeLabel },
  callback = logError
) => (dispatch, getState) => {
  const { loginAccount, orderBooks, outcomesData, marketsData } = getState();
  const order = selectOrder(
    orderId,
    marketId,
    outcome,
    orderTypeLabel,
    orderBooks
  );
  const market = marketsData[marketId];
  if (
    order &&
    market &&
    outcomesData[marketId] &&
    outcomesData[marketId][outcome]
  ) {
    dispatch(
      updateOrderStatus({
        orderId,
        status: CLOSE_DIALOG_PENDING,
        marketId,
        outcome,
        orderTypeLabel
      })
    );
    augur.api.CancelOrder.cancelOrder({
      meta: loginAccount.meta,
      _orderId: orderId,
      onSent: () =>
        dispatch(
          updateOrderStatus({
            orderId,
            status: CLOSE_DIALOG_PENDING,
            marketId,
            outcome,
            orderTypeLabel
          })
        ),
      onSuccess: () => {
        dispatch(
          updateOrderStatus({
            orderId,
            status: CLOSE_DIALOG_CLOSING,
            marketId,
            outcome,
            orderTypeLabel
          })
        );
        callback(null);
      },
      onFailed: err => {
        dispatch(
          updateOrderStatus({
            orderId,
            status: CLOSE_DIALOG_FAILED,
            marketId,
            outcome,
            orderTypeLabel
          })
        );
        setTimeout(
          () =>
            dispatch(
              updateOrderStatus({
                orderId,
                status: null,
                marketId,
                outcome,
                orderTypeLabel
              })
            ),
          TIME_TO_WAIT_BEFORE_FINAL_ACTION_MILLIS
        );
        callback(err);
      }
    });
  }
};
```

However, all we are really doing is updating the status, so we shouldn't repeat all those extra lines. Instead we can refactor this to:

```
export const cancelOrder = (
  { orderId, marketId, outcome, orderTypeLabel },
  callback = logError
) => (dispatch, getState) => {
  const { loginAccount, orderBooks, outcomesData, marketsData } = getState();
  const order = selectOrder(
    orderId,
    marketId,
    outcome,
    orderTypeLabel,
    orderBooks
  );
[...]

PR #2147, opened 2019-05-13
Actions to Typescript and small refactoring

JohnDanz:
Converted all actions to Typescript and refactored some of the more repetitive sections I saw.

https://github.com/AugurProject/augur/issues/2110


bthaile (approved):
moving to .ts files is good. changes look good

PR #2150, opened 2019-05-14
Fix test assertions

justinbarry:


robert-e-davidson3 (approved):
I'm not sure why I wasn't specifying `transactionLogIndex` before. I see that you set it to `0` earlier so this change makes sense.

`f303dc9b952 packages/augur-api/src/api/Events.test.ts (Justin                 2019-04-29 14:10:49 -0700 34)     transactionLogIndex: 0`

You also changed the `topics` value. It used to use what's in `logValues` instead of `logs` but I guess that changed. The code handling this is in `Events.ts`.

I doubt we care very much if the `topics` value in `logValues` or `logs` has precedence so I'm happy with this PR.

Issue #2158, opened 2019-05-14
remove all augur.rpc.* calls in UI


bthaile:
augur.rpc.* is used for these reasons:

* get account types from augur.rpc.constants;
* get ETHER from constants `const { ETHER } = augur.rpc.constants;`
* get accounts ` augur.rpc.eth.accounts((err: any, accounts: Array<any>)`
* get gas price `augur.rpc.getGasPrice()`
* get net version `augur.rpc.net.version((err: any, netVersion: string)`
* clear tx history `augur.rpc.clear(); // clear ethrpc transaction history, registered callbacks, and alerts`
* sha3 utility `const addressHash = augur.rpc.sha3(formattedAddress.toLowerCase());`
* get block by has ` augur.rpc.eth.getTransactionByHash(...`
* get transport `augur.rpc.getTransport().getTransportName()`
* tx relay `augur.rpc.excludeFromTransactionRelay(NO_RELAY);` and `augur.rpc.registerTransactionRelay((transaction: any)`
* set debug options `augur.rpc.setDebugOptions(env.debug);`

justinbarry:
| Old Method | New method |
| --- | --- |
| get account types from augur.rpc.constants;                           | Added here https://github.com/AugurProject/augur/pull/2159 |
| get ETHER from constants const { ETHER } = augur.rpc.constants;       | Added here https://github.com/AugurProject/augur/pull/2159 |
| get accounts augur.rpc.eth.accounts((err: any, accounts: Array<any>)  | Exists on the EthersProvider [here](https://github.com/AugurProject/augur/blob/master/packages/ethersjs-provider/src/EthersProvider.ts#L54) |
| get gas price augur.rpc.getGasPrice()                                 | This is EthRpc to be unraveled if we need it |
| get net version augur.rpc.net.version((err: any, netVersion: string)  | |
| clear tx history augur.rpc.clear(); // clear ethrpc transaction history, registered callbacks, and alerts | Will probably [this](https://gist.github.com/rsms/3744301784eb3af8ed80bc746bef5eeb) from the event emitter interface. |
| sha3 utility const addressHash = augur.rpc.sha3(formattedAddress.toLowerCase()); | Is [this](https://github.com/AugurProject/augur/blob/master/packages/augur-core/source/libraries/ContractDependenciesEthers.ts#L19) the same thing? |
| get block by has augur.rpc.eth.getTransactionByHash(... | provider.getBlock, unsure it the types actually expose it. |
[...]

PR #2168, opened 2019-05-15
Remove augur rpc

bthaile:
change out calls to augur.rpc with calls to ethers. put them in contractCalls.ts

another poke at https://github.com/AugurProject/augur/issues/2158


packages/augur-ui/src/modules/auth/components/common/hardware-wallet.jsx
```diff
@@ -203,10 +203,7 @@ export default class HardwareWallet extends Component {
           const promises = [];
           walletAddresses.forEach(addr => {
             const getPromise = new Promise((resolve, reject) => {
-              getEtherBalance(addr.address, (err, balance, address) => {
-                if (err) return reject(new Error(err));
-                resolve({ balance, address });
-              });
+              getEthBalance(addr.address).then(balance => resolve({ balance, address })).catch(err => reject(err));
```
andrewdamelio: ```getEthBalance``` used to return balance and address now it just returns balance, is that on purpose?

If so, the resolve on line 206 needs to be revised to remove the address reference.
maybe ```{ balance, address: addr.address }```

bthaile: updated PR, good catch.

andrewdamelio (approved):
LGTM 👍

Issue #2169, opened 2019-05-15
Type functions, actions and reducers


bthaile:
1. convert reducers to .ts files. use BaseAction for action type
2. types that selectors are likely to use go in modules/types.ts file
3. action/reducer constants and actions go in actions file (like current structure)

- [x]    accountDisputes,
- [x]    accountPositions,
~~- [ ]    alerts,~~ Will need major refactoring
- [ ]    appStatus, // isMobile and isMobileSmall will be refactored out.
- [x]    authStatus,
- [x]    blockchain,
- [x]    categories,
- [x]    connection,
- [x]    contractAddresses, // will be removed when SDK is fully wired up
- [x]    disputeCrowdsourcerTokens,
- [x]    env,
- [x]    eventsAPI, // removed, SDK events will replace
- [x]    favorites,
- [x]    filterSortOptions,
- [x]    functionsAPI, // removed, SDK contract calls will replace
- [x]    gasPriceInfo,
- [x]    initialReporters,// remove not needed
- [x]    loginAccount,
- [x]    marketReportState,
- [x]    marketTradingHistory,
- [x]    marketsData,
- [x]    marketsWithAccountReport,
- [x]    modal,
- [x]    newMarket,
- [x]    readNotifications,
- [x]    orderBooks,
- [x]    orderCancellation,
~~- [ ]    orphanedOrders,~~
- [x]    outcomesData,
- [x]    participationTokens,
- [x]    pendingLiquidityOrders,
- [x]    pendingOrders,
- [x]    pendingQueue,
- [x]    filledOrders,
- [x]    reportingWindowStats,
- [x]    reports,
- [x]    scalarMarketsShareDenomination,
- [x]    selectedMarketId,
- [ ]    sidebarStatus, // will get removed with new naviation
- [x]    transactionsData,
- [x]    transactionsStatus,
- [x]    universe,
- [x]    versions,
- [x]    accountShareBalances

PR #2177, opened 2019-05-16
Create typed functions

bthaile:
https://github.com/augurproject/augur/issues/2169

Trying out standard way of typing reducer and actions so we get strict types. Started with login account since it's used in many places. Converted reducer to .ts file. Remove unneeded tests that didn't test anything important.


justinbarry:
@bthaile Can you use discriminate unions to get type safety in the reducers?

Like this:
```typescript
// src/store/chat/types.ts
export const SEND_MESSAGE = 'SEND_MESSAGE'
export const DELETE_MESSAGE = 'DELETE_MESSAGE'

interface SendMessageAction {
  type: typeof SEND_MESSAGE
  payload: Message
}

interface DeleteMessageAction {
  type: typeof DELETE_MESSAGE
  meta: {
    timestamp: number
  }
}

export type ChatActionTypes = SendMessageAction | DeleteMessageAction
```

See: https://redux.js.org/recipes/usage-with-typescript

phoebemirman (approved):

PR #2179, opened 2019-05-16
use dai in the topbar

### LINKED ISSUE #2178: use dai in topbar
DAI will be the trading currency need to use dai instead of ETH in topbar

bthaile:
closes https://github.com/AugurProject/augur/issues/2178

use DAI in the top bar instead of ETH


andrewdamelio (approved):
Looks good now!

Issue #2183, opened 2019-05-16
Augur-sdk: consumer event emitter design


justinbarry:
Subscribe, Unsubscribe, and UnsubscribeAll functions for the connectors.

justinbarry:
ref #1786, #1969

justinbarry:
Some code that might be relevant here. It will give you the discriminant union of the getters params/return type.
```typescript
type GetterFunction = (augur: Augur<ethers.utils.BigNumber>, db: DB<ethers.utils.BigNumber>, params: any) => Promise<any>;
type GetterParamType<T extends GetterFunction> = Parameters<T>[2];
type ReturnPromiseType<T> = T extends (...args: any) => Promise<infer R> ? R : any;
type FunctionPropertyNames<T> = { [K in keyof T]: T[K] extends Function ? K : never }[keyof T];
type GetterFunctionTypes<T> = {
  [P in FunctionPropertyNames<T>]: T[P] extends GetterFunction ? (params:GetterParamType<T[P]>) => ReturnPromiseType<T[P]> : never
};
type UnionType<T> = T[keyof T];
type r = GetterFunctionTypes<typeof Markets>;
type union = UnionType<r>;

type FunctionProperties<T> = Pick<T, FunctionPropertyNames<T>>;

type t = FunctionProperties<typeof Markets>;
```

pgebheim:
@brianosaurus -- This epic really does to you now...

Should this just be a ticket within API design?

brianosaurus:
@pgebheim ok, sounds good.

PR #2202, opened 2019-05-18
Pgebheim/experimental connector

pgebheim:


pgebheim:
@justinbarry -- Here is my first draft of this.

- It uses cross-fetch to bring in a cross-env `fetch` polyfill.
- It only supports HTTP at the moment
- The test is written inline 'cause I didn't spend the time to figure out how/where you're putting the tests for this stuff, this comes with a npm script called `connect` for @augurproject/sdk which just runs that one test. That should be refactored.

General thoughts:
- Augur SDK needs a pass to get the styles the same. We have multiple file name formats for classes etc.
- There are places where we are explicitly importing from `src` -- this will break as soon as this is used as an NPM module. Can't TS automatically point the module at `src` in TS-land so that when its an npm module the main `build` directory will be isomorphic to the old `src` dir and relative imports would still work?

justinbarry:
1. Added discriminant union code to #2183.
2. Added code style issue #2207.

packages/augur-sdk/src/connector/http-connector.ts
```diff
@@ -0,0 +1,37 @@
+import { Connector, Callback } from "./connector";
+// TODO: use cross-fetch?
```
justinbarry:
```suggestion

```

Yes. Unless you prefer another package.

packages/augur-sdk/src/connector/http-connector.ts
```diff
@@ -0,0 +1,37 @@
...
+import fetch from "cross-fetch";
+
+export class HTTPConnector extends Connector {
+
+  constructor(public readonly endpoint: string) {
+    super();
+  }
+
+  public async connect(params: any): Promise<any> {
+
```
justinbarry:
```suggestion
return Promise.resolve();
```

packages/augur-sdk/src/connector/http-connector.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  public async disconnect(): Promise<any> {
+
```
justinbarry:
```suggestion
return Promise.resolve();
```

justinbarry (changes requested):
Requested simple changes. Otherwise looks good!

pgebheim:
@brianosaurus -- Here is my solution to the `sender` thing.

brianosaurus:
@justinbarry, if main in pakage.json points to src or build it'll automatically find the include. You and I chatted about a while ago. 

LGTM, I'll absorb this into the mostly finished websocket stuff. Personally I would have implemented BindTo in the base class and then just did comm out of the individual connectors. Pass it back and let API deal with deserializing  CC: @pgebheim

justinbarry (approved):

Issue #2139, opened 2019-05-13
Make table rows into one reusable row component


phoebemirman:
also move common components to the common-elements folder

PR #2208, opened 2019-05-20
Convert outcomes rows

phoebemirman:
extension of this https://github.com/AugurProject/augur/issues/2139

converted the outcomes list table rows to use shared row component


JohnDanz (approved):
looks good

PR #2221, opened 2019-05-21
small fixes to AN and UI

bthaile:
comment out orders log processor it's causing AN to throw, fix network id mismatch, need to compare numbers


andrewdamelio (approved):
Fixed my problems 🎉

Issue #2106, opened 2019-05-09
Refactor Component Elements


JohnDanz:
Remove unneeded element tags, aka overwrapped elements. We shouldn't have situations where for example we do something like this:

```
<div>
  <label>
    <span> some text </span>
  </label>
</div>
```

Drop the span, and consider dropping the label unless it actually is a label for an input.

Convert JSX to TSX as well and add typing.

The goal is to simplify the HTML involved in our elements so they are easier to style/position using Less.

We should also look to rename elements to remove `-` and `_` characters. Keep Things Simplified and look to merge common elements where it makes sense.

pgebheim:
@JohnDanz Is this still a useful ticket?

JohnDanz:
yeah, It's more of a general note on what we want to do, I could probably make more based on each section of code. I'll look to break it up and then i'll close it

pgebheim:
@JohnDanz @bthaile Closing as a major effort is complete.

Issue #2105, opened 2019-05-09
Refactor Styles


JohnDanz:
We need to do a pass to refactor the style sheets that are still referencing old color values or using verbose names.

Ideally we should be trying to slim down out style code so that we have very few classes that have simple naming conventions. Use PascalCase for classes not camelCase, do not use _ or -, try to limit classes to the overall element wrapper of a component and use encapsulated element targeting within that class to style elements of the component. small modifier/conditional classes are ok, things like `.open` or `.closed` for example to toggle height or display is fine. Do not name the classes with repetitive prefix-text. Example would be something like `market-view.styles.less` having `.MarketViewContainer` then having `.MarketViewTextBox` etc. instead just make a `.Container` class, and avoid the need for a seperate class for TextBox.

pgebheim:
@JohnDanz @bthaile Closing this major effort.

PR #2228, opened 2019-05-21
Account and Alerts refactor

JohnDanz:
This includes styles/hierarchy updates and most of the element refactoring i'd like to do. had some trouble with alerts testing but i'll circle back in a pr in the future.

does a bit of all 3 of these issues, but these issues are large. maybe I'll break them down into individual sections.

https://github.com/AugurProject/augur/issues/2105
https://github.com/AugurProject/augur/issues/2106
https://github.com/AugurProject/augur/issues/2110


phoebemirman (approved):
looks good, except the row font color thing

PR #2229, opened 2019-05-21
Fix some tests

robert-e-davidson3:
Aside from fixing some tests, I also made it possible to run the augur-ui tests from the toplevel. However I disabled actually running any UI tests until the UI code is ready to be tested again. I also consolidated the babel config into `.babelrc.js`.


bthaile (approved):
lgtm

PR #2230, opened 2019-05-22
Place trade

nuevoalex:
Initial work on the `placeTrade` API endpoint.

ALSO includes standardizing on bignumber.js and the removal of the `TBigNumber` generic support from Augur SDK and fixes augur-test tests by adding faucet calling back in.


robert-e-davidson3 (approved):
test changes lgtm

I skimmed the rest which also looks good.

justinbarry (approved):
LGTM

PR #2231, opened 2019-05-22
Upgrade solidity generator

justinbarry:


justinbarry:
Will not address before v2 release. So much has changed we might as well start anew.

PR #2237, opened 2019-05-22
rm unused code, move some components in market view to containers

phoebemirman:
https://github.com/AugurProject/augur/issues/2217


packages/augur-ui/src/modules/common-elements/labels.tsx
```diff
@@ -730,7 +731,7 @@ export const MovementLabel = (props: MovementLabelProps) => {
 };
 
 export const PillLabel = (props: PillLabelProps) => (
-  <span className={Styles.PillLabel}>{props.label}</span>
+  <span className={classNames(Styles.PillLabel, {[Styles.HideOnMobile]: props.hideOnMobile})}>{props.label}</span>
```
JohnDanz: now that I'm looking at this, can we destructure those props? so instead of `(props: PillLabelProps)` use `({ label, hideOnMobile }: PillLabelProps)`? I think that it helps make the props more clear at a glance but also lets use not keep doing `props.whatever`, probably should move that way for all the functional components as part of the refactoring efforts.

packages/augur-ui/src/modules/market-charts/containers/market-charts-pane.ts
```diff
@@ -0,0 +1,35 @@
+import { connect } from "react-redux";
+import { withRouter } from "react-router-dom";
+import MarketChartsPane from "modules/market-charts/components/market-charts-pane/market-charts-pane";
+import { selectMarket } from "modules/markets/selectors/market";
+import { createBigNumber } from "src/utils/create-big-number";
+import { getMarketAgeInDays } from "utils/format-date";
+
+const mapStateToProps = (state, ownProps) => {
```
JohnDanz: Missing Types here

packages/augur-ui/src/modules/market-charts/containers/market-charts-pane.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  return {
+    currentTimestamp,
+    daysPassed,
+    minPrice: market.minPrice || createBigNumber(0),
+    maxPrice: market.maxPrice || createBigNumber(0),
+  };
+};
+
+const mapDispatchToProps = (dispatch, ownProps) => ({});
```
JohnDanz: if there is nothing to dispatch just remove this and only pass `mapStateToProps` to the below `connect` function

JohnDanz (changes requested):
couple small changes otherwise looks good

JohnDanz (approved):

Issue #2112, opened 2019-05-09
Convert existing login methods to us augur SDK


bthaile:
convert existing login methods to use new augur SDK.
- [x] MM
- [x] Trezor
- [x] Ledger
- [x] Edge

PR #2240, opened 2019-05-23
update TrezorConnect to v7

andrewdamelio:
### Update TrezorConnect to v7 
- adds support for Trezor Model T
- starts to address #2112 
- refactor and lint files changes to typescript


packages/augur-ui/src/modules/auth/constants/trezor-manifest.ts
```diff
@@ -0,0 +1,2 @@
+export const TREZOR_MANIFEST_EMAIL = "team@augur.net";
+export const TREZOR_MANIFEST_APP = "https://try.augur.net";
```
bthaile, line 2: what's the manifest app, what about mainnet? since this is Rinkeby augur site.

andrewdamelio: Trezor requires it, see below. Due to the nature of Augur, I assumed try.augur.net would suffice.

>> Starting with Trezor Connect 7, we have implemented a new feature — Trezor Connect Manifest — which requires that you as a Trezor Connect integrator, to share your e-mail and application url with us. This provides us with the ability to reach you in case of any required maintenance. This subscription is mandatory. Trezor Connect raises an error that reads "Manifest not set.

bthaile (approved):
lgtm, made some changes to getEtherBalance. still wasn't able to get a balance to show up getEtherBalance returned 0 even though there is ETH on my trezor address.

PR #2246, opened 2019-05-23
use the correct big number and format atto to display value for rep, …

bthaile:
dont use ethers bignumber use big-number.js also format atto to display value for rep, eth and dai.


andrewdamelio (approved):
LGTM

PR #2247, opened 2019-05-23
process bigNumber in tx then delete from, issue with ethers

bthaile:
issue with ethers have to delete `from` key.


nuevoalex (approved):
👍

andrewdamelio (approved):

PR #2248, opened 2019-05-23
Clean up market styles

phoebemirman:
https://github.com/AugurProject/augur/issues/2245

### DIFF
--- a/packages/augur-ui/src/modules/market/components/market-trade-history/market-trade-history.styles.less
+++ b/packages/augur-ui/src/modules/market/components/market-trade-history/market-trade-history.styles.less
@@ -19,122 +19,112 @@
   }
 }
 
-.MarketTradeHistory__table {
-  display: table;
-  position: sticky;
-  top: 0;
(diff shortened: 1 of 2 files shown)


packages/augur-ui/src/modules/market/components/market-trade-history/market-trade-history.styles.less
```diff
@@ -9,164 +9,122 @@
...
+    background-color: @color-tertiary-background;
+    color: @color-secondary-text;
+    display: flex;
+    font-size: @size-11;
+    padding-bottom: @size-4;
+    position: relative;
+    width: 100%;
+
+    > li {
+      font-family: @roboto-mono;
```
JohnDanz: use the `.mono-whateversizefontthisshouldbe` here instead of directly editing font-family

JohnDanz (changes requested):
Tiny change, otherwise perfect

JohnDanz (approved):

PR #2249, opened 2019-05-23
added 'out' type def file for us to put our own defs in, import corre…

bthaile:
…ct type def for classnames, user correct sdk sig when creating object. added type def for .less and .styles. notice no read underlines. crazy!!!

image


andrewdamelio (approved):
lgtm nice little lint cleanup also

PR #2242, opened 2019-05-23
Populate test chain with seed data

justinbarry:
Closes #2224.

PR #2252, opened 2019-05-24
Populate ganache chain before running tests

robert-e-davidson3:
Also made it optional to write addresses and such to disk since it's unnecessary and messy due to those files being checked into git.


packages/augur-test/src/scripts/GenerateGanacheSeed.ts
```diff
@@ -0,0 +1,87 @@
...
+    leveledDB.createReadStream({
+      keyAsBuffer: false,
+      valueAsBuffer: false,
+    }).on("data", (data: LevelDBRow) => {
+      payload.push({
+        type: "put",
+        ...data,
+      });
+    }).on("error", (err: Error) => {
+        console.log("Oh my!", err);
```
nuevoalex, line 90: 👍

nuevoalex (approved):
LGTM

packages/augur-test/src/scripts/GenerateGanacheSeed.ts
robert-e-davidson3: Credit goes to @justinbarry :)

robert-e-davidson3:
Stole pieces of https://github.com/AugurProject/augur/pull/2242 to make this PR better.

Issue #2259, opened 2019-05-27
Fix TypeScript errors 


andrewdamelio:

pgebheim:
@andrewdamelio -- Are there more TS errors to fix or is this issue closed?

andrewdamelio:
There are MORE issues, but the ones I targeted with this ongoing issue had been completed.

PR #2261, opened 2019-05-27
Update to use types 1

bthaile:
https://github.com/AugurProject/augur/issues/2259

type format-number, format-date and helper.ts


andrewdamelio (approved):
👍 wicked

PR #2262, opened 2019-05-28
Styles refactor

JohnDanz:
App section refactor... still probably have a bit more work to do but i want to get my file deletions/merges in place since we are doing things with typing that may cause merge conflicts/mistakes/files to be recreated that we want to remove.


bthaile (approved):
Lgtm

PR #2272, opened 2019-05-28
type fixes - /positions

andrewdamelio:
#2259 -  starts fixing the errors generated by ```--type-checking``` so we can eventually turn on --type-checking as part of our workflow.

## First pass at the below folders
- [x] /positions/actions
- [x] /positions/reducer
- [x] /positions/selectors


bthaile (approved):
lgtm, @andrewdamelio check out my changes to account position. I looked up the actual shape on old preview site to verify.

PR #2280, opened 2019-05-29
connection/blockinfo refactoring

JohnDanz:
just some style touch ups in connection, but generally left it the same because it's going to be redone in the redesign

block info rearranged hierarchy, stripped out some classes/refactored styling, still some older color values but need to talk about replacing those with design. Updated the Html a bit to be less divs and more logical html like h3s/spans/buttons etc


bthaile (approved):
lgtm

PR #2283, opened 2019-05-29
condense trading wrapper styles, rename trading components to be simp…

phoebemirman:
…ler names


packages/augur-ui/src/modules/trading/components/form/form.styles.less
```diff
@@ -1,81 +1,83 @@
...
-    color: @color-secondary-text;
-    display: flex;
-    flex-flow: column nowrap;
-    font-size: @font-size-normal;
-    width: 100%;
+    > li {
+      color: @color-secondary-text;
+      display: flex;
+      flex-flow: column nowrap;
+      font-size: @font-size-normal;
```
JohnDanz: remove this and use `.text` here, also if you see any other `@font-size-normal` replace them with new variables and or use `.text/.mono` versions for font specific things

packages/augur-ui/src/modules/trading/components/form/form.styles.less
```diff
@@ -150,13 +152,13 @@
   }
 }
 
-.TradingForm__outcome__container__yes {
+.Yes {
   background-color: @color-theme-primary;
   border: 1px solid @color-primary-alternate;
   border-radius: @border-radius-default;
   font-size: @size-14;
```
JohnDanz: Same comment as above, replace with `.text-14-medium` (also make that var if it doesn't exist yet in `redesign_typography`) here instead of directly modifying font-size/font-weight

packages/augur-ui/src/modules/trading/components/form/form.styles.less
```diff
@@ -169,7 +171,7 @@
 .TradingForm__input {
   display: flex;
   flex-grow: 1;
-  padding: 0.4375rem 1rem;
+  padding: 0.4375rem @size-16;
```
JohnDanz: I believe this would be `@size-7` which may or may not already exist, but if it doesn't add it and replace that `0.4375rem` amount

packages/augur-ui/src/modules/trading/components/form/form.styles.less
```diff
@@ -196,22 +198,22 @@
...
   > span {
     background: @color-tertiary-background;
-    border-left: 0.0625rem solid @color-primary-alternate;
+    border-left: 1px solid @color-primary-alternate;
     color: @color-tertiary-text;
     font-size: @font-size-small;
     min-width: 4.28rem;
     overflow: hidden;
-    padding: 0.59375rem 0.75rem;
+    padding: 0.59375rem @size-12;
```
JohnDanz: replace `0.59375rem` with either `size-9` or `size-10`, currently that rem value is 9.5px so it's your call which to use but we should move away from half px in the new design

packages/augur-ui/src/modules/trading/components/form/form.styles.less
```diff
@@ -196,22 +198,22 @@
...
+      border: 1px solid fade(@color-error-text, 50%);
     }
   }
 
   > span {
     background: @color-tertiary-background;
-    border-left: 0.0625rem solid @color-primary-alternate;
+    border-left: 1px solid @color-primary-alternate;
     color: @color-tertiary-text;
     font-size: @font-size-small;
```
JohnDanz: replace font-size with `.text` var

packages/augur-ui/src/modules/trading/components/form/form.styles.less
```diff
@@ -363,8 +365,8 @@
       border-radius: 0;
       bottom: 0;
       left: 0;
-      padding-bottom: 1rem;
-      padding-top: 1rem;
+      padding-bottom: @size-16;
+      padding-top: @size-16;
```
JohnDanz: merge these into a single `padding: @size-16 0;` so it's only one line/less chars

JohnDanz (changes requested):
a few style preferences to change otherwise looks good

bthaile (approved):
Lgtm

JohnDanz (approved):

PR #2286, opened 2019-05-29
add ThunkAction for dispatch

bthaile:
https://github.com/AugurProject/augur/issues/2259

update dispatch type to ThunkAction

### DIFF
--- a/packages/augur-ui/src/modules/app/actions/init-augur.ts
+++ b/packages/augur-ui/src/modules/app/actions/init-augur.ts
@@ -39,7 +39,7 @@ const NETWORK_ID_POLL_INTERVAL_DURATION = 10000;
 
 function pollForAccount(
   dispatch: ThunkDispatch<void, any, Action>,
-  getState: Function,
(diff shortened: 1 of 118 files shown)


packages/augur-ui/src/modules/app/actions/init-augur.ts
```diff
@@ -253,22 +256,22 @@ interface initAugurParams {
 export function initAugur(
   history: any,
   { augurNode, ethereumNodeHttp, ethereumNodeWs, useWeb3Transport }: initAugurParams,
-  callback = logError
+  callback = logError,
```
andrewdamelio: ```NodeStyleCallback```

andrewdamelio (approved):

packages/augur-ui/src/modules/app/actions/init-augur.ts
bthaile: updated to catch those.

andrewdamelio (approved):

PR #2287, opened 2019-05-29
Pending tx get

nuevoalex:


packages/augur-sdk/src/Augur.ts
```diff
@@ -101,4 +101,16 @@ export class Augur<TProvider extends Provider = Provider> {
   public getOrders():ContractInterfaces.Orders {
     return new ContractInterfaces.Orders(this.dependencies, this.addresses.Orders);
   }
+
+  public registerTransactionStatusCallback(key: string, callback: TransactionStatusCallback): void {
```
bthaile, line 105: at the SDK level couldn't `key` be typed? is it `event name` or transaction type? I think we'd benefit with some typing here if I understand how it's going to be used by the UI.

nuevoalex: Its arbitrary. its the name of the generic tx status update handler as provided by the consumer

bthaile: I see in the test "test" is used what would be an example the UI or other consumer would use?

justinbarry: The key exists to give the consumer the ability to deregister the callback. I suppose we could return a closure the  does the same thing but that is more stuff to manage.

nuevoalex: Yea. In practice it would be something like "Notifications" for a handler specific to notifications.

bthaile: gotcha, that makes sense. so we'd only have one handler to manage tx in UI.

justinbarry: If the same key is used twice that would kill the previous callback. Is that what we want here?

bthaile (approved):
lgtm

packages/contract-dependencies-ethers/src/ContractDependenciesEthers.ts
```diff
@@ -83,13 +117,39 @@ export class ContractDependenciesEthers implements Dependencies<BigNumber> {
...
         if (!this.signer) throw new Error("Attempting to sign a transaction while not providing a signer");
         // TODO: figure out a way to propagate a warning up to the user in this scenario, we don't currently have a mechanism for error propagation, so will require infrastructure work
         // TODO: https://github.com/ethers-io/ethers.js/issues/321
         const tx = this.transactionToEthersTransaction(transaction);
         delete tx.from;
-        const receipt = await (await this.signer.sendTransaction(tx)).wait();
+        const txMetadataKey = `0x${transaction.data.substring(10)}`;
+        const txMetadata = this.transactionDataMetaData[txMetadataKey];
+        this.onTransactionStatusChanged(txMetadata, TransactionStatus.AWAITING_SIGNING);
+        const response = await this.signer.sendTransaction(tx);
```
justinbarry: If this throws an exception we will have excess txMetadata hanging around. Would be good to handle if we could.

packages/augur-sdk/src/Augur.ts
nuevoalex: I think that's fine personally. Alternative is to throw an error which I think is obnoxious.

justinbarry (changes requested):
Looks *really* good. Just one feedback item.

packages/contract-dependencies-ethers/src/ContractDependenciesEthers.ts
nuevoalex: will do

justinbarry (approved):

PR #2294, opened 2019-05-31
changes conditionals for master merge

petong:
git command to check for changes is different, now we don't look at
comments.

don't run tests on merge to master, since PRs need to be up to date,
they have already run test/lint


justinbarry (approved):
LGTM

PR #2295, opened 2019-05-31
type fixes for /modal

andrewdamelio:
#2259 - fixing the errors generated by --type-checking so we can eventually turn on --type-checking as part of our workflow.

- [x] ```/modules/modal/*```

**Some items here will need to be revisited.**


bthaile (approved):
Lgtm

PR #2314, opened 2019-05-31
fix open order row toggling on market trading page

phoebemirman:
noticed the open order row on trading page had a toggle, which is wrong.


bthaile (approved):
lgtm

PR #2315, opened 2019-05-31
type fixes for /markets-list

andrewdamelio:
#2259 - fixing the errors generated by --type-checking so we can eventually turn on --type-checking as part of our workflow.

- [x] ```/modules/markets-list/*```


bthaile (approved):
lgtm

PR #2316, opened 2019-05-31
spell out types for now, couldn't get ReturnType to work

bthaile:
had to spell out interface for AppState, couldn't get ReturnType<> to work


andrewdamelio (approved):
works for me!

PR #2337, opened 2019-06-04
Py evm

nuevoalex:
Switch to py-evm and python3


packages/augur-core/tests/conftest.py
```diff
@@ -404,95 +432,101 @@ def uploadAugur(self):
...
-        marketCreationFee = validityBond or universe.getOrCacheMarketCreationCost()
+    def distributeRep(self, universe):
+        # Get the reputation token for this universe and migrate legacy REP to it
+        reputationToken = self.applySignature('ReputationToken', universe.getReputationToken())
+        legacyRepToken = self.applySignature('LegacyReputationToken', reputationToken.getLegacyRepToken())
+        totalSupply = legacyRepToken.balanceOf(self.accounts[0])
+        legacyRepToken.approve(reputationToken.address, totalSupply)
+        reputationToken.migrateFromLegacyReputationToken()
+
+    def getLogValue(self, eventName, argName, last=False):
```
robert-e-davidson3: This should probably be two functions: getFirstLogValue and getLastLogValue.

robert-e-davidson3:
Naming is a bit of a mess. PEP8 standardizes on lower_with_underscores unless the prevailing style is something else.

https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

packages/augur-core/tests/libraries/token/test_standard_token.py
```diff
@@ -162,91 +161,91 @@ def test_transfrom_more_than_total_supply(testStandardTokenFixture):
...
-    assert standardToken.revokeOperator(tester.a0, sender=tester.k1)
+    assert standardToken.revokeOperator(testStandardTokenFixture.accounts[0], sender=testStandardTokenFixture.accounts[1])
 
-    assert not standardToken.isOperatorFor(tester.a0, tester.a1)
+    assert not standardToken.isOperatorFor(testStandardTokenFixture.accounts[0], testStandardTokenFixture.accounts[1])
 
-    with raises(TransactionFailed, message="unauthorized account cannot use operator send"):
-        standardToken.operatorSend(tester.a2, 1)
+    with raises(TransactionFailed):
+        standardToken.operatorSend(testStandardTokenFixture.accounts[2], testStandardTokenFixture.accounts[0], 1, "", "")
```
robert-e-davidson3, line 216: I prefer named arguments when the alternative is a long list of mysterious values.

packages/augur-core/tests/old_eth_utils.py
```diff
@@ -0,0 +1,546 @@
...
+
+
+def safe_ord(value):
+    if isinstance(value, int):
+        return value
+    else:
+        return ord(value)
+
+# decorator
+
```
robert-e-davidson3: extra spaces?

nuevoalex:
Main issue with naming is we have a different standard for Solidity (camel case) and the contract currying in libraries always simply uses the same contract function name.

packages/augur-core/tests/libraries/token/test_standard_token.py
nuevoalex: This is a contract call (so just a wrapper around an ABI call / transaction). Would be quite a bit of work to add this everywhere.

packages/augur-core/tests/old_eth_utils.py
```diff
@@ -0,0 +1,546 @@
...
```
robert-e-davidson3, line 2: This file's name makes me think it's deprecated. Is there a better name? Or add a file-level docstring.

nuevoalex: Not really deprecated in this instance since it'll probably stay here forever. Just a port of the old pyethereum utils since they're heavily depended on in one of the test suites

packages/augur-core/tests/conftest.py
nuevoalex: Never used so I'm gonna just remove it

packages/augur-core/tests/libraries/token/test_standard_token.py
robert-e-davidson3: Ah nevermind then.

robert-e-davidson3:
> Main issue with naming is we have a different standard for Solidity (camel case) and the contract currying in libraries always simply uses the same contract function name.

Ah that's annoying.

robert-e-davidson3 (approved):
I got through a decent chunk of changes but there are too many and almost all are entirely formulaic changes.

What I saw LGTM though.

PR #2341, opened 2019-06-05
Augur 2333 parameterize connectors

### LINKED ISSUE #2333: Parameterize augur-sdk eth node URL via connector

justinbarry:
Things to note:
1. The sort param in load-markets is commented out. Sorting in the getters currently causes and exception to be thrown.
2. The ethnode url is hardcoded on the client side. The UX will need to be sorted out for this. ( paging @bthaile )
3. Google bot language level transpiring has been removed in webpack for external dependencies. Unknown is this impacts indexibility (A new version of the google crawler was just announced at google i/o).
closes #2333


packages/augur-sdk/src/connector/ww-connector.ts
```diff
@@ -1,35 +1,14 @@
...
-import { Controller } from "../state/Controller";
-import { EthersProvider } from "@augurproject/ethersjs-provider";
-import { EventLogDBRouter } from "../state//db/EventLogDBRouter";
-import { JsonRpcProvider } from "ethers/providers";
-import { PouchDBFactory } from "../state//db/AbstractDB";
-import { SubscriptionEventNames } from "../constants";
-import { UploadBlockNumbers, Addresses } from "@augurproject/artifacts";
-
-const settings = require("@augurproject/sdk/src/state/settings.json");
+import {API} from "../state/getter/API";
```
brianosaurus: I think your editor is auto formatting for no spaces in the import lists but others (mine included) are doing the opposite. We should sort that out some day.

packages/augur-sdk/src/events.ts
```diff
@@ -11,7 +11,7 @@ class EventNameEmitter extends EventEmitter {
         event,
       );
     });
-    return super.emit(type, eventsWithName);
+    return super.emit(type, ...eventsWithName);
```
brianosaurus, line 14: Curious why this was needed? I believe it'll break the connectors as written. They map over the event data but that's probably a bug. We'll need to retool the connectors to allow more than one argument to a Callback type. Currently the callback type is: 
```ts
export type Callback = (data: any) => void;
```

packages/augur-sdk/src/state/index.ts
```diff
@@ -1,48 +1 @@
...
-//     const getMarkets = augur.bindTo(Markets.getMarkets);
-//     console.log(await getMarkets({
-//       universe: "0x02149d40d255fceac54a3ee3899807b0539bad60",
-//     }));
-//     console.log("Done");
-//   } catch (e) {
-//     console.log(e);
-//   }
-// })();
+const settings = require("./settings.json");
```
brianosaurus, line 14: I think this needs to be 
```ts
@augurproject/sdk/src/state/settings.json
```

brianosaurus (dismissed):

packages/augur-sdk/src/events.ts
justinbarry: I added it because it is inline with the type definition of the parent's method definition. 

See: https://nodejs.org/api/events.html#events_emitter_emit_eventname_args

packages/augur-sdk/src/state/index.ts
justinbarry: Adding the file to the include block in tsconfig.json did the trick. Relative json file imports work again!

packages/augur-node/src/runServer.ts
```diff
@@ -9,14 +9,19 @@ import { JsonRpcProvider } from "ethers/providers/json-rpc-provider";
 
 import { Addresses } from "@augurproject/artifacts";
 import { ContractDependenciesEthers } from "contract-dependencies-ethers";
+import {EmptyConnector} from "@augurproject/sdk/build";
```
pgebheim: importing from `build` Perhaps 'cause this is just to get AN working?

pgebheim (approved):

packages/augur-node/src/runServer.ts
justinbarry: Nope. Just bad auto complete! Good catch!

PR #2344, opened 2019-06-05
Fixes connector initialization issue

brianosaurus:
Connectors were getting reinitialized with the EmptyConnector in the WorkerConnector and/or the SEOConnector. This fixes that problem.


bthaile (approved):
lgtm

PR #2346, opened 2019-06-06
fix errors in contractCalls, add needed methods in Augur.ts

bthaile:
part of UI integration with SDK.

Basic type fixes in UI. 
Fixed calls to contract methods (added "_")
Exposed a few methods on ethers provider in Augur.ts that the UI uses.


packages/augur-ui/src/modules/app/actions/sync-blockchain.ts
```diff
@@ -40,6 +40,7 @@ export const syncBlockchain = (cb: Function) => (
 
   cb && cb();
 
+  // @ts-ignore
   augur.augurNode.getSyncData((err: any, res: any) => {
```
brianosaurus, line 44: There's an Augur.getSyncData on master. Is that OK to use?

brianosaurus (changes requested):

packages/augur-ui/src/modules/app/actions/sync-blockchain.ts
bthaile: We’ll have a specific effort to pull out the augur-node stuff.

brianosaurus:
👍

brianosaurus (approved):

PR #2352, opened 2019-06-06
moved all files to common-elements from common

JohnDanz:
Finished moving files to common-elements from common, more style cleanup, moved a few SVGs that are only used once to the file they are used in. Still have some more refactoring to do but it depends on new designs so held off for now. Next step is to move `common-elements` to just `common` and replace all imports but wanted to get these changes in first


packages/augur-ui/src/modules/market-charts/components/market-outcome--header-orders/market-outcome--header-orders.tsx
```diff
@@ -1,12 +1,82 @@
 import React from "react";
 import PropTypes from "prop-types";
 import classNames from "classnames";
-
-import { collapseIcon, expandIcon } from "modules/common/components/icons";
-
 import StylesHeader from "modules/market/components/market-outcomes-list/market-outcomes-list.styles.less";
 import Styles from "modules/market-charts/components/market-outcome--header-orders/market-outcome--header-orders.styles.less";
 
+const collapseIcon = (
```
bthaile, line 7: should these icons go in icons.tsx?

bthaile (approved):
lgtm, one non-crit comment.

packages/augur-ui/src/modules/market-charts/components/market-outcome--header-orders/market-outcome--header-orders.tsx
JohnDanz: they are only used once, so i moved them to where they are used simply because eventually we should replace them with icons that are more common/reused. I didn't want to pollute the icons file with more one off icons, there are a few in there now but they are really specific to like menu icons which matt said was going away anyway. I added TODOs for those to remove them when we get the redesigned menu

PR #2357, opened 2019-06-07
Simulate trade

### LINKED ISSUE #2336: Re-implement Simulate Trade
We want an updated cut off Simulate Trade for displaying estimated cost of trades inclusive of fees and cost.

These estimated values use the current order book, and may require being re-evaluated after every chunk of trades that goes through with the high-level placeTrade function to be accurate.

nuevoalex:
Closes #2336


bthaile (approved):
Lgtm

adrake33 (approved):
Lgtm as well. Strange that the tests are failing, though. It doesn't seem like these changes should have caused that.

PR #2359, opened 2019-06-07
filtersort/forking styles

JohnDanz:
updated styles around filter sort and forking, didn't really touch forking elements because they will be redesigned. moved all component files to component folders to reduce nested folder structure.


bthaile (approved):
Lgtm

PR #2360, opened 2019-06-07
style updates + market list hierarchy

JohnDanz:
style updates in markets, moved markets-lists hierarchy to just put everything in components over sub folders


bthaile (approved):
Crushing it, lgtm

PR #2398, opened 2019-06-10
Fix api npe on startup

### LINKED ISSUE #2397: this.api undefined on initial load of application when running web worker connector

justinbarry:
closes #2397


andrewdamelio (approved):
👍 no errors incognito

PR #2400, opened 2019-06-11
2181 output sync status getter data

justinbarry:


packages/augur-sdk/src/state/getter/sync-data.ts
```diff
@@ -16,8 +15,9 @@ export class SyncData {
 
   @Getter()
   public static async getSyncData(augur: Augur, db: DB, params: t.TypeOf<typeof SyncData.SyncDataParams>): Promise<SyncData> {
+    const dbName = db.getDatabaseName("MarketCreated");
     const highestAvailableBlockNumber = await augur.provider.getBlockNumber();
-    const lastSyncedBlockNumber = await db.syncStatus.getHighestSyncBlock("MarketCreated");
+    const lastSyncedBlockNumber = await db.syncStatus.getHighestSyncBlock(dbName);
```
brianosaurus, line 20: In my PR that should be done today I've made it so that getHighestSyncBlock can take no dbName and it'll return the highest one of all of them.

brianosaurus (approved):
LGTM - I'll make whatever mods when I merge my PR.

packages/augur-sdk/src/state/getter/sync-data.ts
justinbarry: @brianosaurus I like that better. What is done if there is dissent on the highest synced block?

Issue #2115, opened 2019-05-09
Remove Lodash


JohnDanz:
Remove all calls to Lodash and replace with regular old JS. Once all references to lodash are removed, remove it from package deps.

pgebheim:
@JohnDanz Should we keep this scheduled for next sprint or should it go to backlog?

JohnDanz:
Lets put this in the next sprint, I've been removing lodash calls when I see it but we should assign someone to handle it all at once and be done with it

bthaile:
looks like there is still work to do here moving to backlog and out of sprint.

pgebheim:
Good idea, but not a priority.

PR #2429, opened 2019-06-11
Removed most of lodash

JohnDanz:
Most of the lodash functions have been removed. Still have a few tougher situations to resolve around the following functions:

```
has (2 occurrences)
orderBy (7 occurrences)
pullAll (1 occurance)
map (1 occurance left)
mapValues (1 occurance)
cloneDeep (1 occurance)
```

Everything else has been refactored/replaced. Will work on these, but wanted to get most of the changes in. https://github.com/AugurProject/augur/issues/2115


bthaile (approved):
lgtm, we really need to get our linting/formatting on the same page. lots of removing commas that were recently added.

PR #2445, opened 2019-06-12
Update sdk login flow for MM/injected web3

andrewdamelio:
https://github.com/AugurProject/augur/issues/2112

- init sdk with the injected provider only if its unlocked and logged in.
- changes around using window.ethereum to retrieve users account so we can respond correctly to locked / disabled wallets
- provide action to update sdk
- wired it up for injected web3 login flow
- tested with MM and dapper


bthaile (approved):
lgtm

Issue #2035, opened 2019-05-06
Login/Account Connection


bthaile:
3rd party wallet integrations.

existing: 
meta mask
Edge
Ledger and Trezor

new:
- [x] dapper
- [x] portis
- [x] Fortmatic

PR #2463, opened 2019-06-13
add portis

andrewdamelio:
relates to #2035 

- Add Portis wallet
- integrations with SDK
- misc type fixes

**NOTE:**
- Currently, when using Portis with a local node (i.e 104) you will see the below errors, expected, but looking into a way to avoid them.

- If your interested, you can access your Portis wallet at https://wallet.portis.io/. **FYI -  You won't see balances related to activity on custom nodes.**


packages/augur-ui/package.json
```diff
@@ -99,6 +100,7 @@
     "shifty": "2.6.1",
     "speedomatic": "2.1.5",
     "trezor-connect": "7.0.3",
+    "web3": "^1.0.0-beta.55",
```
bthaile: don't use automatic upgrades remove `^`

andrewdamelio: 👍 updated

packages/augur-ui/src/modules/auth/actions/login-with-portis.ts
```diff
@@ -0,0 +1,76 @@
...
+import { ACCOUNT_TYPES } from "modules/common/constants";
+import { AppState } from "store";
+
+// TODO find home for all wallet API keys
+const PORTIS_API_KEY = "b67817cf-8dd0-4116-a0cf-657820ddc019";
+
+const getPortisNetwork = (networkId): false | string | INetwork  => {
+  const myPrivateEthereumNode = {
+    nodeUrl: "http://localhost:8545",
+    chainId: "104",
```
bthaile, line 20: we actually have many private node chainIds. See `augur/packages/augur-artifacts/src/addresses.json`
don't use anything with augur-node. use `getNetworkId()` in contractCalls.ts

bthaile (changes requested):
a few comments for changes.

packages/augur-ui/src/modules/auth/actions/login-with-portis.ts
andrewdamelio: ahh that makes more sense!

bthaile (approved):
lgtm

Issue #2469, opened 2019-06-13
JUNIT reporter outputting invalid unicode characters


justinbarry:
Once this is merged and released we will need to upgrade the package version used in our repo.

https://github.com/jest-community/jest-junit/pull/94

justinbarry:
Closing. We have enough nice things.

PR #2470, opened 2019-06-13
Fix junit invalid characters

justinbarry:
Note: this WILL fail initially.

ref #2469


bthaile:
Fixed merge conflicts

PR #2471, opened 2019-06-13
Remove errant .nvmrc file

justinbarry:
nvm will recusively look up this file and use the one at the root of
the repo.


bthaile (approved):
lgtm

PR #2472, opened 2019-06-14
Removed pullAll lodash method

JohnDanz:
removed pullAll lodash method, a few more references to lodash still exist will remove later.


bthaile (approved):
lgtm

Issue #477, opened 2018-12-04
Implement Trading API changes in Augur UI


petong:
stuff needed:

 - [x] display trade confirmation
 - [x] display gas costs
 - [x] call get Order book getter
 - [x] call place trade

--------
* verify:
  * error conditions and warnings work in order form

pgebheim:
EPIC DONE

PR #2508, opened 2019-06-18
Get create order working

bthaile:
https://github.com/augurproject/augur/issues/477

First cut at order form using simulate trade and place trade. order creation tx get sent to MM and they get signed but never get confirmed. Not sure what is going on after the tx leaves the UI/SDK.

Had to reset my MM account, tx get confirmed. looks good


nuevoalex (approved):
Awesome!

PR #2510, opened 2019-06-18
Erc777 cleanup

nuevoalex:
ERC777 implementation changed to be based off OZ provided contracts.
Explicit mention of alterations
Use of safe transfer and burn internally such that hooks arent called.
Cash made into just a standard ERC20 to match reality.


adrake33:
Looks good overall! A few minor things (~~putting this here since GH is being dumb and won't let me view files/add comments on this PR~~ Edit: Apparently now it is showing the files/allowing comments):

- Glad to see comments that can be used for auto-generating documentation!
- I like that `require`s in this PR have error messages. Are we planning to do this throughout the rest of the code before v2 launch? Ideally, we should be doing this for all `require` statements to aid with debugging. 
- Nit: a few variables were declared as uint instead of uint256.

adrake33 (approved):

PR #2512, opened 2019-06-18
Outcome volume

### LINKED ISSUE #2473: add volume to Outcome from getMarketInfos
add volume to individual outcomes that are returned from getMarketsInfo. If no volume, `null` should be returned.

adrake33:
Fixes https://github.com/AugurProject/augur/issues/2473


bthaile (approved):
Lgtm

PR #2525, opened 2019-06-19
Add getMarketOrderBook getter

### LINKED ISSUE #2257: Get OrderBook getter
UI needs to get the order per market per outcome. taking in parameters
"marketId"
"outcomeId"

This getter can consume other getters like "get orders"

Then orders will need to be aggregated based on price and sorted by price.

The data needs to be in the shape (open for discussion)
```
{
marketId: "marketId0x"
outcome: "outcomeId",
 "bids": [ Order1, order2, order3],
 "asks": [ Order1, order2, order3]
}
```

justinbarry:
Closes #2257

### DIFF
--- a/packages/augur-sdk/src/state/logs/types.ts
+++ b/packages/augur-sdk/src/state/logs/types.ts
@@ -200,6 +200,13 @@ export enum OrderEventType {
   Fill = 3,
 }
 
+export enum OrderState {
+  ALL = "ALL",
+  OPEN = "OPEN",
+  FILLED = "FILLED",
+  CANCELED = "CANCELED",
+}
+
 
 export enum OrderEventAddressValue {
   kycToken = 0,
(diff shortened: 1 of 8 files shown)


packages/augur-sdk/src/state/logs/types.ts
```diff
@@ -200,6 +200,13 @@ export enum OrderEventType {
   Fill = 3,
 }
 
+export enum OrderState {
```
bthaile, line 203: trading.ts has these OrderState values as well. are they needed here?

justinbarry: Are you referring to the order_type defined in constants? If so, those are different things.

packages/augur-test/src/tests/state/getter/Markets.test.ts
```diff
@@ -1,1032 +1,1839 @@
...
+          expect.objectContaining({
+            0: expect.objectContaining({}),
+          })
+        );
+      });
+
+      test('can be an array', async () => {
+        const orderBook = (await api.route('getMarketOrderBook', {
+          marketId: yesNoMarket.address,
+          outcomeId: [outcome0.toNumber(), outcome1.toNumber()],
```
adrake33, line 1177: Nit: It might be more intuitive to just call this `outcomeIds` and have it either be an array or be omitted.

adrake33 (approved):
Lgtm

packages/augur-test/src/tests/state/getter/Markets.test.ts
justinbarry: Fair point. I *think reportingState follows the same pattern. So I just followed along.

PR #2534, opened 2019-06-20
removed linting in augur-ui in favor of monorepo linting

robert-e-davidson3:


bthaile (approved):
I like augur-ui repo to be consistent, yea!

PR #2567, opened 2019-06-24
No child REP migration and auto-update fork values

nuevoalex:


bthaile (approved):
I think I understand, get total REP supply if genesis return initial rep supply. if after fork return current universe REP supply. if during forking return current REP supply plus parent rep supply.

PR #2571, opened 2019-06-25
2481 types for events

brianosaurus:
Fixes the ww-connector


bthaile:
As pointed out in discord. 
```
  public messageReceived(data: any) {
    if (this.subscriptions[data.eventName]) {
      this.subscriptions[data.eventName].callback(data.result);
    }
  }
```
`data.eventName` is undefined so callback never gets called

bthaile (approved):
Lgtm

PR #2577, opened 2019-06-25
Various audit nits

nuevoalex:
- Remove the boolean return of onTransferOwnership in ownable
 - Check if contract when registering with Augur
 - LegacyReputationToken’s decimals public getter does not return uint8
 - Remove "WithCash" variants in complete sets. Do this for Trading too with "totalCost".
 - remove afterInitialized.
 - Remove the return values from private and internal functions that dont use return value
 - Make market ownership transfer also give ownership of the rep bond


adrake33 (approved):
Looks good! 👍

Issue #2539, opened 2019-06-20
create component for number lists


phoebemirman:
![image.png](https://images.zenhubusercontent.com/574e216bbf93871ccd9be5fb/3df78658-04d8-4e75-a979-c1371e37f32f)

cillianhunter0:
Link for the min max states of this component
https://www.figma.com/file/IpIAYVQPeQfuT5TTUboQpo4S/Market-creation?node-id=1027%3A69449

pgebheim:
Completed Jun 25

PR #2579, opened 2019-06-25
Numbered Lists Component

JohnDanz:
https://github.com/AugurProject/augur/issues/2539

added numbered list and numberedListItem to create-market common components, added PlusIcon to icons, added an option icon prop to SecondaryButton


packages/augur-ui/src/modules/create-market/components/common.styles.less
```diff
@@ -1,38 +1,106 @@
...
+  align-items: center;
+  display: flex;
+  flex-flow: row nowrap;
+
+  > span {
+    .text-14;
+
+    color: @color-alternate-info-text;
+    flex: 0 0 auto;
+    padding: 1rem;
```
phoebemirman: should this and the one below be @size-16?

phoebemirman (approved):

packages/augur-ui/src/modules/create-market/components/common.styles.less
JohnDanz: good catch, fixed

Issue #2428, opened 2019-06-11
MC: Review step


cillianhunter0:
Figma (yes/no):
https://www.figma.com/file/IpIAYVQPeQfuT5TTUboQpo4S/Market-creation?node-id=804%3A247

Figma (multiple choice):
https://www.figma.com/file/IpIAYVQPeQfuT5TTUboQpo4S/Market-creation?node-id=426%3A6

Figma (scalar):
https://www.figma.com/file/IpIAYVQPeQfuT5TTUboQpo4S/Market-creation?node-id=804%3A0

Figma (mobile):
https://www.figma.com/file/IpIAYVQPeQfuT5TTUboQpo4S/Market-creation?node-id=463%3A41512

**Applies to all**
- Preview market button opens a full screen overlay (fades in) displaying the trading page [Figma desktop] (https://www.figma.com/file/IpIAYVQPeQfuT5TTUboQpo4S/Market-creation?node-id=714%3A3) [Figma mobile] (https://www.figma.com/file/IpIAYVQPeQfuT5TTUboQpo4S/Market-creation?node-id=725%3A375)
   - Chart set to depth chart
   - Trade history set to empty state 
   - Open orders - hide the cancel orders action 
   - Close button closes overlay (fade out) 
- IF user doesn't have enough funds (REP/DAI/ETH), THEN display a box that highlights this and provides a link to purchase [Figma] (https://www.figma.com/file/IpIAYVQPeQfuT5TTUboQpo4S/Market-creation?node-id=714%3A1975)
   - IF it's DAI/ETH, THEN display box below validity bond data, IF it's REP, THEN display box below no-show bond data 
   - Link will open modal for relevant purchase flow, 0xinstant etc or other 

**Yes/No**
- Market details
   - Displays: market type, category, sub-category, market quesiton
- Resolution information
   - Displays: reporting start date and time, resolution details, resolution source, designated reporter 
- Funds required
   - Validity bond (REP)
   - Estimated gas cost (ETH)
   - No-show bond (DAI)
   - Estimated gas cost(ETH)
   - Initial liquidity (DAI)
   - Estimated gas cost (ETH)
- Totals
   - Total DAI
   - Total ETH 
   - Total REP 

**Multiple choice**
- Same as Yes/No, except for
   - Outcomes displayed in the market details section 

**Scalar**
- Same as Yes/No, except for
   - Unit of measurement displayed in the market details section 
   - Numeric range displayed in the market details section 
   - Precision displayed in the market details section

PR #2580, opened 2019-06-25
Review step

phoebemirman:
more of the review setup and made it so pages could be reused easily. 

https://github.com/AugurProject/augur/issues/2428


packages/augur-ui/src/modules/create-market/components/common.tsx
```diff
@@ -26,4 +34,50 @@ export const Subheaders = (props: SubheadersProps) => (
...
+);
+
+export interface ExplainerBlockProps {
+  title: string;
+  subtexts: Array<string>;
+}
+
+export const ExplainerBlock = (props: ExplainerBlockProps) => (
+  <div className={Styles.ExplainerBlock}>
+    <span>
```
JohnDanz: can we make this a hSomething tag, like h2 or something

packages/augur-ui/src/modules/create-market/components/common.tsx
```diff
@@ -26,4 +34,50 @@ export const Subheaders = (props: SubheadersProps) => (
...
+      {props.title}
+    </span>
+    {props.subtexts.map(subtext => 
+      <span>
```
JohnDanz: and assuming this is actually a paragraph style subtext then use a p tag, otherwise the span is fine if it's just like one sentence.

packages/augur-ui/src/modules/create-market/components/form.tsx
```diff
@@ -96,40 +97,27 @@ export default class Form extends React.Component<
     } = this.props;
     const s = this.state;
 
+    const content = CUSTOM_CONTENT_PAGES[newMarket.currentStep];
```
JohnDanz: since we don't actually use content directly, we should deconstruct this:

` const { mainContent, explainerBlockTitle, explainerBlockSubtexts, <etc> } = CUSTOM_CONTENT_PAGES[newMarket.currentStep];`

packages/augur-ui/src/modules/create-market/components/review.styles.less
```diff
@@ -1,5 +1,17 @@
 @import (reference) "~assets/styles/shared";
 
 .Review {
-	
+	display: flex;
```
JohnDanz, line 4: this looks weirdly formatted (the whole page has odd tabbing)

packages/augur-ui/src/modules/create-market/components/review.tsx
```diff
@@ -30,7 +39,90 @@ export default class Review extends React.Component<
```
JohnDanz, line 39: similar to the above comment, we should not keep repeating the `newMarket` part, deconstruct the newMarket object and then just reference `category` or `tag1` etc

JohnDanz (changes requested):

PR #2581, opened 2019-06-25
added market creation methods to Augur.ts

robert-e-davidson3:


packages/augur-sdk/src/api/Market.ts
```diff
@@ -0,0 +1,104 @@
...
+    const universe = this.augur.contracts.universe;
+
+    const events = await universe.createCategoricalMarket(
+      params.endTime,
+      params.feePerCashInAttoCash,
+      params.affiliateFeeDivisor,
+      params.designatedReporter,
+      params.outcomes,
+      formatBytes32String(params.topic),
+      params.extraInfo
```
justinbarry, line 68: Should we be validating extraInfo here vs expecting a string?

robert-e-davidson3: It is actually just a string. In the contract, it's purely a string that we log. We interpret it as JSON in the UI (and maybe also the middleware?).

justinbarry (changes requested):
One feedback item.

bthaile (approved):
lgtm, I have code in the UI that converts from display values to onChain values. I'll put that into Augur.ts when the market creation branch gets merged up to master

packages/augur-sdk/src/api/Market.ts
justinbarry: It is a string in the contract for event index reasons or some such efficiency thing. They would be first class params if there were room.

justinbarry: I'll file another ticket to get this sorted.

justinbarry (approved):

packages/augur-sdk/src/api/Market.ts
robert-e-davidson3: We can make it an interface then do the json serialization just before the contract call. That will break a ton of tests though so it'd be better in its own ticket.

Issue #2536, opened 2019-06-20
Order form changes


bthaile:
Order form changes:

- [x] 1. insufficient funds when user has plenty of dai to trade
- [x] 2. make every market type have outcome dropdown 
- [x] 3. test that user has approval for trade amount (create new warning, looks like insufficient funds)
- [x] 4. remove RED outcome, change to more muted color (design ultimate will say)
- [x] 5. move invalid to the bottom, added sort order in market selector
- [x] 6. identify default selected outcome in market selector
- [x] 7. add gas cost to order form 
- [x] 8. add error message for insufficient funds (DAI)
- [x] 9 add error message for insufficient ETH to cover gas

PR #2589, opened 2019-06-26
Order form updates

andrewdamelio:
#2536

#### Order form changes
- remove RED outcome color

- add an error message for insufficient funds (DAI)

- add an error message for insufficient ETH to cover gas

 
- add an error message for insufficient approved DAI to cover order

 
- make every market type have outcome dropdown and move invalid to the bottom (dropdown and outcome list)

image
image
image
- add gas cost to order form


bthaile (approved):
lgtm

PR #2598, opened 2019-06-27
De-normalize params and export keyed object vs io-ts validator

justinbarry:


adrake33 (approved):
LGTM if we decide to make this change.

Issue #2409, opened 2019-06-11
MC: landing page 


cillianhunter0:
Figma (desktop):
https://www.figma.com/file/IpIAYVQPeQfuT5TTUboQpo4S/Market-creation?node-id=433%3A4699

Figma (desktop - incl saved drafts): 
https://www.figma.com/file/IpIAYVQPeQfuT5TTUboQpo4S/Market-creation?node-id=419%3A2984

Figma (mobile):
https://www.figma.com/file/IpIAYVQPeQfuT5TTUboQpo4S/Market-creation?node-id=459%3A37543

Figma (desktop - incl saved drafts): 
https://www.figma.com/file/IpIAYVQPeQfuT5TTUboQpo4S/Market-creation?node-id=459%3A38470

Four sections: 
- Saved drafts
   - Only visible if the user has saved drafts (hidden otherwise) 
   - If user clicks a market they are taken to the step of the flow where they saved the draft
- Use a market template
   - Learn more links will open modals using the content layout 
      - Including the "learn more" under "don't see your category" we won't allow users to create or select a different primary category in the templated flow, we will tell them to use custom flow
   - Categories use the radio button card component 
- Start from scratch
   - Learn more link will open modals using the content layout 
   - Create a custom market button takes user to the custom flow 
- Import a market template 
   - Learn more link will open modals using the content layout 
   - Import a template button opens a modal with ability to upload a market template file 
      - See flow and states [here] (https://www.figma.com/file/IpIAYVQPeQfuT5TTUboQpo4S/Market-creation?node-id=817%3A3)

IF REP in users wallet doesn't match current no-show bond
- THEN display a panel at the top of the page that highlights they can't create a market [see mock] (https://www.figma.com/file/IpIAYVQPeQfuT5TTUboQpo4S/Market-creation?node-id=898%3A550)
- IF their REP balance exceeds the no-show bond, THEN don't show this panel

PR #2601, opened 2019-06-27
add landing page

phoebemirman:
https://github.com/AugurProject/augur/issues/2409


packages/augur-ui/src/modules/common/icons.tsx
```diff
@@ -21,6 +21,58 @@ export const AugurLoadingLogo = (
   </svg>
 );
 
+export const Sports = (
```
JohnDanz, line 24: These icons need to be cleaned up a little bit, remove the extra attributes like xmlns. Also fix all the `-` to use camel case instead as our linter will eventually not like this. so `stroke-width` => `strokeWidth`

JohnDanz (approved):
slight change request to the icons added.

PR #2605, opened 2019-06-28
call getters to get market history, user's open orders

### LINKED ISSUE #2537: Call get orders
getOrders can populate:

Data:
- [x] Market trading history
- [x] User open orders
- [x] User filled orders

hydrate orders redux state when:

1. user logs in 
    - [x] * get orders to populate portfolio open orders and filled orders
2. when they land on market trading page.
    - [x] * get market trade history
    - [x] * when user logs in market trading page will show user's open/filled orders

### LINKED ISSUE #2602: SDK doesn't recognize user changing accounts
Using web3 provider the `this.signer` is null so `ContractDependenciesEthers` so the address can't be obtained from signer and SDK defaults to the first address of provider. 

```
    public async getDefaultAddress(): Promise<string> {
        if (this.signer) {
            return this.signer.getAddress();
        }

        const addresses = await this.provider.listAccounts();
        if (addresses.length > 0) return addresses[0];

        return <string>this.address;
    }
```

needs to check if `this.address` is set and use it.

bthaile:
fixes https://github.com/AugurProject/augur/issues/2602
fixes https://github.com/AugurProject/augur/issues/2537

 * added creator and filler to market trading history orders from getter
 * added a fallback in Contracts Dependencies Ethers to look for this.account and return it before falling back to asking the provider and returning first address in the list
 * remove outcomesData reducer
 * fixed random issue to show market history, user's open orders and filled orders

image

image

image


andrewdamelio (approved):
lgtm

PR #2606, opened 2019-06-28
clean up UI hand-off when creating a market

bthaile:
first integration with Augur.ts new methods to create market.

When market creation redesign has the new categories hierarchy worked out, then push display to onChain conversions into Augur.ts.


andrewdamelio (approved):
lgtm

PR #2616, opened 2019-06-28
Consume positions getter

### LINKED ISSUE #2457: UI: get market trading page working
Use all needed getters to populate market trading page. 

show correct price information, show market trading history, order book

use new simulate trade. 
allow user to place trade.

bthaile:
fixes https://github.com/AugurProject/augur/issues/2457

pulled in add data for the market trading page having to trade. Charts are in a different tix. 

fixed incorrectly showing user was self trading when we weren't parsing up data form order book getter correctly. all fixed.


packages/augur-ui/src/modules/account/components/open-markets.tsx
```diff
@@ -8,7 +8,7 @@ import { SizeTypes } from "modules/types";
 import Styles from "modules/account/components/open-markets.styles.less";
 
 function filterComp(input: any, market: any) {
-  return market.description.toLowerCase().indexOf(input.toLowerCase()) >= 0;
+  return market ? market.description.toLowerCase().indexOf(input.toLowerCase()) >= 0 : false;
```
bthaile, line 11: protect if market is null

phoebemirman (approved):

PR #2617, opened 2019-06-28
Fix response values in getUserTradingPositions

### LINKED ISSUE #2612: changes to getUserTradingPositions getter
user's position needs to be in display terms

example:
      "position": "10000000000000000", needs to be       "position": ".1",

calculation issue with realized percent.
      "realizedPercent": "NaN",

timestamp is 0
      "timestamp": 0,a

adrake33:
Fixes https://github.com/AugurProject/augur/issues/2612

### DIFF
--- a/packages/augur-sdk/src/state/getter/Users.ts
+++ b/packages/augur-sdk/src/state/getter/Users.ts
@@ -66,7 +66,6 @@ export interface MarketTradingPosition {
 
 export interface TradingPosition {
   timestamp: number;
-  position: string;
   frozenFunds: string;
   marketId: string;
   outcome: number; // user's position is in this market outcome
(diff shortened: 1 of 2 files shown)


packages/augur-sdk/src/state/getter/Users.ts
```diff
@@ -229,7 +229,9 @@ export class Users {
                 ].balance
               );
             }
-            tradingPosition.position = rawPosition.toFixed();
+            tradingPosition.position = rawPosition
```
bthaile: I think I made a mistake on `position` the UI should use netPosition. `position` can be removed.

bthaile (approved):
lgtm, since we use netPosition, `position` property can be removed.

PR #2624, opened 2019-06-30
Update readme and fix coverage

nuevoalex:
Coverage takes over an hour to run  but works now.

Added tests for some missing lines and fixed a bug it exposed as well.


packages/augur-core/tests/conftest.py
```diff
@@ -357,8 +428,6 @@ def resetToSnapshot(self, snapshot):
         if not 'snapshot_id' in snapshot: raise "snapshot is missing 'snapshot_id'"
         if not 'contracts' in snapshot: raise "snapshot is missing 'contracts'"
         self.eth_tester.revert_to_snapshot(snapshot['snapshot_id'])
-        #if self.coverageMode:
-        #    self.chain.head_state.log_listeners.append(self.writeLogToFile)
```
adrake33, line 361: Should these lines be removed?

adrake33 (approved):
Changes look good, but test_reporting.py is currently failing.

PR #2648, opened 2019-07-03
New block event

### LINKED ISSUE #2620: NewBlock event needs timestamp property
* Chnage SyncStatus to include a *syncing* property for each db
* Add timestamp of each block processed to SyncStatus
* When done syncing then sync status tables should say *listening*
* Update new block event and getter to use these so that
```typescript
      augurEmitter.emit(SubscriptionEventName.NewBlock, {
        eventName: SubscriptionEventName.NewBlock,
        highestAvailableBlockNumber,
        lastSyncedBlockNumber: blockNumber,
        blocksBehindCurrent,
        percentBehindCurrent,
        timestamp,
    }
```

### LINKED ISSUE #2628: on NewBlock event give ethereum latest block separate from last block processed
The UI needs to know if the middleware block processing has fallen behind for some reason. 
1. basing latest block number off of pre-log processing. 
2. then latest processed block number will based on the logs processed.

brianosaurus:
Closes #2620 
Closes #2628 

* Refactors how NewBlock events are emitted
* Changes BlockAndLogStreamListener so it lets modules listen for new blocks
* Changes provider so that it allows retrieval of a block by blockNumber
* Changes SyncStatus so that it knows of a SyncableDB is still syncing or not.


bthaile (commented):
looks good so far

packages/augur-artifacts/src/addresses.json
```diff
@@ -110,19 +110,19 @@
...
-  "ClaimTradingProceeds": "0xF265D8D30a1A2CdB9857e124010b02765c9a7c70",
-  "CompleteSets": "0x4DdEBCEbE274751dfB129EfC96a588a5242530Ab",
-  "CreateOrder": "0x905419e7A1f96A973AB8988Ba19C84135C6f7122",
-  "FillOrder": "0xce77d3c706e0fe6dA5d71b1249756b5ea0D77202",
-  "Orders": "0x635C8EF61A07dfe49b683D1A34da3547b0d6705E",
-  "ProfitLoss": "0x257b56115544E9519Fb2De5b7b99C3c2150528Fb",
-  "ShareToken": "0x8bDAd8Ee0b47E2e7294A5B8b8B74EA3968f4a177",
-  "SimulateTrade": "0x204cDD1689c8A4da426894C150a0fa672a2eab4C",
-  "Trade": "0xf3bCabd8FAE29F75be271eBE2499EDB4c7C139b7"
+  "Universe": "0x8062dA104239cf70C76B77c61eA988bf6382736a",
```
bthaile: Prob dont need to check in the addresses changes.

packages/augur-artifacts/src/upload-block-numbers.json
```diff
@@ -1,13 +1,13 @@
...
-  "4": 4663954,
-  "19": 1340162,
-  "42": 7926279,
-  "101": 113,
-  "102": 4,
-  "103": 3,
-  "104": 1,
-  "8995": 0,
-  "12346": 123
+ "1": 5926223,
```
bthaile: Or the upload block changes

packages/augur-sdk/src/state/create-api.ts
```diff
@@ -0,0 +1,51 @@
+import { Augur } from "../Augur";
```
bthaile, line 1: Should the UI be using this to create augur abject?

brianosaurus: Good question. I took your changes and moved them into a filename other than index. That was all I did re: crate and buildAPI.

packages/augur-test/src/tests/connector/seo-connector.test.ts
```diff
@@ -120,11 +120,12 @@ test('SEOConnector :: Should route correctly and handle events', async done => {
...
         {
           eventName: SubscriptionEventName.NewBlock,
-          blocksBehindCurrent: 0,
+          blocksBehindCurrent: expect.any(Number),
           highestAvailableBlockNumber: expect.any(Number),
           lastSyncedBlockNumber: expect.any(Number),
-          percentBehindCurrent: '0.0000',
-        },
+          percentBehindCurrent: expect.any(String),
+          timestamp: expect.any(Number),
```
justinbarry, line 127: Do we have enough control to use specific values here to check calculations?

brianosaurus: Sadly no because events come in during processing. The values keep changing and it's ephemeral (timestamp for example) so I can't nail them down.

packages/augur-artifacts/src/upload-block-numbers.json
brianosaurus: These are the values on master, no idea why they came up in this PR. I merged with master and double checked. The values are the same.

packages/augur-artifacts/src/addresses.json
brianosaurus: These are the values on master, no idea why they came up in this PR. I merged with master and double checked. The values are the same.

packages/augur-sdk/src/state/create-api.ts
brianosaurus: Looking further ... I don't know how to get the UI to use this stuff. There's a significant issue here in that Augur doesn't set the correct connector. It uses the EmptyConnector so these methods can only be used by backend stuff. It isn't too clean but I think this should be fixed in another PR.

justinbarry: Not quite. This creates a provider. The UI might already have one injected. There could be some drying up to do here.

bthaile: Having a clear way for the UI to create the api object would be helpful. UI switches between creating without account address, with account address or switching account address.

justinbarry: We can probably sugar this up for you but it is out of scope here.

packages/augur-test/src/tests/connector/seo-connector.test.ts
```diff
@@ -54,15 +54,17 @@ jest.mock('@augurproject/sdk/build/state/index', () => {
...
 
   john = await ContractAPI.userWrapper(ACCOUNTS[0], provider, addresses);
   db = mock.makeDB(john.augur, ACCOUNTS);
 
   await john.approveCentralAuthority();
+
+  connector = new SEOConnector();
+  console.log("Connector connecting")
+  await connector.connect('');
 }, 120000);
```
robert-e-davidson3, line 68: This value can likely be much lower, like `15000`.

robert-e-davidson3 (approved):
lgtm

Issue #2146, opened 2019-05-13
UI wire up Events and pending tx events


bthaile:
In UI keep track of object `status` remove `pending`. 

example:

1. User creates order, UI calls SDK to create order (UI handles instant failure)
2. SDK emits event when tx is submitted. The event will have all properties that were used in calling the tx.
3. based pending info, UI adds order in state with status as `pending`
4. when tx confirms (pending tx gets updated minus the confirmed tx), UI gets a log for `OrderCreated` and replaces the order in state. 
 - if user is on market trading page for the market the whole order is refreshed.

UI notes on orders:
building the order book based on pending tx. 
order book only gets created/updated when user is on the market trading page.
log emitted that affects the orderbook
cancel, filled, created. 

signed Transactions the UI is interested in (list not complete)

Misc:
ReputationToken.faucet
Cash.faucet

Trading:
Trade.publicFillBestOrder
Trade.publicTrade
CancelOrder.cancelOrder
ClaimTradingProceeds.claimTradingProceeds
CompleteSets.publicSellCompleteSets
CreateOrder.publicCreateOrder
Market.finalize

Reporting:
DisputeWindow.buy
DisputeWindow.redeem
Market.contribute
Market.doInitialReport
Universe.redeemStake
InitialReporter.redeem
DisputeCrowdsourcer.redeem

middleware side <https://github.com/AugurProject/augur/issues/2111> tix that Ui will hook up to.

pgebheim:
@bthaile -- is this in progress w/ the stuff on Monday about submitting txns?

PR #2660, opened 2019-07-04
fix issue where log-handler event callbacks weren't getting the event…

bthaile:
https://github.com/AugurProject/augur/issues/2146

a few fixes to get event logs to log-handler. also fixed account comparison issue.


andrewdamelio (approved):
LGTM

PR #2665, opened 2019-07-05
if address isn't filled, don't pass default account

### LINKED ISSUE #2615: Order book shows mySize but no user is logged in
no user is logged in, don't use default user by provider.

bthaile:
fixes https://github.com/augurproject/augur/issues/2615

doesn't get default account if user isn't logged in


andrewdamelio (approved):
LGTM

PR #2669, opened 2019-07-06
fix accum shares calc and add spread to order book getter

### LINKED ISSUE #2584: add spread to order book
```
  const spread = () => {
    const bestBid =
      marketDepth[BIDS].length > 0 &&
      createBigNumber(`${marketDepth[BIDS][0][1]}`, 10);
    const bestAsk =
      marketDepth[ASKS].length > 0 &&
      createBigNumber(`${marketDepth[ASKS][0][1]}`, 10);

    return marketDepth[ASKS].length === 0 || marketDepth[BIDS].length === 0
      ? null
      : createBigNumber(bestAsk).minus(createBigNumber(bestBid));
  };

```

this is the calc the UI does to get the spread. In terms take the highest bid price and lowest ask price and subtract big number by little number.

### LINKED ISSUE #2667: Cumulative shares is incorrect in order book
image

Screen Shot 2019-07-05 at 12 52 54 PM

UI is showing the wrong information.

bthaile:
fixes https://github.com/augurproject/augur/issues/2667
fixes https://github.com/augurproject/augur/issues/2584

fixed `cumulativeShares` in order book getter. Issue was groupBy messed up sort,  moved sorting after groupBy.

added spread to order book, needs to be part of the getter, moved logic from UI.

added `no orders` in order as messaging to user that the order book has loaded but no orders.

image


packages/augur-sdk/src/state/getter/Markets.ts
```diff
@@ -468,15 +469,20 @@ export class Markets {
...
       isbids: boolean = false
     ): OrderBook[] => {
-      const orders = Object.values(unsortedOrders).sort((a, b) =>
-        isbids
-          ? new BigNumber(b.price).minus(a.price).toNumber()
-          : new BigNumber(a.price).minus(b.price).toNumber()
-      );
+      const orders = Object.values(unsortedOrders);
       const buckets = _.groupBy<Order>(orders, order => order.price);
+      const keysSorted: string[] = isbids
```
pgebheim: I think this should by DRYd up a bit, it took me a couple reads

pgebheim (commented):
Are we really doing all the BN conversions inline like this again?

I find this style to lead to absurdly verbose code that is very hard to follow.

@justinbarry @adrake33 @achapman

In v1 we used a method that would bignumberize all fields when reading from the DB so payloads in the getter context were pre converted.

packages/augur-sdk/src/state/getter/Markets.ts
bthaile: updated to pull out the confusing part into a new method.

andrewdamelio (approved):
UI changes look good @bthaile

PR #2672, opened 2019-07-07
Add additional topics to blockstream listener

justinbarry:
Also ran tslint fix.

@nuevoalex There is still a type issue here. We are storing additional topics as an array of arrays. Can they be flattened or do they need to be maintained?

### DIFF
--- a/packages/augur-sdk/src/state/db/BlockAndLogStreamerListener.ts
+++ b/packages/augur-sdk/src/state/db/BlockAndLogStreamerListener.ts
@@ -1,17 +1,33 @@
-import { EthersProviderBlockStreamAdapter, ExtendedLog } from "blockstream-adapters";
-import { Block, BlockAndLogStreamer, Log as BlockStreamLog } from "ethereumjs-blockstream";
-import { Filter } from "ethereumjs-blockstream/output/source/models/filters";
-import { EthersProvider } from "@augurproject/ethersjs-provider";
-import { Log, ParsedLog } from "@augurproject/types";
-import { EventLogDBRouter } from "./EventLogDBRouter";
+import {
+  EthersProviderBlockStreamAdapter,
+  ExtendedLog,
+} from 'blockstream-adapters';
+import {
+  Block,
+  BlockAndLogStreamer,
+  Log as BlockStreamLog,
+} from 'ethereumjs-blockstream';
+import { Filter } from 'ethereumjs-blockstream/output/source/models/filters';
+import { EthersProvider } from '@augurproject/ethersjs-provider';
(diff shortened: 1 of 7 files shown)


nuevoalex (approved):
Looks good. Are there any tests for blockstream yet?

justinbarry:
After speaking with @MicahZoltu it became clear we need to "de-duplicate" the filter sent to blockstream to minimize eth node requests and unpack/filter the payload on the callback side.

Also, this PR shows proves I don't know how additional topics work. The nested arrays in the filter serve as an OR mechanism. I'm flattening them all here.

packages/augur-sdk/src/state/db/BlockAndLogStreamerListener.ts
```diff
@@ -1,77 +1,122 @@
...
+  ) => string;
 }
 
 export interface BlockAndLogStreamerListenerDependencies {
   address: string;
   blockAndLogStreamer: BlockAndLogStreamerInterface<Block, ExtendedLog>;
   getBlockByHash: (hashOrTag: string) => Promise<Block>;
   eventLogDBRouter: EventLogDBRouter;
-  getEventTopics: (eventName: string) => Array<string>;
+  getEventTopics: (eventName: string) => string[];
```
MicahZoltu, line 38: I read at some point somewhere that `Array<string>` is the preferred/recommended syntax here, `string[]` is supported for legacy/backward compatibility reasons or something.

packages/augur-sdk/src/state/db/BlockAndLogStreamerListener.ts
```diff
@@ -1,77 +1,122 @@
...
     const dependencies = new EthersProviderBlockStreamAdapter(provider);
-    const blockAndLogStreamer = new BlockAndLogStreamer<Block, ExtendedLog>(dependencies.getBlockByHash, dependencies.getLogs, (error: Error) => {
-      console.error(error);
-    });
+    const blockAndLogStreamer = new BlockAndLogStreamer<Block, ExtendedLog>(
+      dependencies.getBlockByHash,
+      dependencies.getLogs,
+      (error: Error) => {
+        console.error(error);
+      }
```
MicahZoltu, line 78: You can just have the parameter here be `console.error`, no need to wrap it in a lambda.

MicahZoltu:
Recommend splitting this into multiple PRs, one that does the formatting changes and another that does the actual code changes.  With the two mixed, it is difficult to identify where the changes are.

justinbarry:
Already done in another PR. We cannot really merge multiple filters in an advantageous way.

PR #2674, opened 2019-07-09
turn on augur status, backend still needs to fix some issues with get…

bthaile:
wire up augur status, getting new block info from backend. Still an issue on backend where last processed block is zero. They are working through that issue.

image


andrewdamelio (approved):
LGTM

Issue #2559, opened 2019-06-24
categories selection component


phoebemirman:
![image](https://user-images.githubusercontent.com/3970376/60460273-e3fa1680-9c08-11e9-967f-1a54495f13c5.png)

<https://www.figma.com/file/IpIAYVQPeQfuT5TTUboQpo4S/Market-creation?node-id=1300%3A928>
link to category design

JohnDanz:
This issue seems to share a common component with this issue: https://github.com/AugurProject/augur/issues/2560

JohnDanz:
```
Categories Structure

Primary => Secondary => Tertiary

Primary:
  - Required
  - has a set list (which will dictate Secondary set list)
  - Allows custom input (no autocomplete)

Secondary:
  - Required
  - can have a set list (dictated from Primary selection from set list, if Primary was custom then no set list for Secondary)
  - Allows custom input (with autocomplete)

Tertiary:
  - Optional
  - can have a set list (dictated from Secondary selection from set list, if Secondary was custom then no set list for Tertiary)
  - Always custom input (with autocomplete)
```

JohnDanz:
@pgebheim @adrake33 ☝️ take a look at this categories structure, i think this would be helpful to do throughout and we can then add it to docs

bthaile:
category structure will fit in the create market API call and extraInfo structure.

PR #2676, opened 2019-07-09
Categories select

JohnDanz:
https://github.com/AugurProject/augur/issues/2559

no mobile designs yet, so punted on that. Might end up refactoring this into a container when i revisit for mobile styling


phoebemirman (approved):
looks good. for the future don't think  this should have all this height here unless the other dropdowns are showing
Screen Shot 2019-07-09 at 1 19 48 PM

PR #2684, opened 2019-07-09
Fix bad accounts request on kovan

justinbarry:
Run
```bash
ETHEREUM_NETWORK=kovan yarn workspace @augurproject/ui dev
```

and open in an incognito window.


andrewdamelio:
@justinbarry will test this very soon

packages/augur-ui/src/services/initialize.ts
```diff
@@ -6,25 +6,18 @@ import { windowRef } from "utils/window-ref";
...
 
   // Use injected provider if returning User has a unlocked injected account that matches the current logged in account
   if (injectedAccount && (loggedInAccount === injectedAccount[0])) {
-    provider = new Web3Provider(windowRef.web3.currentProvider);
-    account = windowRef.web3.currentProvider.selectedAddress,
-    isWeb3 = true;
+    const provider = new Web3Provider(windowRef.web3.currentProvider);
+    const account = windowRef.web3.currentProvider.selectedAddress;
+
+    await augurSdk.makeApi(provider, account, provider.getSigner(), env, false);
```
andrewdamelio, line 17: question... what is the isWeb3 parameter suppose to denote?

I thought it was to indicate providers that were injected, vs providers we need to bootstrap (hardware wallets, non-logged-in users)

If that correct? If so, then line 17 should pass in TRUE and line 19 should pass in false.

andrewdamelio (approved):
Tested with variety of wallets types, looks good

packages/augur-ui/src/services/initialize.ts
justinbarry: You are right! Will push in a few.

PR #2685, opened 2019-07-09
Remove augur.js and augur-node

bthaile:
remove augur.js and augur-node code from monorepo. only think left is documentation. which will be in a different PR once we figure out what needs to be documented.


justinbarry (approved):
LGTM

PR #2686, opened 2019-07-10
fix order book for canned markets

bthaile:
The UI trades on outcome 2 for YesNo and Scalar markets. For the canned market the order book it is more convenient if orders are on outcome 2.


robert-e-davidson3 (approved):
sorry about that!

PR #2693, opened 2019-07-10
filter profit and loss for calculating position, need all latest to calc frozen funds

bthaile:
`getProfitLossRecordsByMarketAndOutcome` is pulling in profit and loss results from changes in user's frozen funds. Those records need to be filtered out when calculating market outcome positions.

but all those profit and loss results are needed to calculate frozen funds


packages/augur-sdk/src/state/getter/Users.ts
```diff
@@ -584,7 +597,7 @@ async function getProfitLossRecordsByMarketAndOutcome(
   account: string,
   request: PouchDB.Find.FindRequest<{}>
 ): Promise<_.Dictionary<_.Dictionary<ProfitLossChangedLog[]>>> {
-  const profitLossResult = await db.findProfitLossChangedLogs(account, request);
+  let profitLossResult = await db.findProfitLossChangedLogs(account, request);
```
adrake33: This can be left as `const` instead of `let`.

adrake33 (approved):
LGTM! Just had one minor comment.

PR #2695, opened 2019-07-10
Prep work to add flash scripts to browser

robert-e-davidson3:
A lot of work still needs to be done to make it viable to run the flash scripts in the browser. Much of the node-only code has been separated out but the reliance on augur-core means there's still a lot left.


justinbarry (approved):
Very cool! Add auto-connect to local eth node in interactive mode and otherwise it LGTM.

PR #2701, opened 2019-07-11
Wire up trading events

bthaile:
initial wiring up of order events, handles create, cancel and fill. Also fixed issue where order book doesn't show up. A property that didn't exist `market.marketId` was being used. Since we are use to using it, I added it in market selector. 

also added new property for default selected outcome for market view page.

Order events come in sometimes and stop coming in. This PR will help debugging with backend guys.


andrewdamelio (approved):
LGTM

PR #2707, opened 2019-07-11
order book returning cancelled orders

### LINKED ISSUE #2706: getOrders returns cancelled orders
The order book passes 

```
    const orders = await Trading.getOrders(augur, db, {
      ...params,
      orderState: OrderState.OPEN,
    });
```

and gets Cancelled orders. 

Issue looks like "Open" means amount greater than zero
```
      request.selector = Object.assign(request.selector, {
        amount: { $gt: '0x00' },
      });
```
this isn't correct b/c amount doesn't change `amountFilled` increases.
need to filter out orderType === 1

bthaile:
don't allow cancelled orders to be returned when order status param is OPEN in getOrders. Order book was returning cancelled orders. There is an existing tix to handle partially filled orders not showing up in order book. the full amount is showing up

fixes https://github.com/AugurProject/augur/issues/2706


adrake33:
LGTM 👍

adrake33 (approved):

Issue #2702, opened 2019-07-11
Unable to add new block error in DerivedDB


bthaile:
![image](https://user-images.githubusercontent.com/3970376/61058472-edc80a80-a3bc-11e9-9e6c-345f314ebb49.png)

After taking all the asks orders on a binary market and getting MM confirmed got this error.

tried to capture the stack trace
<https://gist.github.com/bthaile/52724870444d28565ae34409e8c7f0b3>

node from Alex in discord

*"Ah I see what it is. There really needs to be a queue of some sort for the DerivedDB handleMergeEvent function
Issue is that its handling MarketVolumeChanged and MarketOIChanged events at the same time which for this case both have the same document id (the market id) and its conflicting"*

PR #2712, opened 2019-07-11
be more efficient during bulk and fix bug

nuevoalex:


packages/augur-sdk/src/state/db/DerivedDB.ts
```diff
@@ -65,15 +67,19 @@ export class DerivedDB extends AbstractDB {
...
+      success = await this.bulkUpsertUnorderedDocuments(documents);
     }
 
-
     if (success) {
       if (!syncing) {
+        // The mutex behavior here is needed since a derived DB may be responding to multiple updates in parallel
+        while (this.updatingHighestSyncBlock) {
+          await sleep(10);
+        }
```
robert-e-davidson3, line 108: It's possible but unlikely that two threads could get past the while-loop at the same time. You can use the mutex library since it uses a queue: https://github.com/DirtyHairy/async-mutex

packages/augur-sdk/src/state/db/DB.ts
```diff
@@ -222,6 +222,7 @@ export class DB {
       );
     }
 
+    console.log(`Syncing generic log DBs`);
```
robert-e-davidson3: Should probably put this just before populating the lists, since errors can pop up in early promises while the others are still being invoked. Same with the other message below.

packages/augur-test/src/tests/state/MarketDerivedDB.test.ts
```diff
@@ -18,6 +18,50 @@ beforeAll(async () => {
   augur = await makeTestAugur(ACCOUNTS);
 }, 120000);
 
+test("Doc merge update", async () => {
+  const db = await mock.makeDB(augur, ACCOUNTS);
+  let DBName = mock.constants.networkId + "-MarketCreated";
```
robert-e-davidson3, line 23:
```suggestion
  let DBName = `${mock.constants.networkId}-MarketCreated`;
```

packages/augur-test/src/tests/state/MarketDerivedDB.test.ts
```diff
@@ -18,6 +18,50 @@ beforeAll(async () => {
...
+    {
+      _id: "0x1111111111111111111111111111111111111111",
+      blockNumber: 2,
+      market: "0x1111111111111111111111111111111111111111",
+      marketOI: "0x2",
+    },
+  ];
+  await db.addNewBlock(DBName, OIBlockLogs);
+
+  await db.sync(augur, mock.constants.chunkSize, 0);
```
robert-e-davidson3: I can use `mock.constants.blockstreamDelay` instead of `0`.

robert-e-davidson3 (approved):
lgtm

packages/augur-sdk/src/state/db/DerivedDB.ts
robert-e-davidson3: This turned out not to be a concern due to how javascript implements its pseudo-multithreading.

bthaile (approved):
lgtm

bthaile:
Will fix this `unable to add new block` issue 
https://github.com/AugurProject/augur/issues/2702

justinbarry:
Added #2720 to add tests for rollback.

justinbarry (approved):
LGTML

PR #2713, opened 2019-07-11
add visibility component

phoebemirman:


packages/augur-ui/src/modules/create-market/components/review.styles.less
```diff
@@ -5,8 +5,9 @@
     flex-direction: column;
 
     > div:first-of-type {
-	    grid-template-columns: repeat(3, 1fr);
-    	display: grid;
+        grid-template-columns: repeat(3, 1fr);
+        display: grid;
+        grid-row-gap: 2rem;
```
JohnDanz: should be @size-32

JohnDanz (changes requested):
looks good, just change that one 2rem

JohnDanz (approved):

PR #2715, opened 2019-07-11
Negate boolean values as they were backwards

justinbarry:
@andrewdamelio Missed this on the other PR. It isn't critical but something to be fixed.


bthaile (approved):
lgtm

PR #2719, opened 2019-07-11
need to add back in condition amount > 0 

bthaile:
made a mistake removing check for order amount > 0 when they are filled.


adrake33 (approved):
Looks good!

PR #2751, opened 2019-07-15
No longer consider unlocked accounts when asking for default address

### LINKED ISSUE #2750: When running against an unlocked node the UI can pass in the wrong account address

justinbarry:
Close #2750


bthaile (approved):
looks magnificent.

PR #2758, opened 2019-07-15
Validations for step 1 

phoebemirman:
added validations for step 1 to see how validations structure should work


packages/augur-ui/src/modules/common/form.tsx
```diff
@@ -140,6 +141,27 @@ export const TimezoneDropdown = (props: TimezoneDropdownProps) => (
...
 );
 
+interface ErrorProps {
+  header?: string;
+  subheader?: string;
+}
+
+export const Error = (props: ErrorProps) => (
+  <section className={Styles.ErrorLabel}>
+    <div>
```
JohnDanz: instead of a div here, can we just put the svg in there? do we have to wrap it?

packages/augur-ui/src/modules/common/form.tsx
```diff
@@ -140,6 +141,27 @@ export const TimezoneDropdown = (props: TimezoneDropdownProps) => (
...
+      {ExclamationCircle}
+    </div>
+    <div>
```
JohnDanz, line 152: same here, do we really need this div her?

JohnDanz (changes requested):
a few small changes. I didn't explicitly mention it below but i think categories array has to be an empty array of strings, not sure that we want to put nulls in there... maybe something to discuss which is why i didn't bother making it it's own comment

JohnDanz (approved):

PR #2759, opened 2019-07-15
closes #2690 - debounces NewBlock events

### LINKED ISSUE #2690: NewBlock is firing 24 times each block
NewBlock event is coming into the UI 24 times a block

brianosaurus:
Simple method to throttle NewBlock events


bthaile (approved):
lgtm

PR #2762, opened 2019-07-15
Fixes new block event debouncing

brianosaurus:
Now throttles calls.
Makes sure the latest block is accurate (was a bug before)


justinbarry (approved):
LGTM

PR #2768, opened 2019-07-16
Fix filled amount

### LINKED ISSUE #2708: getOrders doesn't have filledAmount populated
Order book has old order amounts even after partial amount of order was filled.

Screen Shot 2019-07-11 at 11 19 56 AM

see above that the order is FILLED but the filledAmount is 0x00

1. login as non-contract owner
2. make sure to approve
3. make sure to get dai
4. partially fill an order from order book
5. notice that in Trading.ts `getOrders` the order that was partially filled has the correct status of orderType 3, but the filledAmount isn't populated it's still `0x00`

market trading history is correct. Order book didn't subtract the 5 taken at 0.28
Screen Shot 2019-07-11 at 11 24 46 AM

### LINKED ISSUE #2755: SDK: getOrders doesn't return user's open orders
steps to repro:

1. login with non-contract owner
2. approve cash so user can trade and get dai using faucet
3. go to market trading page
4. create order that will get put on the order book and not make a trade
5. sign tx, verify that the tx goes through (meta mask confirmation pop up)
6. notice that the user's open order doesn't show up in the `open orders` section

reason is `getOrders` in Trading.ts doesn't return the user's open orders. If you clear application data and refresh the UI it will show up.

adrake33:
Closes https://github.com/AugurProject/augur/issues/2708
Closes https://github.com/AugurProject/augur/issues/2755


bthaile (approved):
lgtm

PR #2769, opened 2019-07-16
update-set-categories-and-dropdowns

JohnDanz:
updated dropdown and set-categories - fixes issue with categories getting default selections when they should be just a placeholder label at first


phoebemirman (changes requested):
not working for custom inputs

phoebemirman (approved):

PR #2771, opened 2019-07-16
fix order form issue where user changes price after clicking on order…

### LINKED ISSUE #1911: order book bug
Screen Shot 2019-04-25 at 5 02 41 PM

i think this happened when pressing backspace in the price field!!!!

valid order but the button doesn't not light up

1. clicked on order in the book (that was mine)
2. increased price
3. then switch to sell 

happened in a few different cases, not sure the other reproductions

https://try.augur.net/#!/market?id=0x599c4cdee0dc31b7c21e89c3352bc1b6afb40ed6
clicked on the order and then did backspace and typed .4

bthaile:
… book order

fixes https://github.com/AugurProject/augur/issues/1911

Issue was order form didn't allow for `lastInputModified` variable to be empty when trying to calc order form values.

also noticed that market-view can throw if market hasn't been loaded, need to default selected outcome id. 

also saw that user open orders is filtering on order status, FILLED. since we get user open orders specifically we don't need to filter. Also, FILLED means partially filled.


andrewdamelio (approved):
LGTM

PR #2772, opened 2019-07-16
fixed direct load of market trading page

bthaile:
there was an issue where the "loading ..." modal didn't go away if the user navigates directly to a trading page. 

Fixed issue by using `isConnected` set when sync'd to sdk. before then `isConnected` is false.


packages/augur-ui/src/modules/market/components/market-view/market-view.tsx
```diff
@@ -260,7 +262,8 @@ export default class MarketView extends Component<MarketViewProps, MarketViewSta
       );
     }
 
-    const outcome = outcomes.find( outcomeValue => outcomeValue.id === s.selectedOutcomeId)
+    const outcomeId = s.selectedOutcomeId ? s.selectedOutcomeId : market.defaultSelectedOutcomeId;
```
bthaile, line 261: if market is lazy loaded need to set default selected outcome id.

andrewdamelio: 👍

andrewdamelio (approved):

PR #2775, opened 2019-07-16
Group all non major renovate upgrades

justinbarry:


petong (approved):
sweet, lets see how well this works

PR #2777, opened 2019-07-17
Add outcome to getMarketsInfo

### LINKED ISSUE #2749: Move calculate-payout-numerators-value.ts to middleware & use it to get outcome in getMarketsInfo

adrake33:
Closes https://github.com/AugurProject/augur/issues/2749


packages/augur-ui/src/modules/alerts/actions/set-alert-text.ts
```diff
@@ -255,9 +254,11 @@ export default function setAlertText(alert: any, callback: any) {
             loadMarketsInfoIfNotLoaded([alert.to], () => {
               const marketInfo = selectMarket(alert.to);
               const outcome = calculatePayoutNumeratorsValue(
-                marketInfo,
-                alert.params._payoutNumerators,
-                alert.params._invalid
+                marketInfo.maxPrice,
```
bthaile, line 257: These are display prices. Is that correct for calculatePayoutNumeratorsValue method?

bthaile (approved):
Lgtm if sdk calculate payout numerator value takes display price.

packages/augur-ui/src/modules/alerts/actions/set-alert-text.ts
adrake33: In the original function in augur-ui, the first line was: ```  const { maxPrice, minPrice, numTicks, marketType } = market;```
so I think it should be ok to pass these in.

bthaile: Not a question about passing them. Its a question of are they display values or on chain values. 
Since both the sdk and ui are calling it. I wanted to confirm that display values ate passed in from the sdk. Ui already uses display prices.

adrake33: Good point, the getter is passing in the value from the log, so it's not using the display price. 
 Should we make it accept on-chain values instead?

bthaile: That is the challenge, prob need to rename the method params to indicate if onChain or display so we dont make mistakes on these type of helper methods. UI only has display values. Could do this:

Have two different methods one for onchain and the other for display 
Or just take in display values. What ever you think fix best in the sdk.

adrake33: Actually, seems easier to keep it as display price since theres a conversion function from onchain to display in the middleware.

I'll rename the input params to be clearer.

bthaile: That makes the most sense to me.

PR #2789, opened 2019-07-17
commenting out until we can figure out what is causing errors.

bthaile:
weird error in the UI with not being able to import util functions from sdk.
also there were references missed when createBigNumber was moved to sdk. re-created create-big-number file.


andrewdamelio (approved):

adrake33 (commented):
load-universe-info.ts was also edited. Not sure if it needs updating or not.

adrake33 (approved):
lgtm

bthaile:
found the issue with sdk, closing this PR

PR #2792, opened 2019-07-17
wait until connected to load filtered markets

bthaile:
fix initial load of markets list page. sometimes UI tries to get markets when the sdk hasn't processed logs yet.


andrewdamelio (approved):
looks good

PR #2793, opened 2019-07-17
use cumulativeShares instead of shares in order book

### LINKED ISSUE #2787: order book accumulated shares not correct
user created order of quantity 1 at a price between existing.  Refreshed the UI and nav to the market and saw this malformed order book

bthaile:
image

fixes https://github.com/AugurProject/augur/issues/2787

image


bthaile:
not an issue, it's a feature, closing

PR #2810, opened 2019-07-18
Can now manage seeds with the flash scripts.

robert-e-davidson3:


packages/augur-tools/src/libs/ganache.ts
```diff
@@ -2,23 +2,36 @@ import * as ganache from "ganache-core";
...
 import memdown from "memdown";
 import { MemDown } from "memdown";
+import { Contracts as compilerOutput } from "@augurproject/artifacts";
 
 import { Account } from "../constants";
+import crypto from "crypto";
+import { EthersProvider } from "@augurproject/ethersjs-provider/build";
+const levelup = require("levelup");
+import * as path from "path";
+import { ContractAddresses } from "@augurproject/artifacts/build";
```
bthaile: are we standardizing to use top level references

packages/augur-tools/src/libs/ganache.ts
```diff
@@ -2,23 +2,36 @@ import * as ganache from "ganache-core";
 import { ethers } from "ethers";
 import memdown from "memdown";
 import { MemDown } from "memdown";
+import { Contracts as compilerOutput } from "@augurproject/artifacts";
 
 import { Account } from "../constants";
+import crypto from "crypto";
+import { EthersProvider } from "@augurproject/ethersjs-provider/build";
```
bthaile: are we standardizing to use top level references

packages/augur-test/src/tests/api/Contracts.test.ts
```diff
@@ -1,6 +1,6 @@
-import { makeProvider, seedPath } from "../../libs";
+import { makeProvider } from "../../libs";
 import { Contracts } from '@augurproject/sdk/build/api/Contracts';
```
bthaile: are we standardizing to use top level references

bthaile (approved):
lgtm

packages/augur-tools/src/libs/ganache.ts
robert-e-davidson3: good catch!

PR #2825, opened 2019-07-19
Closes #2808 - all tracked users for db create

### LINKED ISSUE #2808: transaction events aren't emitted to UI on kovan
This seems to be an issue locally.

load http://v2.augur.net
1. make sure user is approved to trade
2. grab some dai from faucet
3. navigate to market to trade
3a. verify `Subscribing to Augur events` shows up in web console
4. take order off order book
5. create order to trade
6. in web console there should be console logs indicating `AwaitingSigning Transaction` and `TxPending Transaction` ....
and a order in the user's open orders section with `pending` wait indicator. none of those happen.

I've cleared Application data (indexdb) and tried again, still can't get transaction events. NewBlock event does come in but then stops coming in. 

seeing <https://github.com/AugurProject/augur/issues/2704> `Document update conflict` this could be the main issue.

brianosaurus:
Only one account is passed into create-api methods. That account was turning into trackedUsers but there are more than one in the TrackedUsers db. So...create would create one db but rollback/sync would get all of the users and try to rollback/sync. This caused an exception and events wouldn't fire.

### DIFF
--- a/packages/augur-sdk/src/state/db/DB.ts
+++ b/packages/augur-sdk/src/state/db/DB.ts
@@ -135,20 +135,25 @@ export class DB {
     this.blockAndLogStreamerListener = blockAndLogStreamerListener;
 
     // Create SyncableDBs for generic event types & UserSyncableDBs for user-specific event types
-    for (let eventName of this.genericEventNames) {
+    for (const eventName of this.genericEventNames) {
       new SyncableDB(this.augur, this, networkId, eventName, this.getDatabaseName(eventName), []);
     }
 
-    for (let derivedDBConfiguration of this.basicDerivedDBs) {
+    for (const derivedDBConfiguration of this.basicDerivedDBs) {
       new DerivedDB(this, networkId, derivedDBConfiguration.name, derivedDBConfiguration.eventNames, derivedDBConfiguration.idFields);
     }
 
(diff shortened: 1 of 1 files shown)


adrake33 (approved):
Lgtm

bthaile (approved):
Lgtm

packages/augur-sdk/src/state/db/TrackedUsers.ts
```diff
@@ -7,9 +7,8 @@ export class TrackedUsers extends AbstractDB {
     super(networkId, networkId + "-TrackedUsers", dbFactory);
   }
 
-  async setUserTracked(user: string): Promise<PouchDB.Core.Response> {
+  async setUserTracked(formattedUser: string): Promise<PouchDB.Core.Response> {
```
bthaile: Does the UI need to specifically add user addresses or does middleware just add them ad the user changes accounts

brianosaurus: @bthaile UI needs to initialize another sdk/augur with the new address.

bthaile: that is what the UI is doing it destroys old SDK object and creates a new one.

PR #2830, opened 2019-07-19
draft and validations polish

phoebemirman:
- don't show draft saving modal when only validations are in newMarket
- don't compare validation differences to tell whether draft is saved
- show error when they save draft without a market description
- make sure validations get cleared when newMarket gets cleared
- clear min and max errors when min or max gets changed


bthaile (approved):
lgtm

PR #2866, opened 2019-07-22
add outcomes for different market types

phoebemirman:
scalar, yes no, categorical outcomes display

### DIFF
--- a/packages/augur-ui/src/modules/market-cards/common.tsx
+++ b/packages/augur-ui/src/modules/market-cards/common.tsx
@@ -22,7 +22,7 @@ export const Percent = (props: PercentProps) => (
 
 export interface OutcomeProps {
   name: string;
-  percent: number;
+  percent?: number;
   invalid?: Boolean;
   index: number;
 }
@@ -41,7 +41,7 @@ export interface ScalarOutcomeProps {
   scalarDenomination: string;
   min: number;
   max: number;
-  percent: number;
+  percent?: number;
 }
 
 function calculatePosition(min, max, percent) {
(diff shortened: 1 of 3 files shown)


packages/augur-ui/src/modules/market-cards/common.tsx
```diff
@@ -0,0 +1,111 @@
...
+    </div>
+    <Percent percent={props.percent} />
+  </div>
+);
+
+export interface ScalarOutcomeProps {
+  scalarDenomination: string;
+  min: number;
+  max: number;
+  percent: number;
```
bthaile: Need a ‘?’ So percent? To indicate it can be null

packages/augur-ui/src/modules/market-cards/common.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+export const Percent = (props: PercentProps) => (
+  <div className={Styles.Percent}>
+  	<span style={{width: props.percent + "%"}}></span>
+  </div>
+);
+
+export interface OutcomeProps {
+  name: string;
+  percent: number;
```
bthaile: Need ‘?’ percent? To indicate it can be null.

packages/augur-ui/src/modules/market-cards/common.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+	return percent === null
+	  ? 50
+	  : pricePercentage
+}
+
+export const ScalarOutcome = (props: ScalarOutcomeProps) => (
+  <div className={Styles.ScalarOutcome}>
+  	<div>
+  		<span style={{left: calculatePosition(props.min, props.max, props.percent) + '%'}}>{props.percent}</span>
```
bthaile: Dont use percent ‘%’ for scalar just the number. If percent which is actually last trade price is null then put’-‘

phoebemirman: this is for the style/ how much to move the dot over

bthaile (approved):
Ok, sounds good.

Issue #2885, opened 2019-07-23
make learn more open a modal


phoebemirman:
"Each section of the form has a "Learn more" link that will open a modal with help content" is it all linking to the same modal or different modals?

PR #2886, opened 2019-07-23
add info modal

phoebemirman:
https://github.com/AugurProject/augur/issues/2885

mobile styling


bthaile (approved):
Lgtm, if need be we can generic’ize the info modals and link more, looks good for now.

PR #2887, opened 2019-07-23
Change percent behind current to percent synced - Fix DerivedDB, upsert fix 

brianosaurus:
This PR attempts to fix document upsert conflicts however they can still happen.

Also, change percentBehindCurrent to percentSynced which is more of a correct name.

Fixes DerivedDB which wasn't setting synced to false at the right time. This, in turn, gets the UI to show the right status.


justinbarry (approved):
LGTM

PR #2896, opened 2019-07-24
Use upsert pouchdb plugin to fix 409 errors

justinbarry:


packages/augur-sdk/src/state/db/AbstractDB.ts
```diff
@@ -43,15 +46,17 @@ export abstract class AbstractDB {
     }
   }
 
-  protected async upsertDocument(id: string, document: object): Promise<PouchDB.Core.Response> {
+  protected async upsertDocument(id: string, document: object): Promise<PouchDB.UpsertResponse> {
```
pgebheim, line 49: I don't *love* this implementation, as it will overrwite things if the document that is passed in has fields that are changed by another writer.

The reason it takes a thunk is so that when retrying you get the current version of the document to modify, so that your logic for determining how to do the update is fully run against the current version.

Now the basic version will probably work for us in many cases, but since I haven't seen all the call sites I would prefer that this not be the default behavior, and instead it should take a thunk like `upsert`.

pgebheim (dismissed):

packages/augur-sdk/src/state/db/AbstractDB.ts
brianosaurus: @pgebheim Turns out this conflict only happens with SyncStatus updates. I think the reason is that for all other things, it's an insert. SyncStatus is the only time we're modifying a record.

So...while I understand your concern here....IMNSHO it falls into the 'done not better' category. 

Obviously up for debate @justinbarry @pgebheim. I'm happy to modify this myself and make it pull a new doc and merge the two.

packages/augur-sdk/src/state/db/AbstractDB.ts
```diff
@@ -43,15 +46,17 @@ export abstract class AbstractDB {
...
     const previousBlockRev = await this.getPouchRevFromId(id);
     const obj = Object.assign(
       previousBlockRev ? { _rev: previousBlockRev } : {},
       { _id: id },
       document
     );
 
-    return this.db.put(obj);
+    return this.db.upsert(id, () => {
```
nuevoalex, line 51: Do we still need all of the above logic since we're letting a plugin handle the revision / update machinery?

packages/augur-sdk/src/state/db/AbstractDB.ts
```diff
@@ -60,7 +65,7 @@ export abstract class AbstractDB {
       result[prevDoc._id] = prevDoc;
       return result;
     }, {} as DocumentIDToDoc);
-    return await this.bulkUpsertDocuments(previousDocs, documents);
+    return this.bulkUpsertDocuments(previousDocs, documents);
```
nuevoalex: why remove await?

packages/augur-sdk/src/state/db/AbstractDB.ts
```diff
@@ -69,7 +74,7 @@ export abstract class AbstractDB {
       result[prevDoc.id] = prevDoc.doc;
       return result;
     }, {} as DocumentIDToDoc);
-    return await this.bulkUpsertDocuments(previousDocs, documents);
+    return this.bulkUpsertDocuments(previousDocs, documents);
```
nuevoalex: same question here

packages/augur-sdk/src/state/db/AbstractDB.ts
brianosaurus: @nuevoalex 

My editor tslnt's those things and tells me to remove them. Also, I googled it one day and ppl said to remove them since it's extra, unnecessary logic. I guess returning a promise gets the compiler to not add a promise.

packages/augur-sdk/src/state/db/SyncStatus.ts
```diff
@@ -25,12 +28,26 @@ export class SyncStatus extends AbstractDB {
...
-    const document: SyncDocument = { blockNumber, syncing };
-    return this.upsertDocument(dbName, document);
+  public async setHighestSyncBlock(dbName: string, blockNumber: number, syncing: boolean, rollback: boolean = false): Promise<PouchDB.UpsertResponse> {
+
+    const highestKnownBlock = await this.getHighestSyncBlock(dbName);
+
+    // NOTE: dbName, in this case, is actually the id of the record in the SyncStatus db.
+    return this.db.upsert(dbName, (document: SyncDocument) => {
+      // make sure the truly highest block is always being used
+      if (!rollback) {
```
nuevoalex, line 37: is this just a paranoia thing or would this happen normally?

packages/augur-sdk/src/state/db/AbstractDB.ts
brianosaurus: @nuevoalex see above answer.

packages/augur-sdk/src/state/db/SyncStatus.ts
brianosaurus: Paranoia and a suggestion by @pgebheim. In a browser this won't happen. In a server maybe.

packages/augur-sdk/src/state/db/AbstractDB.ts
brianosaurus: @nuevoalex Fixing. Good catch. Thanks!

nuevoalex (changes requested):
Had some questions but otherwise looks good!

nuevoalex (approved):
🆙

PR #2930, opened 2019-07-25
Smoke Test

robert-e-davidson3:
Created a system smoke test that runs the webapp using Puppeteer. It's currently disabled/skipped because we need to use the evm CREATE2 so that contract addresses are predictable. The test can be run locally by removed the`.skip` call and also running a local node.

Also there's some cleanup in this PR.


packages/augur-test/src/tests/flash/ReasonableMarket.test.ts
```diff
@@ -0,0 +1,21 @@
+import { ACCOUNTS, FlashSession, addScripts, addGanacheScripts } from "@augurproject/tools";
```
robert-e-davidson3, line 1: I renamed CannedMarkets to this because creating a single market is much faster than creating 20 plus many orders... and the original test was taking a very long time to run.

petong (approved):

PR #2934, opened 2019-07-25
Templates

JohnDanz:
This is some work toward templates, It's not a complete implementation


bthaile (approved):
lgtm for an initial check in. We'll flush out feature requirements more in the near future.

Issue #2933, opened 2019-07-25
Reporting: Wire up log handlers


bthaile:
Wire up log handlers, dispute information is on MarketInfo object. need to get latest information when reporting/disputing logs come in. Also need to update users assets. 

  
1. Disputing logs
 - [x] crowdsourcer completed
 - [x] crowdsourcer contribution
 - [x] crowdsourcer created
 - [x] crowdsourcer redeemed
2. Reporting logs
 - [x]  initial report submitted
 - [x] initial report redeemed
 - [x] initial report transferred
3. Dispute window logs
 - [x] dispute window created
4. participation tokens
 - [x] participation tokens redeemed
 
(token transferred) might get fired when user buys participation tokens (needs investigation)
markets have a dispute window associated to them. When market is in fast resolution then the dispute window will be 24 hour timeframe. When in slow resolution it's 7 day timeframe.

 - [x] remove old reporting-window-stats

pgebheim:
WE talked today about merging all this information into the Derived DB for faster lookups @justinbarry

PR #2948, opened 2019-07-25
Wire up report dispute log events

bthaile:
clean up event handlers, remove transactionsData. It's not needed anymore for constructing bell alerts. 

log-handlers will get simplified, for now update user's assets per block. I suspect there was an issue when multiple logs come in on the same block to caused calling update user's assets more than once per block.

user's assets is their ETH, REP and DAI balances.

just reference https://github.com/augurproject/augur/issues/2933

more work to do when middleware getters are ready


packages/augur-ui/src/modules/events/actions/log-handlers.ts
```diff
@@ -290,173 +237,165 @@ export const handleOrderFilledLog = (log: Logs.ParsedOrderEventLog) => (
...
-              linkPath: makePath(MY_MARKETS),
-            })
-          );
-        } else if (!doesntExist) {
-          handleAlertUpdate(log, dispatch, getState);
-        }
-      }
-    })
+) => {
+  console.log('handleProfitLossChangedLog');
```
andrewdamelio: can we remove this?

andrewdamelio (approved):
LGTM love the deletions

PR #2952, opened 2019-07-26
OICash

nuevoalex:
A wrapper around Cash (DAI) which will be counted as OI within the context of a Universe while held. Fees are owed upon withdrawal and additional fees can be contributed whenever.

The purposes of this mechanism is to support side chain Augur use cases where they would trade using this token and we would be accounting for OI appropriate therefore in their trading system.


packages/augur-ui/src/modules/alerts/actions/set-alert-text.ts
```diff
@@ -460,7 +459,7 @@ export default function setAlertText(alert: any, callback: any) {
       case GETORCACHEDESIGNATEDREPORTSTAKE:
         alert.title = "Get stake size required for desginated reports";
         break;
-      case GETORCACHEMARKETCREATIONCOST:
+      case getOrCacheValidityBond:
```
adrake33: Where is this case defined since GETORCACHEMARKETCREATIONCOST got removed? Also, should the title below be updated too?

packages/augur-core/source/contracts/libraries/token/ERC777.sol
```diff
@@ -222,7 +222,7 @@ contract ERC777 is IERC777, IERC20 {
     */
     function transferFrom(address holder, address recipient, uint256 amount) public returns (bool) {
         require(recipient != address(0), "ERC777: transfer to the zero address");
-        require(holder != address(0), "ERC777: transfer from the zero address");
+        require(holder != address(0));
```
bthaile, line 225: reason to remove the require messaging?

packages/augur-core/source/contracts/trading/OICash.sol
```diff
@@ -0,0 +1,77 @@
...
+import 'ROOT/IAugur.sol';
+import 'ROOT/libraries/Initializable.sol';
+import 'ROOT/libraries/token/VariableSupplyToken.sol';
+
+
+/**
+ * @title OI Cash
+ * @dev A Wrapper contract for the deployed Cash contract which Augur considers OI. Cash can be deposited and will count toward OI for reporting fee calculations and will extract a reporting fee on withdrawl
+ */
+contract OICash is VariableSupplyToken, Initializable {
```
adrake33: Shouldn't this be declared as IOICash? Otherwise, IOICash doesn't seem to be getting used.

adrake33 (approved):
Looks good overall. Just a few minor questions/comments.

packages/augur-core/source/contracts/trading/OICash.sol
adrake33: Never mind, it is getting used in OICashFactory.

packages/augur-core/source/contracts/trading/OICash.sol
```diff
@@ -0,0 +1,77 @@
...
+
+        return true;
+    }
+
+    function payFees(uint256 _feeAmount) external returns (bool) {
+        burn(msg.sender, _feeAmount);
+        uint256 _reportingFeeDivisor = universe.getOrCacheReportingFeeDivisor();
+        uint256 _openInterestAmount = _feeAmount.mul(_reportingFeeDivisor);
+        universe.withdraw(address(universe.getOrCreateNextDisputeWindow(false)), _feeAmount, address(0));
+        totalAmountFeesPaid = totalAmountFeesPaid.add(_openInterestAmount);
```
bthaile, line 67: does this need to return true?

bthaile (approved):
lgtm, 

I assume that the OICash just keeps track of OI from side chains. Do they get fees based on the fees they add? if so lgtm

packages/augur-core/source/contracts/libraries/token/ERC777.sol
nuevoalex: To reduce the contract size so Universe deployment is below an acceptable gas limit :[

packages/augur-ui/src/modules/alerts/actions/set-alert-text.ts
nuevoalex: ah this should just be removed. GETORCACHEMARKETCREATIONCOST is gone in favor of just asking for the validity bond

packages/augur-core/source/contracts/trading/OICash.sol
nuevoalex: Should still be IOICash though, will add

packages/augur-core/source/contracts/trading/OICash.sol
nuevoalex: yup!

Issue #1718, opened 2019-04-18
Mobile Trading Page: Reduce header sizes to 48px


matt-bullock:
Currently look bigger than the design

Left:Build — Right:Design
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/5cad9f22-4959-462a-8ec0-5df3df3d01b8)

PR #2957, opened 2019-07-26
fix-mobile-headers-trading-page

JohnDanz:
fixed some issues with header heights and converted some html to the proper tags (p => h3, div => h3)

https://github.com/augurproject/augur/issues/1718


bthaile (approved):
lgtm

PR #2996, opened 2019-07-30
Enable use of envvars for deploy

robert-e-davidson3:
Real vs fake time; production vs non-production; create genesis universe or not.


bthaile (approved):
lgtm

PR #3005, opened 2019-07-31
added initial report to flash scripts

### LINKED ISSUE #1891: Flash script: "initial-report"

bthaile:
fixes https://github.com/AugurProject/augur/issues/1891

added connect to sdk in `connect` script
added calls to sdk to get needed info to call initial report on Market contract.

to use, make sure to push time so that the market is in `designated reporting` or `open reporting` state.
>yarn flash interactive
>connect -u
>initial-report -m <marketid> -o <outcome>

### DIFF
--- a/packages/augur-sdk/src/Augur.ts
+++ b/packages/augur-sdk/src/Augur.ts
@@ -161,7 +161,7 @@ export class Augur<TProvider extends Provider = Provider> {
   }
 
   bindTo<R, P>(f: (db: any, augur: any, params: P) => Promise<R>): (params: P) => Promise<R> {
-    return Augur.connector.bindTo(f);
+    return Augur.connector && Augur.connector.bindTo(f);
   }
 
   async on(eventName: SubscriptionEventName | TXEventName | string, callback: Callback | TXStatusCallback): Promise<void> {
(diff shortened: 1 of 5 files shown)


packages/augur-sdk/src/Augur.ts
```diff
@@ -161,7 +161,7 @@ export class Augur<TProvider extends Provider = Provider> {
   }
 
   bindTo<R, P>(f: (db: any, augur: any, params: P) => Promise<R>): (params: P) => Promise<R> {
-    return Augur.connector.bindTo(f);
+    return Augur.connector && Augur.connector.bindTo(f);
```
justinbarry, line 164: @bthaile Why did you need to do this?

bthaile: for some reason bindTo was called when `connector` was null. it was in a test. also made sense to not `.bindTo` if connector doesn't exist.

justinbarry: Create needs to be called first. I wonder what we were doing where that wasn't happening.

justinbarry (approved):
LGTM

packages/augur-tools/src/flash/scripts.ts
```diff
@@ -344,9 +351,12 @@ export function addScripts(flash: FlashSession) {
       if (extraStake) {
         preEmptiveStake = new BigNumber(extraStake).dividedBy(QUINTILLION).toFixed();
       }
+      if (!this.usingSdk) this.log("This script needs sdk, make sure to connect with -u flag");
```
robert-e-davidson3, line 354: This doesn't leave this function so it's going to error later when trying to actually do things.

PR #3006, opened 2019-07-31
fix condition on default outcome

### LINKED ISSUE #2985: Error when user directly to market trading page
image

bthaile:
fixes https://github.com/AugurProject/augur/issues/2985

fixed issue with conditional. need to test for null and undefined.

### DIFF
--- a/packages/augur-ui/src/modules/market/components/market-view/market-view.tsx
+++ b/packages/augur-ui/src/modules/market/components/market-view/market-view.tsx
@@ -265,7 +265,7 @@ export default class MarketView extends Component<MarketViewProps, MarketViewSta
       );
     }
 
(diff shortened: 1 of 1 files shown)


packages/augur-ui/src/modules/market/components/market-view/market-view.tsx
```diff
@@ -265,7 +265,7 @@ export default class MarketView extends Component<MarketViewProps, MarketViewSta
       );
     }
 
-    let outcomeId = selectedOutcomeId !== null ? selectedOutcomeId : market.defaultSelectedOutcomeId;
+    let outcomeId = selectedOutcomeId ? selectedOutcomeId : market.defaultSelectedOutcomeId;
```
andrewdamelio: I think we will need to test specifically test for NULL UNDEFINED here

andrewdamelio (approved):

Issue #1989, opened 2019-05-02
Increase vertical hit area of text buttons on touch devices


matt-bullock:
Increase the vertical hit area of these buttons on touch devices to 24px . Currently, they're very small to tap

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/a2a447c2-d391-44b3-b468-97567fafcc63)

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/522f5cfa-bc58-491b-9e3c-89b159c3f830)

andrewdamelio:
@pgebheim we can move this one to the Mobile epic

bthaile:
https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=7268%3A326
in figma

pgebheim:
🎉

PR #3011, opened 2019-07-31
increase-text-touch-area-mobile

JohnDanz:
upped line height of SubmitTextbutton on mobile from 10 to 24px in order to increase surface area of touch events

https://github.com/augurproject/augur/issues/1989


bthaile (approved):
lgtm, that makes sense to just increase line height

Issue #2966, opened 2019-07-26
market creation errors


bthaile:
User logs in and navigates to create a market. The errors in the gist occur on the market creation landing page.
<https://gist.github.com/petong/1f94c2d56fbe4eaea5968af2368a2d40>

These two errors in the above gist will get fixed with this tix: https://github.com/AugurProject/augur/issues/2964
```
Invalid prop `availableEth` of type `number` 
Failed prop type: Invalid prop `availableRep` of type `number` supplied to `
```

This tix work these errors:
`Check the render method of `LocationDisplay`. See https://fb.me/react-warning-keys for more information.`
`Check the render method of `ExplainerBlock`. See https://fb.me/react-warning-keys for more information.`

PR #3021, opened 2019-07-31
remove warnings

phoebemirman:
https://github.com/AugurProject/augur/issues/2966


bthaile (approved):
Lgtm, i’ve created a tix to standardize how users balances are handled throughout the site. So formatted values are always available and components dont have to worry about it.

Issue #3023, opened 2019-07-31
Investigate category 3 indexing


adrake33:
Discussion on category 3 indexing is in https://github.com/AugurProject/augur/pull/3024

pgebheim:
Follow on to:
https://github.com/AugurProject/augur/issues/2864

PR #3024, opened 2019-07-31
Add new params to getMarkets

### LINKED ISSUE #2864: modified getMarkets getter
Market creation logs need to parse out categories from extraInfo
categories will come in as string array. order is primary, secondary and tertiary (optional).

Initial loading of market lists page. UI will call getter with default filters. 

Filters: 
market reporting states included
liquidity spread
market fee
invalid market (show | hide)
search term (optional)
category array (["primary", "secondary", "tertiary"] format)
sort by
limit: items per page
offset: page number

Sort options (should always be assumed to be descending order):

 * liquidity
 * open interest
 * volume
 * created timestamp
 * end date
 * last traded timestamp
 * last liquidity-depleted timestamp

return values:

UI needs:
 * market ids in filtered search
 * categories in filtered search (whole category tree)
 * market count in category (primary, secondary, tertiary
 ~~* market status counts (open, in reporting, resolved)~~ (no longer needed)
 * market count filtered out (only liq filter and trading fee apply)

 ```json
{
  "results": [
   ...MatchingMarketsInfo.....
  ],
  "meta": {
    "categories": [
      {
        "name": "sports",
        "count": 2,
        "children": [{
          "name": "soccer",
          "count": 2
        }]
      }
    ],
    "filteredOutCount": 10,
    "marketCount": 40,
  }
}
```

### PR DESCRIPTION
**Pull request by adrake33**
Closes https://github.com/AugurProject/augur/issues/2864. Remaining TODO items are in https://github.com/AugurProject/augur/issues/3023

### CODE REVIEW

File: packages/augur-sdk/src/state/db/MarketDB.ts [hunk 1]
```diff
@@ -36,43 +53,42 @@ interface LiquidityResults {
...
+          field: [
+            "market",
+            "category1",
+            "category2",
+            "category3",
+            "description",
+            "longDescription",
+            "resolutionSource",
+            "_scalarDenomination",
+            "tags",
```
bthaile: Tags should not exist now. Also can we add market creater

packages/augur-sdk/src/state/getter/Markets.ts
```diff
@@ -366,23 +381,37 @@ export class Markets {
     if (!(await augur.contracts.augur.isKnownUniverse_(params.universe))) {
       throw new Error('Unknown universe: ' + params.universe);
     }
+    const validLiquiditySpreads = ['10', '15', '20', '100'];
+    if (params.maxLiquiditySpread && !validLiquiditySpreads.includes(params.maxLiquiditySpread)) {
+      throw new Error('Invalid maxLiquiditySpread');
+    }
+    // Set sort defaults
+    if (typeof params.sortBy === 'undefined') {
+      params.sortBy = GetMarketsSortBy.OPEN_INTEREST;
```
bthaile: We might want default sort to be either volume or something else. Need to run this by @joeykrug and @Chwy5

bthaile (commented):
A few comments atthis point

packages/augur-sdk/src/state/getter/Markets.ts
joeykrug: Neither --- it'll be the liquidity sort, cc @nuevoalex

packages/augur-sdk/src/state/db/MarketDB.ts
adrake33: Good catch! I'll remove the tags and create another task for making marketCreator full-text-searchable.

packages/augur-sdk/src/state/getter/Markets.ts
adrake33: Sounds good. I've updated https://github.com/AugurProject/augur/issues/3023 to include modifying the default sort to be liquidity.

packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -19,6 +18,23 @@ import { MarketData, MarketType, OrderType } from "../logs/types";
...
+  market: string;
+  category1: string;
+  category2: string;
+  category3: string;
+  description: string;
+  longDescription: string;
+  resolutionSource: string;
+  _scalarDenomination: string;
+  start: Date;
+  end: Date;
```
bthaile, line 36: can we add market creator address, just in case users want to search on it, to see all the markets created by a address.

packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -36,43 +52,41 @@ interface LiquidityResults {
...
+      {
+        doc: {
+          id: "id",
+          start: "start",
+          end: "end",
+          field: [
+            "market",
+            "category1",
+            "category2",
+            "category3",
```
bthaile, line 81: what if category3 is empty will that mess up flex search?

packages/augur-sdk/src/state/getter/Markets.ts
```diff
@@ -35,12 +37,15 @@ const getMarketsParamsSpecific = t.intersection([
     creator: t.string,
     category: t.string,
     search: t.string,
-    reportingState: t.union([t.string, t.array(t.string)]),
+    reportingStates: t.array(t.string),
```
bthaile: shouldn't this be typed so that if a unknown reportingState is passed it it'll get caught and return an error?

packages/augur-sdk/src/state/getter/Markets.ts
```diff
@@ -414,106 +442,153 @@ export class Markets {
...
-          previousValue[fullTextSearchResult.market] = fullTextSearchResult;
-          return previousValue;
-        },
-        []
-      );
+    if (params.search || params.categories) {
+      let keyedFullTextResults: any = {};
+      let searchResults: any = [];
+      if (params.search) {
+        searchResults = await db.fullTextMarketSearch(params.search, null);
```
bthaile, line 444: Just a note we'll need to document wildcard character and any `advanced` search capabilities we get.

packages/augur-sdk/src/state/getter/Markets.ts
```diff
@@ -414,106 +442,153 @@ export class Markets {
...
+        keyedFullTextResults = _.keyBy(
+          searchResults,
+          (searchResult: MarketFields) =>  { return searchResult.market; }
+        );
+      }
+      let keyedCategoryResults: any = {};
+      if (params.categories) {
+        const extendedSearchOptions: ExtendedSearchOptions[] = [];
+        for (let i = 0; i < params.categories.length; i++) {
```
bthaile, line 453: prob need to filter out empty strings that might get passed in the category array.

packages/augur-sdk/src/state/getter/Markets.ts
```diff
@@ -35,12 +37,15 @@ const getMarketsParamsSpecific = t.intersection([
     creator: t.string,
     category: t.string,
     search: t.string,
-    reportingState: t.union([t.string, t.array(t.string)]),
+    reportingStates: t.array(t.string),
     disputeWindow: t.string,
```
bthaile, line 64: looks like this is a dispute window. not sure how we'd use this as a passed in parameter.  It would be useful to know the dispute window if the market is in reporting.

packages/augur-sdk/src/state/db/MarketDB.ts
adrake33: Added to https://github.com/AugurProject/augur/issues/3023

packages/augur-sdk/src/state/getter/Markets.ts
```diff
@@ -414,106 +442,153 @@ export class Markets {
...
-              selector: { market: marketCreatedLogInfo[0], amount: '0x00' },
-            });
-            if (
-              orderCreatedLogs.length - orderCanceledLogs.length ===
-              orderFilledLogs.length
-            ) {
-              includeMarket = false;
-            }
-          }
+      if (params.disputeWindow) {
```
bthaile, line 492: It would be helpful to have disputeWindow information for the MarketInfo if the market has a dispute window.
having startTime and endTime of the dispute window would be awesome. (this can be added in a different PR, created tix <https://github.com/AugurProject/augur/issues/3025>)

packages/augur-sdk/src/state/getter/Markets.ts
```diff
@@ -414,106 +442,153 @@ export class Markets {
...
-            });
-            const reportingState = await getMarketReportingState(
-              db,
-              marketCreatedLogInfo[1],
-              marketFinalizedLogs
-            );
-            if (!reportingStates.includes(reportingState)) {
-              includeMarket = false;
-            }
+      if (params.includeMarketsWithoutOrders === false) {
```
bthaile: This got removed by design. UI won't be using this filter. The liquidity filter replace `no orders` filter.

bthaile (approved):
I have some comments, overall looks good. Currently the UI isn't hooked up to test this. UI will be adding filters in the near term then we can hook this up and test.

packages/augur-sdk/src/state/db/MarketDB.ts
adrake33: If category3 is specified but a market does not have it, should it be included in the results?

bthaile: I might not be understanding the code above. Does this above just create the index for search? not actually add data?

packages/augur-sdk/src/state/getter/Markets.ts
adrake33: Yeah, I was also going to do that with the sortBy param. This is meant to allow multiple reporting states, but I'll add validation for that.

packages/augur-sdk/src/state/getter/Markets.ts
adrake33: Ok, I'm not sure if flex search supports the wildcard character or not. I'll have to check that.

packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -268,10 +285,9 @@ export class MarketDB extends DerivedDB {
               longDescription = info.longDescription ? info.longDescription : "";
               resolutionSource = info.resolutionSource ? info.resolutionSource : "";
               _scalarDenomination = info._scalarDenomination ? info._scalarDenomination : "";
-              tags = info.tags ? info.tags.toString() : ""; // convert to comma separated so it is searchable
             }
 
-            this.flexSearch.add({
+            this.flexSearchIndex.add({
```
bthaile, line 290: below is prob where my comment should have been. what if some of the field values are empty like "" or undefined does that string get indexed?

packages/augur-sdk/src/state/db/MarketDB.ts
bthaile: gotcha on the above. 👍

adrake33: Correct, this just creates the index for the field.

packages/augur-sdk/src/state/getter/Markets.ts
bthaile: I'd like to know what is available in order to discuss with design. Maybe they want to offer an advanced search option. not pressing just curious.

packages/augur-sdk/src/state/db/MarketDB.ts
adrake33: It defaults to being set to empty string and then gets set to the category if it's specified. I don't think indexing the empty string here should be an issue.

packages/augur-sdk/src/state/getter/Markets.ts
adrake33: Ok, will remove this then.

packages/augur-sdk/src/state/getter/Markets.ts
adrake33: Yeah, I'll take another look at handling empty strings before merging.

packages/augur-sdk/src/state/getter/Markets.ts
adrake33: This was left in because we were trying to keep the getters as similar to the getters from v1. If disputeWindow isn't very useful as a param, I'll remove it.

packages/augur-sdk/src/state/getter/Markets.ts
adrake33: Sounds good!

packages/augur-sdk/src/state/getter/Markets.ts
bthaile: Yes remove it.

adrake33:
@bthaile Thanks for the feedback! Will make a few more changes to this tomorrow and then merge.

packages/augur-sdk/src/state/getter/Markets.ts
adrake33: Yeah, sounds like we'll need to discuss this more at some point.

adrake33:
I removed the `reportingStates` & `includeMarketsWithoutOrders` params from getMarkets & updated the tests. I made a note to make market creator full-text-searchable & take another look at category3 indexing as part of #3023.

PR #3026, opened 2019-08-01
mobile-outcome-select

### LINKED ISSUE #1768: Mobile Trading Page Outcome selection
Add 16 px of padding to the button of this
Add same opacity (90%) and background blur as the subnav

Should look like this when content scrolls under

JohnDanz:
fixed opacity and scroll under for outcome select

fixes https://github.com/AugurProject/augur/issues/1768


bthaile (approved):
lgtm

PR #3029, opened 2019-08-01
liquidity / market depth fix

### LINKED ISSUE #3027: Create Market - Initial Liquidity
Fix market depth chart to not be flat.

andrewdamelio:
fixes #3027 

- misc fixes
- The depth chart will show tooltips now
- The depth chart will not be flat
- Depth chart / Order book now shows correct cumulative shares

### DIFF
--- a/packages/augur-ui/src/modules/markets/reducers/new-market.ts
+++ b/packages/augur-ui/src/modules/markets/reducers/new-market.ts
@@ -5,11 +5,11 @@ import {
   CLEAR_NEW_MARKET
 } from "modules/markets/actions/update-new-market";
 import { RESET_STATE } from "modules/app/actions/reset-state";
-import { 
-  SETTLEMENT_FEE_DEFAULT, 
-  EXPIRY_SOURCE_GENERIC, 
-  DESIGNATED_REPORTER_SELF, 
-  AFFILIATE_FEE_DEFAULT, 
+import {
+  SETTLEMENT_FEE_DEFAULT,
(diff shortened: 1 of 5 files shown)


packages/augur-ui/src/modules/markets/reducers/new-market.ts
```diff
@@ -78,22 +78,25 @@ export default function(newMarket: NewMarket = DEFAULT_STATE, { type, data }: Ba
...
-        const orderInfo = Object.assign({}, order);
+
+      const updatedOrders = existingOrders.map((order: LiquidityOrder) => {
+          const orderInfo = Object.assign({}, order);
         if (createBigNumber(order.price).eq(createBigNumber(price)) && order.type === type) {
-          orderInfo.quantity = createBigNumber(order.quantity).plus(createBigNumber(quantity)).toNumber();
+          orderInfo.quantity = createBigNumber(order.quantity).plus(createBigNumber(quantity)).toString();
+          orderInfo.shares = createBigNumber(order.quantity).plus(createBigNumber(quantity)).toString();
           orderInfo.orderEstimate = createBigNumber(order.orderEstimate).plus(
             createBigNumber(orderEstimate.replace(" DAI", ""))
```
bthaile, line 93: not a fan of this logic, shouldn't be removing a string in order to treat something like a number.  need to just add `totalCost` to order that is used in new Market reducer and not parse out order estimate.

bthaile (approved):
lgtm, some legacy crap code, but can be addressed at future time.

Issue #2986, opened 2019-07-30
Trading order form flickers


bthaile:
Trading order form changes from logged in to logged out and back to logged in.

Look in either sign in with MM code or in init-augur. There is code that auto logs out the user.

bthaile:
I see the order form flicker when navigating to the first market trading page after logging in. Doesn't happen every time.  The UI doesn't "logout".

Another reason the form might flicker is if  availableEth updates and has 0 for a value for split seconds.

I'll keep an eye out.

bthaile:
noticed the flicker is still happening. easier to see on local chain. reopening

justinbarry:
LGTM

Chwy5:
Have Not seen it flicker in the last few days....closing

PR #3037, opened 2019-08-01
fix wiring up events when user logs in.

### LINKED ISSUE #2990: Available funds doesn't decrease after trade
there's a `new block` event failing issue https://github.com/AugurProject/augur/issues/2702 

The UI depends on new block event to refresh user's assets (DAI balance) even when the UI gets a new block event the available funds doesn't decrease after a trade, even after the user refreshes the UI the available balance doesn't decrease.

bthaile:
doesn't totally fix this issue https://github.com/augurproject/augur/issues/2986
fixes https://github.com/augurproject/augur/issues/2990

but does fix the subscription to events when user logs in.


andrewdamelio (approved):
Lgtm

Issue #1748, opened 2019-04-18
Mobile Dropdown Menus on Portfolio and Account Summary: Make menu items bigger


matt-bullock:
Make menu items the same size as the dropdown button. Increase the type size by 1px of the menu items 

They look too small on the current build
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/ee13d50c-ca38-4a62-936c-aaed03ca9e87)

PR #3045, opened 2019-08-02
Bigger-Dropdowns-Mobile

JohnDanz:
made the menu size/dropdown size more consistent on mobile, cleaned up some classes to remove _ style class names and repeated class name text.

https://github.com/augurproject/augur/issues/1748


bthaile (approved):
lgtm

PR #3052, opened 2019-08-02
dropdown new rules

### LINKED ISSUE #2595: Input components: additional rules
Dropdown component
- The dropdown list min width is the same width as the input itself
   - If any of the strings are longer than the input width, then increase the width of the dropdown list
   - If the string selected is wider than the text input, truncate and allow hover to reveal tooltip

Text input component 
- If the user enters a string that is longer than the text input, replicate this material design approach: 
   - https://storage.googleapis.com/spec-host-backup/mio-design%2Fassets%2F1Wp0W2DNhQfZq3bhLq7EDSvMXhi0y8kwr%2Ftextfields-input-motion2.mp4

phoebemirman:
closes https://github.com/AugurProject/augur/issues/2595

### DIFF
--- a/packages/augur-ui/src/modules/common/form.styles.less
+++ b/packages/augur-ui/src/modules/common/form.styles.less
@@ -722,8 +722,8 @@
       color: @color-primary-alternate-two;
 
       > svg {
-        height: @size-12;
-        width: @size-12;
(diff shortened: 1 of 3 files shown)


bthaile (approved):
lgtm

PR #3053, opened 2019-08-02
Make market creator full-text-searchable & add more info to getMarketDisputeInfo

### LINKED ISSUE #3025: Disputing: MarketInfo dispute window
Add dispute window information on MarketInfo. 

if the market is in reporting then it'll have a dispute window assigned. Need to add

add dispute pacing to the market info object
`DisputePacingOn`

`DisputeWindow` : {
address: <contract address>
startTime: <number> timestamp
endTime: <number> timestamp
}

Also make market creator full-text searchable

adrake33:
Closes #3025.


bthaile (approved):
Lgtm, dispute window is there and pacing property was added. We talked about also adding designated reporter as filter property like market creater. It can be added in the future.

adrake33:
@bthaile `designatedReporter` is already a param for getMarkets, so filtering by it is already possible.

PR #3090, opened 2019-08-06
1802 get account time ranged stats

brianosaurus:
adds the getAccountTimeRangedStats and tests\

Specifically ads these return vales
```ts
export interface AccountTimeRangedStatsResult {
  // Yea. The ProfitLossChanged event then
  // Sum of unique entries (defined by market + outcome) with non-zero netPosition
  positions: number;

  // OrderEvent table for fill events (eventType == 3) where they are the orderCreator or orderFiller address
  // if multiple fills in the same tx count as one trade then also couldting just the unique tradeGroupId from those
  numberOfTrades: number;

  marketsCreated: number;

  // Trades? uniq the market 
  marketsTraded: number;

  // DisputeCrowdsourcerRedeemed where the payoutNumerators match the MarketFinalized winningPayoutNumerators
  successfulDisputes: number;

  // For getAccountTimeRangedStats.redeemedPositions use the InitialReporterRedeemed and DisputeCrowdsourcerRedeemed log?
  redeemedPositions: number;
}
```


packages/augur-sdk/src/event-handlers.ts
```diff
@@ -33,7 +33,7 @@ export interface CompleteSetsSold extends FormattedEventLog {
   numCompleteSets: string;
   marketOI: string;
   fees: string;
-  timestamp: number;
+  timestamp: Address;
```
adrake33: I think this be defined as a string instead of an Address. Currently, Address is the same as a string, but my original thinking was that an Address should be formatted like an Ethereum address. (I don't think there's currently a way to do this in TS, but I could be wrong.) I defined the type Address to make the type clearer in the docs.

packages/augur-sdk/src/state/Server.ts
```diff
@@ -1,8 +1,9 @@
 import * as HTTPEndpoint from "./HTTPEndpoint";
 import * as Sync from "./Sync";
 import * as WebsocketEndpoint from "./WebsocketEndpoint";
-import {EndpointSettings} from "./getter/types";
-import {EventEmitter} from "events";
+import { EndpointSettings } from "./getter/types";
+import { EventEmitter } from "events";
+import { DB } from "./db/DB";
```
adrake33: DB can be removed here since it's not being used.

packages/augur-sdk/src/state/getter/get-account-time-ranged-stats.ts
```diff
@@ -0,0 +1,160 @@
...
+import * as _ from 'lodash';
+import { DisputeCrowdsourcerRedeemed, MarketFinalized, ProfitLossChanged, OrderEvent } from "../../event-handlers";
+
+export interface AccountTimeRangedStatsResult {
+  // Yea. The ProfitLossChanged event then
+  // Sum of unique entries (defined by market + outcome) with non-zero netPosition
+  positions: number;
+
+  // OrderEvent table for fill events (eventType == 3) where they are the orderCreator or orderFiller address
+  // if multiple fills in the same tx count as one trade then also couldting just the unique tradeGroupId from those
```
adrake33: "couldting" should be "counting" here?

packages/augur-sdk/src/state/test-connector.ts
```diff
@@ -10,6 +10,8 @@ import { MarketCreated, NewBlock } from "../events";
 import { Markets } from "./getter/Markets";
 import { SubscriptionEventName } from "../constants";
 import { SEOConnector } from "../connector/seo-connector";
+// import { WebWorkerConnector } from "../connector/ww-connector";
+import { DB } from "./db/DB";
```
adrake33: These imports can be removed as well.

adrake33 (approved):
Lgtm overall. Just a few minor things.

PR #3093, opened 2019-08-06
Set newblock event timestamp to Augur time

### LINKED ISSUE #3036: Set NewBlock event timestamp using Fake time if enabled

justinbarry:
This will handle both fake and real time.

closes #3036


bthaile:
```
export interface NewBlock extends FormattedEventLog {
  blocksBehindCurrent: number;
  highestAvailableBlockNumber: number;
  lastSyncedBlockNumber: number;
  percentSynced: string;
  timestamp: number;
}
```

timestamp is a number. does this type apply?

justinbarry:
@bthaile Good catch.

bthaile (approved):
lgtm

PR #3111, opened 2019-08-08
DeployerConfiguration is now an interface

robert-e-davidson3:
We've had two bugs so far due to technical debt from DeployerConfiguration so I rewrote it to be more reliable. It could use more work in refactoring how environment variables are used since defaults should be specified at the client not in the library.


nuevoalex (approved):
👍

Issue #3109, opened 2019-08-07
date time / timezone picker incorrect date


bthaile:
User chosen values
![image](https://user-images.githubusercontent.com/3970376/62648377-489b4600-b918-11e9-81a7-5abc2250375d.png)

Displayed on Review page
![image](https://user-images.githubusercontent.com/3970376/62648349-39b49380-b918-11e9-8a2e-31a96a87c7ee.png)

I think the user's sequence of picking values influences the timestamp creation.

PR #3112, opened 2019-08-08
New market creation end time display in review

bthaile:
design agreed to adding both timezone date time and UTC date time to review page.

Added new date time display, showing both timezone time and utc time, fixed a few types, moved complexity of building endTime formatted to format-date utility

https://github.com/AugurProject/augur/issues/3109

fixes user selecting date time logic, a little clean up.

image


packages/augur-ui/src/modules/create-market/components/form-details.tsx
```diff
@@ -127,42 +125,41 @@ export default class FormDetails extends React.Component<
...
               minute={minute}
               meridiem={meridiem}
-              onChange={(label: string, value: string) => {
+              onChange={(label: string, value: number) => {
                 onChange(label, value)
               }}
               onFocusChange= {(focused: Boolean) => {
                 if (!hour) {
-                  onChange("hour", "12");
+                  onChange("hour", 12);
```
phoebemirman: time selector won't let you type "0"

phoebemirman: clicking on the time selector doesn't fill in the values as supposed to

phoebemirman: i'm also not seeing the validations for time show up

packages/augur-ui/src/modules/create-market/components/form-details.tsx
```diff
@@ -172,12 +169,12 @@ export default class FormDetails extends React.Component<
               focused={s.timeFocused}
               errorMessage={validations[currentStep].hour}
             />
-            <TimezoneDropdown onChange={(value: number, offset: string) => {
-              onChange("offsetName", value);
+            <TimezoneDropdown onChange={(offsetName: string, offset: number, timezone: string) => {
               onChange("offset", offset)
-            }} timestamp={endTime} timezone={offsetName} />
+              onChange("timezone", timezone)
```
phoebemirman, line 174: i don't see timezone being used at all?

bthaile:
I’ll take a look and design want 4px between the times

packages/augur-ui/src/modules/create-market/components/common.styles.less
```diff
@@ -212,7 +212,10 @@
 		.text-16-medium;
 
 		color: @color-alternate-info-text;
-	}
+  }
+  > span:last-of-type {
```
phoebemirman: this isn't right. it will effect all the other small subheaders

phoebemirman (changes requested):
see comments, also not sure if it is showing the times correct still

packages/augur-ui/src/modules/create-market/components/form-details.tsx
phoebemirman: also "0" isn't valid for minutes in the time selector. which makes it a bit messed up

packages/augur-ui/src/modules/create-market/components/form-details.tsx
bthaile: it is on my branch and everything is checked in. I'll see

bthaile: everything is checked in
image

packages/augur-ui/src/modules/create-market/components/form-details.tsx
bthaile: hour is defaulted to 12, I'll change that back to null

packages/augur-ui/src/modules/create-market/components/form-details.tsx
phoebemirman: but then timezone is never used anywhere? don't see why it needs to get set

packages/augur-ui/src/modules/create-market/components/common.styles.less
bthaile: what is the right way?

phoebemirman: do it in the individual time component or > span:nth-of-type(3) could work.

packages/augur-ui/src/modules/create-market/components/form-details.tsx
phoebemirman: you don't need both offsetName and timezone

phoebemirman (changes requested):
not seeing time validations:
Screen Shot 2019-08-08 at 10 09 12 AM

needs to set time when clicked on selector, and shouldn't be showing a single zero here at this time:
Screen Shot 2019-08-08 at 10 09 02 AM

packages/augur-ui/src/modules/create-market/components/form-details.tsx
bthaile: timezone is needed only to pick the correct item in the timezone dropdown. True, only the parts of the selected timezone is used.

phoebemirman (approved):
need to fix the timezone for real in other PR

PR #3113, opened 2019-08-08
Gnosis initial spike

nuevoalex:
Initial Spike for using the Gnosis Safe Wallet for Augur.

- Gnosis Relay API: Package for communicating with the Gnosis Relay API service: https://github.com/gnosis/safe-relay-service . The tests in this package currently test against rinkeby since there were problems I encountered with running locally.
- contract-dependencies-gnosis: contract dependencies that are capable of handling normal transactions (via their parent implementation) as well as sending transactions meant for a Gnosis Safe. When using a safe one can use the relay service or sign and submit transactions directly.
- Augur SDK Gnosis support: Support for getting a deterministic safe address and manually creating the safe. In practice we will likely end up only using the function to get a safe address however currently the relay does not let us pass all the params used during safe creation so we can't create safes with a pre-determined address without going through their relay for that part and storing the result it sends back.


packages/augur-sdk/src/Augur.ts
```diff
@@ -124,11 +129,38 @@ export class Augur<TProvider extends Provider = Provider> {
   }
 
   async getAccount(): Promise<string | null> {
-    const account = await this.dependencies.address;
+    let account;
+    if (this.dependencies.useSafe) {
+      account = this.dependencies.safeAddress;
```
bthaile, line 134: Does this need an ‘await’ to get the safeAddress?

packages/augur-sdk/src/Augur.ts
```diff
@@ -124,11 +129,38 @@ export class Augur<TProvider extends Provider = Provider> {
...
+    } else {
+      account = await this.dependencies.address;
+    }
     if (!account) return account;
     return getAddress(account);
   }
 
+  async sendETH(address: string, value: BigNumber): Promise<void> {
```
bthaile, line 142: Are other individual methods needed to also send DAI and REP?

packages/gnosis-relay-api/src/tests/GnosisRelayAPI.test.ts
```diff
@@ -0,0 +1,132 @@
...
+        nonce: await wallet.getTransactionCount(),
+        to: safeAddress,
+        value: new ethers.utils.BigNumber(10).pow(16),
+    };
+    const txResponse = await wallet.sendTransaction(depositTransaction);
+    console.log(`Waiting on TX: ${txResponse.hash}`);
+    await provider.getTransactionReceipt(txResponse.hash);
+
+    // Lets send a transaction through the safe using the relay service
+    const gnosisSafe = new ethers.Contract(safeAddress, abi["GnosisSafe"], provider);
```
bthaile, line 80: This looks like its using Rinkeby? Will the account thats being used to fund the test eventually run out of rinkeby eth?

bthaile (approved):
Lots to digest in this PR. Doesnt look like it would directly break anything currently working. Lgtm

The UI will have to make some changes to give the user a gnosis safe login. We’ll need to have a meeting to figure out the flow.

packages/gnosis-relay-api/src/tests/GnosisRelayAPI.test.ts
nuevoalex: Yea I mentioned that above. There were some issues I encountered with running a local relay that I punted on in favor of just testing on Rinkeby for now. This isnt in CI so it'll only spend funds when run manually at the moment. That said its a public private key so anyone could drain the account.

packages/augur-sdk/src/Augur.ts
nuevoalex: nope its just a property

pgebheim:
> Lots to digest in this PR. Doesnt look like it would directly break anything currently working. Lgtm
> 
> The UI will have to make some changes to give the user a gnosis safe login. We’ll need to have a meeting to figure out the flow.

The current thoughts on this are that if a user shows up with anything in the wallet they connect (eth, tokens etc), then they never get the Safe, and they just sign txns as normal. If a user signs up with a FRESH account however, then the middleware will do the mumbo-jumbo to generate the SAFE address. This address would be the DEPOSIT address on the `Deposit` page, or in the case of using Edge or Portis or any other fancy wallet, we would pass this pre-calculated Safe address as the depository address for their fiat onboarding.

So for the UI the work is really the same as any other account except that the middleware will need to be told to use the Safe path.

The only thing truely needed that we've identified is for the UI to display failed sends through the Relay service with a notification they need to be delivered and the user will need Eth to do so.

PR #3128, opened 2019-08-08
Pending market

phoebemirman:
part of https://github.com/AugurProject/augur/pull/3122


bthaile:
look at `loadPendingOrdersTransactions` in pending-orders-management.ts

it checks confirmations before adding pending order from localstorage when user logs in.

need to do the same for pendingQueue.

packages/augur-ui/src/modules/pending-queue/actions/pending-queue-management.ts
```diff
@@ -13,12 +13,14 @@ export const addPendingData = (
   pendingId: string,
   queueName: string,
   status: string,
+  info?: object,
```
bthaile: might as well type this

packages/augur-ui/src/modules/pending-queue/reducers/pending-queue.ts
```diff
@@ -10,16 +10,18 @@ const DEFAULT_STATE: PendingQueue = {};
 export default function(pendingQueue: PendingQueue = DEFAULT_STATE, { type, data }: BaseAction): PendingQueue {
   switch (type) {
     case ADD_PENDING_DATA: {
-      const { pendingId, queueName, status, blockNumber } = data;
+      const { pendingId, queueName, status, blockNumber, info } = data;
```
bthaile, line 12: add `info` to `PendingQueue` in types.ts

packages/augur-ui/src/modules/markets/selectors/user-markets.ts
```diff
@@ -1,21 +1,39 @@
...
+  selectCurrentTimestamp,
+  (allMarkets, pendingQueue, marketTradingHistory, authorId, currentTimestamp) => {
     if (!allMarkets || !authorId) return null;
-    const filteredMarkets = allMarkets.filter(
+    let filteredMarkets = allMarkets.filter(
       market => isSameAddress(market.author, authorId)
     );
+    const pendingMarkets = Object.keys(pendingQueue[CREATE_MARKET] || {}).map(key => {
+      let data = pendingQueue[CREATE_MARKET][key];
+      const extraInfo = JSON.parse(data.data._extraInfo);
```
bthaile: I'm not a fan of the `_` in the variable names showing up in a selector. I think those would have been removed when the pendingQueue object was created and saved to state.


Also might as well type the Transaction parameters object so we get type checking on `retryCreateMarket(txParams: CreateMarketTxParameters)` method when that is created.

packages/augur-ui/src/modules/events/actions/add-update-transaction.ts
```diff
@@ -30,6 +45,16 @@ export const addUpdateTransaction = (txStatus: Events.TXStatus) => (
         }
         break;
       }
+      case CREATEMARKET:
+      case CREATECATEGORICALMARKET:
+      case CREATESCALARMARKET:
+      case CREATEYESNOMARKET: {
+        if (hash) dispatch(addPendingData(hash, CREATE_MARKET, eventName, transaction.params));
```
bthaile: build `info` typed object here (call translation helper/method). add another property `txParams` for the transaction params (which will be used for retry)

bthaile (approved):
lgtm

PR #3134, opened 2019-08-08
Fix date ranges in selectors

### LINKED ISSUE #3126: Fix date ranges on getAccountTimeRangeStats
Currently the date ranges tare broken for this getter.

brianosaurus:
Date ranges were broken beforehand. The notation causes an allDocs().

Fixes that.


adrake33 (approved):
Lgtm

bthaile:
Does it need it's own file, seems to not follow current convention. Can this getter be put in Users.ts?

adrake33:
@bthaile Yeah, that's true. This would probably be good to move this to Users.ts. Can you please create a task for that?

Issue #3054, opened 2019-08-05
Remove Gas estimate on trading order form


bthaile:
remove ETH gas cost from order form.

pgebheim:
Done in #3135

Issue #2663, opened 2019-07-05
Tell user number of tx needed to take order


bthaile:
It's possible that the user will need to sign multiple tx to make a trade.

In order form confirmation display number of transactions needed to make trade. 

This task is waiting for design.

bthaile:
moving this to trading, we now have the information needed from sim trade to tell the user roughly how many tx will be needed.

Chwy5:
The box notifying number of transactions needed does appear in the order ticket.  But there is an issue with Xing out the box to see the information behind it.  When I click on the x to close the box, it returns again

joeykrug:
done afaik

PR #3135, opened 2019-08-09
multiple transactions to trade

bthaile:
removed label telling user amount of gas needed, direction from produce https://github.com/AugurProject/augur/issues/3054

dev task, PR is mostly for adding the infrastructure 
https://github.com/AugurProject/augur/issues/2663

bug fix in sdk to trade needing multiple transactions, added warning to user there will be multiple tx

image

placeholder to tell user multiple transactions are needed. Might change if design comes up with standard way of telling user multiple transactions are needed to complete a task.

To test, 
created a lot of 1 DAI orders on the book

Using a different user
take the multiple orders, notice the warning in the order confirmation.


nuevoalex (approved):
👍

phoebemirman (approved):

Issue #510, opened 2018-12-04
Need to add TOTAL or Total ETH


petong:
I thought this from the first day i tried this and still as a trader am not comfortable not seeing Total Eth per open order and Filled order.

As a trader I want to know how much total eth is tied up in each open order (quantity x price), so if i need to free up eth I can cancel very quickly...Same for filled orders...I want to know the total eth consumed in a trade...in case I want to replicate or partial the trade again.  A user does not want to have to plug numbers into the order ticket to see and Est Cost (if they already have an open position or filled trade)

Also, (already brought up in another ticket) but the est cost does not work properly in my opinion as if a user has an open position,  it shows the user the est cost taking into account the open position....a user wants to see what the true est cost would be without the open position.
...

> Also, (already brought up in another ticket) but the est cost does not work properly in my opinion as if a user has an open position, it shows the user the est cost taking into account the open position....a user wants to see what the true est cost would be without the open position.

@chwy Are you talking about how we show the cost of an order in both ETH and shares? For example if you are holding 10 shares, and you go to place in a sell order of 20 shares, that will show you an estimated cost of X ETH and 10 shares?

> As a trader I want to know how much total eth is tied up in each open order (quantity x price), so if i need to free up eth I can cancel very quickly...Same for filled orders...I want to know the total eth consumed in a trade...in case I want to replicate or partial the trade again. A user does not want to have to plug numbers into the order ticket to see and Est Cost (if they already have an open position or filled trade)

Are you saying we should have a total cost column in the Open Orders section here on the Trading page as well as on the Portfolio page:

![clipboard 2018-22-06 at 4 56 06 pm](https://user-images.githubusercontent.com/11491365/49621339-af2b2700-f979-11e8-9707-22ef5da60e75.png)

What if part of the order cost is in shares?  For example, if you were holding 10 shares and then you placed a sell order for ten shares?
 -sharkcrayon
...

It might just be a terminology issue.....To me est cost (is trade based) is (what i explained above), what Augur calculates for est cost (if you have an open position) is the estimated change in position for that trade.

what most betters and traders will be familiar with (imo) should be similar to centralized exchanges.....On a CEX when you input your amount and price it shows the amount to be traded.  For example, If i was long 10 eth on Gdax and entered to sell 2 lots at $500...it will show my total (Est cost) as $1000
 -Chwy5
...

Sorry back to the original question...Are you saying we should have a total cost column in the Open Orders section here on the Trading page as well as on the Portfolio page:

Yes i do
[...]

PR #3190, opened 2019-08-10
added contributeToTentative to flash scripts

### LINKED ISSUE #3018: Flash script: "contributeToTentative"
create flash script to add REP to current tentative winning outcome.

bthaile:
fixes https://github.com/AugurProject/augur/issues/3018

only added  contributeToTentative to flash script. Will add fork in a different PR.

fixed some copy paste parameter descriptions.
added some pass throughs in contract-api.


packages/augur-tools/src/flash/scripts.ts
```diff
@@ -444,14 +468,82 @@ export function addScripts(flash: FlashSession) {
...
+        abbr: 'o',
+        description:
+          'outcome of the market, non scalar markets 0,1,3,... for scalar markets use price',
+        required: true,
+      },
+      {
+        name: 'amount',
+        abbr: 'a',
+        description: 'amount of REP to dispute with, use display value',
+        required: false,
```
robert-e-davidson3: I think this has to be `true` due to line #510, where you do `new BigNumber(amount)`. A missing `amount` is going to be an empty string (I think?) not `'0'`.

bthaile: ok, prob copy paste error, will check it out

robert-e-davidson3 (changes requested):
Flash script's `amount` parameter needs to be `required: true`.

packages/augur-tools/src/flash/scripts.ts
bthaile: good catch, pushed change

robert-e-davidson3 (approved):
LGTM!

Issue #3194, opened 2019-08-12
Scalar Market on the trading pg displaying incorrect Min/Max


Chwy5:
In the header, the word MAX isnt even being shown and the inputs for the max range are incorrect

range used was 0-50

![8-12d.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/48637fd8-e6c1-4ef9-9671-fa9461714c4a)

Chwy5:
fixed - closing

Issue #3193, opened 2019-08-12
Scalar Market on trading pg displaying Scientific Notation


Chwy5:
Under the Outcome section the Max of the range is being displayed as scientific notation

the range I used to create this market is 0-50

![8-12c.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/249e5e6f-1ee4-4474-9a4f-29902b3b92a7)

bthaile:
issues found while creating scalar market.

"maxPrice":"0.00000000000000005",
"cumulativeScale":"0.00000000000000005",
"settlementFee":"0.1000000000000001",
"reportingFeeRate":"0.0000000000000001",
tickSize":"0.00000000000000000001"

Chwy5:
fixed - closing

PR #3200, opened 2019-08-12
fix prices for scalar markets

bthaile:
I think these are caused by the same issue
https://github.com/augurproject/augur/issues/3194
https://github.com/augurproject/augur/issues/3193

Needed to format/convert price to on chain value before creating market. 

Fixed min max price on trading page for scalar markets

image


phoebemirman (changes requested):
can't create a market
Screen Shot 2019-08-12 at 12 48 48 PM

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🆒
 To view it, be sure to [join Augur Project](https://dashboard.featurepeek.com/login) on FeaturePeek.

 This link will always point to this PR's latest changes: https://peek.run/jvgh2qz

packages/augur-ui/src/modules/create-market/constants.ts
```diff
@@ -387,7 +387,9 @@ export const VALIDATION_ATTRIBUTES = {
     checkFilledNumber: true,
     checkFilledNumberMessage: 'Enter precision',
     checkPositive: true,
-    checkLessThan: true,
+    checkLessThan: false,
+    checkDividedBy: true,
+    dividedBy: true,
```
phoebemirman: you don't need dividedBy here

bthaile: oops, mean to remove that

phoebemirman (dismissed):

packages/augur-ui/src/modules/create-market/components/form.tsx
```diff
@@ -408,7 +408,7 @@ export default class Form extends React.Component<FormProps, FormState> {
         ? isLessThan(value, readableName, newMarket.maxPrice, lessThanMessage)
         : '',
       checkDividedBy ? dividedBy(value, readableName, newMarket.minPrice, newMarket.maxPrice) : '',
-      checkGreaterDate ? dateGreater(value, currentTimestamp) : '',
+      checkGreaterDate && newMarket.endTimeFormatted ? dateGreater(newMarket.endTimeFormatted.timestamp, currentTimestamp) : '',
```
phoebemirman: this should be generic

bthaile: since the property is aggregated from all date components that wouldn't be possible. will add conditional as you suggested. if (label === "setEndTime") ...

phoebemirman (approved):

phoebemirman (approved):

PR #3245, opened 2019-08-15
issue with getMarkets throwing

bthaile:
issue in getMarkets getter causing failure


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 👩‍🎤
 To view it, be sure to [join Augur Project](https://dashboard.featurepeek.com/login) on FeaturePeek.

 This link will always point to this PR's latest changes: https://peek.run/kq3hj97

adrake33 (approved):
Lgtm. The reason it's checking the 0th index is because the query should only return one result since it's searching for a specific market ID. That's odd that the market it's searching for isn't getting returned. I added a TODO to look into this.

robert-e-davidson3 (approved):
lgtm

PR #3269, opened 2019-08-16
case issue with String and added await so we can see speed of getters

bthaile:
string got lowered cased should be upper case causes UI to explode


petong (approved):
lgtm

PR #3295, opened 2019-08-20
Create meta object after offset/limit

### LINKED ISSUE #3282: getMarkets filters markets by max fee/pagination, but they still show in returned meta object
only return meta for the markets that are included in `markets` collection as part of the returned object from getMarkets getter.

markets can be filtered out by any of the filters or just not included because of pagination.

adrake33:
Closes #3282. The actual bug was occurring because there was only one market result, and the offset was set to 1. The empty `markets` array was correct, but the `meta` object was not. When I implemented offset/limit a while back, I forgot to move the line that creates the `meta` object after that block of code.

### DIFF
--- a/packages/augur-sdk/src/state/getter/Markets.ts
+++ b/packages/augur-sdk/src/state/getter/Markets.ts
@@ -559,14 +559,14 @@ export class Markets {
       }
     }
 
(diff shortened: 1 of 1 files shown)


bthaile (approved):
lgtm

PR #3303, opened 2019-08-20
0x spike

nuevoalex:
Added the ZeroXTradeToken. This is the contract that will fulfill signed 0x orders for Augur.

Also removed the code to change on chain order price and to support ignoring shares since they both add complexity that we don't really want.


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 💪

 This link will always point to this PR's latest changes: https://peek.run/5m3hg39

packages/augur-core/source/contracts/trading/ZeroXTradeToken.sol
```diff
@@ -0,0 +1,282 @@
+pragma solidity 0.5.10;
+pragma experimental ABIEncoderV2;
+
+import "ROOT/libraries/math/SafeMathUint256.sol";
+import "ROOT/libraries/ContractExists.sol";
```
adrake33, line 5: This import can be removed.

packages/augur-core/source/contracts/ZeroXExchange.sol
```diff
@@ -0,0 +1,1425 @@
...
+    // A valid order remains fillable until it is expired, fully filled, or cancelled.
+    // An order's state is unaffected by external factors, like account balances.
+    enum OrderStatus {
+        INVALID,                     // Default value
+        INVALID_MAKER_ASSET_AMOUNT,  // Order does not have a valid maker asset amount
+        INVALID_TAKER_ASSET_AMOUNT,  // Order does not have a valid taker asset amount
+        FILLABLE,                    // Order is fillable
+        EXPIRED,                     // Order has already expired
+        FULLY_FILLED,                // Order is fully filled
+        CANCELLED                    // Order has been cancelled
```
adrake33, line 82: Nit: We use "canceled" (with one "l") in the rest of the code. Recommend doing the same here for consistency.

adrake33 (approved):
Looks good! Are ZeroXPoC.sol & test_0x.py still needed?

packages/augur-core/source/contracts/trading/ZeroXTradeToken.sol
nuevoalex: Will remove it in a follwup update

packages/augur-core/source/contracts/ZeroXExchange.sol
nuevoalex: This contract is just for testing context and is an amalgamation of 0x contracts so not really expected to follow our same patterns

nuevoalex:
Gonna keep the other 0x code here for now as its sort of a separate use case

Issue #3285, opened 2019-08-20
Fix unsign orders modal and add bulk order creation


bthaile:
add bulk create orders, need to batch X orders in one transaction.

PR #3306, opened 2019-08-21
Fix unsigned orders modal

### LINKED ISSUE #3286: MC: total cost for liquidity doesn't update
create scalar market with orders and notice the amount of DAI needed. Go back and remove orders and create different orders. 

notice that the total cost doesn't update.

### LINKED ISSUE #3308: Trading page order book doesn't show scalar liquidity orders
orders created with `createOrder.publicCreateOrder` are not showing up in the order book. 

create market with initial liquidity, when market is created click on `view orders` on Portfolio page in my markets. create the orders. verify the orders were created and navigate to market's trading page notice the orders aren't in the orderbook. 

If user creates a order through trading order form the order does show up in the orderbook.

bthaile:
fixes https://github.com/AugurProject/augur/issues/3286
fixes https://github.com/AugurProject/augur/issues/3308
associated with https://github.com/AugurProject/augur/issues/3285

change generate transaction id algo to use sha256 so the transaction parameters can be larger. 
updated lots of types.
moved calculate liquidity to reducer, there was a race condition doing the calculations outside of the reducer.


packages/augur-ui/src/modules/contracts/actions/contractCalls.ts
```diff
@@ -267,15 +274,15 @@ export function createMarketRetry(market: CreateMarketData) {
     designatedReporterAddress: market.txParams._designatedReporterAddress,
     minPrice: market.txParams._prices && market.txParams._prices[0],
     maxPrice: market.txParams._prices && market.txParams._prices[1],
-    tickSize: market.txParams._numTicks,
+    numTicks: market.txParams._numTicks,
```
phoebemirman, line 278: don't you need to change the other createMarket call in submit-new-market to send numTicks instead of tickSize?

phoebemirman (commented):
when i add liquidity, then go back and switch market types, the total DAI cost at the end doesn't update.

phoebemirman (commented):
this isn't working:
1. create a market with liquidity
2. it fails
3. retry
4. liquidity shows up

not sure if it was broken on purpose

phoebemirman (commented):
liquidity isn't clearing from local storage after i submit an order

phoebemirman (commented):
- [x] liquidity shows up for all market types after first market creation try
- [x] liquidity shows up for all market types after second market creation try
- [x] total liquidity amount is correct after going back and changing market type
- [ ] liquidity modal has right amounts and information
- [ ] after submitting liquidity successfully, the liquidity goes away
- [x] after market fails, then you refresh, market shows, then after a successful retry the correct liquidity shows

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 💚

 This link will always point to this PR's latest changes: https://peek.run/36whqnv

packages/augur-ui/src/modules/contracts/actions/contractCalls.ts
bthaile: it works

bthaile:
found a bug with creating scalar markets `tickSize` isn't on newMarket object. 

fixed bug with scalar markets retry. Fixed the tickSize/numTicks issue.

 *  liquidity modal has right amounts and information
 * after submitting liquidity successfully, the liquidity goes away
These will be done on a different PR

phoebemirman (approved):

PR #3307, opened 2019-08-21
Replay logs

robert-e-davidson3:
Can replay logs either into a ganache seed or onto another chain. This involves first grabbing logs from a chain (including a V1 chain like mainnet!) then sending transactions to another chain. Target chains must be V2.

I stopped fauceting so much Cash and REP during deployment because it was masking testing issues where we were depending on Cash and REP magically being available. I fixed the tests that broke as a result of this change but calculating the cost of a Fill Order has turned out to be really difficult.

Replaying onto a geth chain is not yet tested but is likely to work since replaying onto another ganache chain works.


ghost (commented):

 This link will always point to this PR's latest changes: https://peek.run/654h50o

packages/augur-tools/src/flash/replay-logs-v1.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    let result;
+    switch(marketType) {
+      case 0: // YES_NO
+        result = await user.createYesNoMarket({
+          endTime: adjustedEndTime,
+          feePerCashInAttoCash,
+          affiliateFeeDivisor: feeDivisor,
+          designatedReporter: reporter.publicKey,
+          extraInfo,
```
bthaile: category has moved to extraInfo. 

If I'm understanding this, V1 log has _topic on market creation and to be compliant with V2 _topic will need to be added to `categories` in extraInfo.

categories: [_topic, "from V1", ""]

_topic or category, don't rem what comes in on V1 log

packages/augur-tools/src/libs/contract-api.ts
```diff
@@ -59,17 +74,10 @@ export class ContractAPI {
       feePerCashInAttoCash: new BigNumber(10).pow(16),
       affiliateFeeDivisor: new BigNumber(25),
       designatedReporter: this.account.publicKey,
-      extraInfo: JSON.stringify({categories: [" "], description: "description"}),
+      extraInfo: JSON.stringify({categories: [' '], description: 'description'}),
```
bthaile: might as well add categories for when Flash creates markets, we will want them to show up in the UI. 

I suggest `categories: ['flash', 'YesNo Market]`

packages/augur-tools/src/libs/contract-api.ts
```diff
@@ -79,18 +87,11 @@ export class ContractAPI {
...
```
bthaile: same comment as above.

packages/augur-tools/src/libs/contract-api.ts
```diff
@@ -102,7 +103,7 @@ export class ContractAPI {
       feePerCashInAttoCash: new BigNumber(10).pow(16),
       affiliateFeeDivisor: new BigNumber(25),
       designatedReporter: this.account.publicKey,
-      extraInfo: JSON.stringify({categories: [" "], description: "description", _scalarDenomination: "scalar denom 1"}),
+      extraInfo: JSON.stringify({categories: [' '], description: 'description', _scalarDenomination: 'scalar denom 1'}),
```
bthaile: same comment as above

bthaile:
a few comments about V1 category moving to extra info in V2. 

I'd like to see more documentation on how to use the replay, I think it's a really cool feature and more documentation would help us use it more. An example of how to replay kovan V2 to local seed would be great.

robert-e-davidson3:
> I'd like to see more documentation on how to use the replay, I think it's a really cool feature and more documentation would help us use it more. An example of how to replay kovan V2 to local seed would be great.

Good idea! Added docs: https://github.com/AugurProject/docs/pull/29

packages/augur-tools/src/libs/contract-api.ts
robert-e-davidson3: sure!

packages/augur-tools/src/flash/replay-logs-v1.ts
robert-e-davidson3: looks like it's just `topic`. added!

bthaile (approved):
lgtm

PR #3313, opened 2019-08-21
added transacation hash to marketsinfo to match liquidity

bthaile:
adds transaction hash to markets info


phoebemirman (approved):

bthaile:
seems we don't need this. we can just calculate a hash id for liquidity.

PR #3323, opened 2019-08-22
Add tx hash market

bthaile:
Some clean up and
Went with transaction hash on marketInfo object to pull liquidity orders. 
There were some blocks to re-create the tx params using marketInfo object. now market creation is a fire and forget operation.

Also I think we can use transactionHash to remove pending items. An example is orders. each order has a transactionHash. something to think about.


bthaile:
tested: 
 * creating each market with liquidity, market gets created and liquidity shows up. 
 * creating market with liq. kill browser after signing. when logged back in liquidity shows up
 * created categorical market with liq, retry to create market liquidity shows then refresh, liquidity shows back up

We can chat about it. it works current with my most recent changes.

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🌠

 This link will always point to this PR's latest changes: https://peek.run/vv2h936

phoebemirman (approved):
looks good, just doesnt work though if they aren't around when the pending log comes in though

PR #3341, opened 2019-08-23
Add indexes to Generic EventLog DBs

### LINKED ISSUE #3110: Need a way to specify indexes needed when PouchDBs are created
**What**
We need pouch to be fast so we'll need indexes.

From the PouchDB website:
```
At a basic level, there are two steps to running a query: createIndex() (to define which fields to index) and find() (to query the index).
```

**What**
So far there seems to be a few ideas:
* hardcode a definition of what indexes to make for what tables to Augur.js
* do the same thing to DB
* maybe do the same thing to SyncableDB

**When**
This should probably be done during performance testing of Pouch.

justinbarry:
Closes #3110


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). ⛳️

 This link will always point to this PR's latest changes: https://peek.run/xjlhgw3

bthaile (approved):
lgtm, for initial adding of indexes

Issue #2863, opened 2019-07-22
market listing categories section


bthaile:
Create category picker component. it will take meta and category array filter used to gen the meta from getMarkets and provide the selected category when the user selects it. It will need to provide an array of categories ["primary", "secondary", "tertiary"].

https://www.figma.com/file/GKhft5ZOwuH3ANrupGNCLHUU/Market-list-page?node-id=187%3A14954

There are 5 top level categories that are always shown.
Sports, Politics, Entertainment, Finance, Crypto. They are sorted by number of markets then by alphabetical order

below popular categories is `All categories` that shows all primary categories that are not popular categories.  
There is a "show more" functionality to show the rest of the "below the fold" primary categories. 

Notice the category section is a max-height scrollable section.

User selects primary category
https://www.figma.com/file/GKhft5ZOwuH3ANrupGNCLHUU/Market-list-page?node-id=187%3A31453

User selects secondary category
https://www.figma.com/file/GKhft5ZOwuH3ANrupGNCLHUU/Market-list-page?node-id=187%3A33683

User selects tertiary  category
https://www.figma.com/file/GKhft5ZOwuH3ANrupGNCLHUU/Market-list-page?node-id=187%3A35913

notice that only the active level is indented the parents are left aligned

PR #3357, opened 2019-08-26
Category filter markets list v2

### LINKED ISSUE #3094: Remove `tags` from UI
tags are no longer used. They can be removed from the UI. Primary, secondary and tertiary categories are replacing tags

categories are listed at top of market card and are clickable to navigate the user to that category on market list page.

andrewdamelio:
closes #3094 #2863 

- Wired up Category filter to the loadMarkets call
- Removed old Category/Tag code
- Removed a bunch of unused code from App.tsx


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 💎

 This link will always point to this PR's latest changes: https://peek.run/m96hvxn

bthaile (commented):
code lgtm, going to test it out.

phoebemirman (commented):
Screen Shot 2019-08-27 at 4 23 26 PM
Screen Shot 2019-08-27 at 4 23 14 PM
why are these numbers different after clicking on the category?

it's also wrong for the crypto one, says one but has two

bthaile:
could be error in data from getter, def needs to be run to ground

andrewdamelio:
@bthaile @phoebemirman 
> why are these numbers different after clicking on the category?

Assuming we should only be getting back category counts for CURRENT filtered results this seems to make sense. When you click through to a specific category you are making a more specific search so you should get back more results for that selected category if they exist.

I can add the loading so we don’t see that flash from one number to another.

> https://recordit.co/bCxKz2PrTM

bthaile:
The category counts should include all the markets in the category and subcategories. 

For example, if Sports has 2 and Tennis has 2 then the meta data should have Sports as having 4 markets. If the data isn't coming back like this from the getter we have a bug.

bthaile (approved):
lgtm

Issue #3210, opened 2019-08-13
Current Value in the Positions section of the Portfolio pg not working


Chwy5:
displaying only 0.000s

Both for long and short positions

Chwy5:
@bthaile still showing me 0.000s

Chwy5:
after the last fix no longer showing me zeros, but showing me incorrect value.

I sold 11 at .27 and last trade price was .28, my current value s/b 7.92= (1-.28)*11
it currently shows me with -7.02 value which is incorrect

phoebemirman:
seems like getUserTradingPositions is sending back the wrong currentValue

bthaile:
This might have been addressed with newest change to currentValue

Chwy5:
this is working correctly now...closing

Issue #3106, opened 2019-08-07
Frozen Funds Values not Always Updating Correctly


bthaile:
Frozen Funds isn't correct.

Scenario:

1. create order and check that frozen funds updates correct when order is created
2. create order that take a trade and verify that frozen funds updates correctly

verified that situation 1. (create order) works fine.
2., when user closes a position frozen funds doesn't update from Users.ts `getProfitLossSummary` getter.

pgebheim:
Will re-visit during next QA phase.

pgebheim:
@Chwy5 -- Joey said you just had an example here,

mind putting it here?

pgebheim:
need to know which cases they aren't updating correctly, and are all of these listed on that P/L spreadsheet?

bthaile:
updated the description of ticket with more information

Chwy5:
Testing with Tom....when user has a position and closes it out and reverses a position the FF doesnt update...

Example Buying 1 share at 0.25

Sell 2 shares at .15

My FF should increase by 0.7...i get this b/c Selling out my 1 long at .15 ( I would get back .15) and opening a new position of selling 1 at .15 (FF would be .85)= (.85+.15)= .7

joeykrug:
fixed

PR #3359, opened 2019-08-26
calculate correct current value based on last price and if the user h…

### LINKED ISSUE #3360: last trade price for calculating user position is wrong
user traded to have a long position avg price of 0.31. 
Then there was a trade at 0.35 but the user's position doesn't reflect the last trade price. 
```json
{"1":{"timestamp":1566860676,"frozenFunds":"3.1031","marketId":"0x25D690355b8E2C5e6966FB902C16373fEBbD0f48","outcome":1,"netPosition":"10.01","rawPosition":"10.01","averagePrice":"0.31","realized":"0","unrealized":"0","total":"0","unrealizedCost":"3.1031","realizedCost":"0","totalCost":"3.1031","realizedPercent":"0","unrealizedPercent":"0","totalPercent":"0","currentValue":"3.1031"}}
```

looks like query was only getting last market fills by user and not actual last market fill.

bthaile:
…as long or short position
fixes https://github.com/AugurProject/augur/issues/3360
https://github.com/AugurProject/augur/issues/3210
https://github.com/augurproject/augur/issues/3106

should fix long and short positions current value calculations

create a long position in a market
create a short position in a market

the current value should be the same as total cost initially. 

have a second user traded in the markets to move market price (up or down)

look at user's positions the current value should reflect the last price of the market. So that total cost and current value are no longer the same.


phoebemirman (approved):

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🏅

 This link will always point to this PR's latest changes: https://peek.run/exoheg2

PR #3361, opened 2019-08-26
Implement getDisputeWindow getter & tests

robert-e-davidson3:


bthaile:
This getter isnt for the user. The amount of participation tokens purchased is total not by user. User specific data will go on getUserReporting history getter.

packages/augur-sdk/src/state/getter/Universe.ts
```diff
@@ -0,0 +1,87 @@
...
+  static async getDisputeWindow(
+    augur: Augur,
+    db: DB,
+    params: t.TypeOf<typeof Universe.getDisputeWindowParams>
+  ): Promise<DisputeWindow|null> {
+    const currentDisputeWindow = await getCurrentDisputeWindow(db, params.universe);
+    if (currentDisputeWindow === null) return null;
+    const { disputeWindow, startTime, endTime } = currentDisputeWindow;
+
+    const account = await augur.getAccount();
```
bthaile: this isn't account specific getter

packages/augur-sdk/src/state/getter/Universe.ts
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  @Getter('getDisputeWindowParams')
+  static async getDisputeWindow(
+    augur: Augur,
+    db: DB,
+    params: t.TypeOf<typeof Universe.getDisputeWindowParams>
+  ): Promise<DisputeWindow|null> {
+    const currentDisputeWindow = await getCurrentDisputeWindow(db, params.universe);
+    if (currentDisputeWindow === null) return null;
```
bthaile: dispute window should never be null. it will need to be created.

packages/augur-test/src/tests/state/getter/Universe.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    john = await ContractAPI.userWrapper(ACCOUNTS[0], provider, seed.addresses);
+    mary = await ContractAPI.userWrapper(ACCOUNTS[1], provider, seed.addresses);
+    db = await mock.makeDB(john.augur, ACCOUNTS);
+    api = new API(john.augur, Promise.resolve(db));
+    await john.approveCentralAuthority();
+    await mary.approveCentralAuthority();
+    console.log('CLOCKWORK', 1);
+  }, 120000);
+
+  test('getDisputeWindow', async () => {
```
bthaile, line 31: I'm thinking, need to test for sdk to create a dispute window by calling method on universe and verifying a dispute window is created.

robert-e-davidson3: That solution bothers me. Aside from the naming issue (getters shouldn't set), how will that look to the user? They navigate to the dispute page and get a metamask popup? That might be the right flow but I think I'd be more comfortable with a version of the dispute window page that has a clearly-labelled button for creating the dispute window.

ghost (commented):

 This link will always point to this PR's latest changes: https://peek.run/1rehxn2

packages/augur-sdk/src/state/getter/Universe.ts
```diff
@@ -0,0 +1,111 @@
...
+      startTime: previousDisputeWindow.startTime + disputeRoundDurationSeconds,
+      endTime: previousDisputeWindow.endTime + disputeRoundDurationSeconds,
+      purchased: new BigNumber(0),
+      fees: new BigNumber(0),
+    };
+  } else { // Use a default for the clients
+    return {
+      address: '',
+      startTime: '0',
+      endTime: `0x${currentTime.toString(16)}`,
```
robert-e-davidson3: @bthaile I'm worried about this part. I couldn't think of a good default to use so I went with something obviously wrong. My thinking is that users will find it less confusing if they see an obviously wrong date range suddenly change into the correct date change (when the dispute window is created) than if a valid-looking dispute window suddenly changes its range.

bthaile: just make end time 0

packages/augur-sdk/src/state/getter/Universe.ts
```diff
@@ -0,0 +1,111 @@
...
+import {
+  Augur,
+} from '../../index';
+
+import * as t from 'io-ts';
+
+export interface DisputeWindow {
+  address: Address;
+  startTime: Timestamp;
+  endTime: Timestamp;
```
bthaile: lets be consistent with timestamps. 
Users.ts uses number `  endTime: t.number,` 
we don't need BigNumber or string for timestamps, imo

bthaile (changes requested):
lgtm, few changes

packages/augur-sdk/src/state/getter/Universe.ts
robert-e-davidson3: done!

packages/augur-sdk/src/state/getter/Universe.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    const { disputeWindow, startTime, endTime } = currentDisputeWindow;
+    const purchased = await getParticipationTokens(augur, disputeWindow);
+    const fees = await getFees(augur, disputeWindow);
+    return {
+      address: disputeWindow,
+      startTime,
+      endTime,
+      purchased: purchased.toNumber(),
```
bthaile: This needs to be a string. could be very large

packages/augur-sdk/src/state/getter/Universe.ts
```diff
@@ -0,0 +1,107 @@
...
+      fees: fees.toNumber(),
```
bthaile: same here, could be very large

bthaile (approved):
lgtm, nice!!! 👍

PR #3366, opened 2019-08-27
Implement lastTradedTimestamp sort in getMarkets

### LINKED ISSUE #3322: Implement lastTradedTimestamp sort in getMarkets

adrake33:
Closes #3322


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🤩

 This link will always point to this PR's latest changes: https://peek.run/5m3hgrj

packages/augur-sdk/src/state/getter/Markets.ts
```diff
@@ -560,8 +560,36 @@ export class Markets {
       }
     }
 
+    if (params.sortBy === GetMarketsSortBy.lastTradedTimestamp) {
+      // Create market ID => index mapping
+      const keyedMarkets = {};
+      for (let i = 0; i < marketsResults.length; i++) {
+        keyedMarkets[marketsResults[i].market] = i;
+      }
+      const orderFilledLogs = await db.findOrderFilledLogs({
```
bthaile, line 577: does this get to latest fill order log for the market or all fill order logs for the market? Does pouch queries have a sort?

adrake33: This gets the fill orders for all logs. Pouch does have a sort, but it requires indexes to exists. I think Justin recently made some changes related to indexes, so I'll see if PouchDB's sort can now be used.

adrake33: Actually, sorting using PouchDB doesn't really help here, since you'd still have to loop through all the results. There's a `limit` parameter in Pouch, but it limits the total number of documents. I don't think there's a way to limit the number of results per market in a single query.

bthaile: if you could sort and only grab the top fill order, then there would be less logs to loop through.

adrake33: How would you grab the top fill order? Pouch can't do it, and if you loop through returned results, you'd still have to check if the next market was different from the current market.

bthaile: ok, 👍

adrake33: I suppose you could do it by keeping track of how many markets you've encountered so far, and once you get the latest for that many markets, you can stop iterating near the end of the results. It probably wouldn't result in a very noticeable performance improvement, but it would be one way to optimize.

bthaile (approved):
lgtm

Issue #3369, opened 2019-08-27
frozen funds is off when shorting


bthaile:
these are prob related. 

short YesNo market. User shorted 5 at 0.25 and current value is -3.75. 
frozen funds was current before trade, after frozen funds goes to 56.6514. 

(new ff value) 341.569 - (previous ff value) 294.3953 + (short trade) 3.6 + (recent short trade) 0.81 = 51.5837 off

pgebheim:
Closed with #3376

PR #3376, opened 2019-08-27
re-organize query for profit loss, too many records were being returned

bthaile:
https://github.com/augurproject/augur/issues/3369

queries needed to be changed, too many records were being returned after short trade.


adrake33 (approved):
Lgtm

PR #3383, opened 2019-08-28
fix bug where user populates order form but then logs outs

bthaile:
user logs in and clicks order to populate order form then logs out. 

issue was allowanceBigNumber is undefined when they logout and causes explosion.


phoebemirman (approved):

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 💚

 This link will always point to this PR's latest changes: https://peek.run/5m3hgve

PR #3390, opened 2019-08-28
use get trading history to get user's filled orders

bthaile:
calling getOrders with orderState of FILLED isn't working, created a tix for backend work. meanwhile to get UI testing going calling different getter.


robert-e-davidson3 (approved):
lgtm

Issue #3002, opened 2019-07-30
Market cards hook up icons to data


phoebemirman:
![Screen Shot 2019-07-30 at 3.34.23 PM.png](https://images.zenhubusercontent.com/574e216bbf93871ccd9be5fb/ce973a9f-e734-42e5-9985-1eceedbf7a8d)

Market creator and designated reporter information is on marketData object from market selector.

DAI icon data comes from user's positions. 

reporting/disputing info will be from a new reporting getter. user reporting information won't be on MarketInfo from getMarketsInfo getter. 
use flash scripts to report and dispute to get an idea of disputeInfo object shape.

pgebheim:
This depends on Dispute and Stakes data being available for the market, which will come with Reporting.

bthaile:
waiting for get account reporting history getter to be finished

bthaile:
now we have data this can be closed

PR #3411, opened 2019-08-29
wired up login account to augur logo in market card

bthaile:
https://github.com/AugurProject/augur/issues/3002
This ticket isn't completely done, get account reporting history getter is only stubbed out. There is a little UI work when to fully hook up getter results. 

This PR depends on previous login account reporting reducer changes.

image


packages/augur-ui/src/modules/account/selectors/has-stake-in-market.ts
```diff
@@ -0,0 +1,32 @@
+import { createSelector } from 'reselect';
+import { Getters } from '@augurproject/sdk/build';
```
andrewdamelio: are we supposed to be referencing the build folder?

andrewdamelio (approved):
Looks good - 1 comment

packages/augur-ui/src/modules/account/selectors/has-stake-in-market.ts
bthaile: good catch, nope. my mistake

bthaile: fixed

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🎯

 This link will always point to this PR's latest changes: https://peek.run/p37hpz5

PR #3424, opened 2019-08-30
0x mesh relay

nuevoalex:
0x Mesh Relay and middleware initial spike.

On the DB level there is a ZeroXOrders table that will get cleared on every startup and rebuilt. It subscribes to new events from the mesh to update the existing orders during a session.

On the API layer there is a zeroX namespace for placing orders and making trades. The trade making API will get matching orders from the DB and do 1 or more transactions to fill / create until the order is satisfied.

Things still TODO:

 - Derived Markets DB that uses the ZeroXOrders table for liquidity metrics
 - Specific getter functionality so when a market orderbook is requested if we havent synced it yet we just pull from the mesh (not possible currently since there is not filtering API for mesh relay)
 - Simulate Order support for 0x trading
 - All the various TODOs from the code aded


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🌶

 This link will always point to this PR's latest changes: https://peek.run/7g6hj43

bthaile (approved):
lgtm, can CI passed, 👍

packages/augur-artifacts/src/addresses.json
```diff
@@ -73,24 +73,24 @@
...
+  "CreateOrder": "0xaF9D4A4915d850F49cF384727A98D9aEbFD414e4",
+  "FillOrder": "0xfDC85a01D1db389c4dC15771Ab516e4dC5A0CdE1",
+  "OICash": "0x56150981f7E17459b4D9F2c26d141E1147D184AC",
+  "Orders": "0xf7eFed92220cDF2d8d23eAB47B0cb2c11DE5e80C",
+  "ProfitLoss": "0x2c36908607b0ECcDe7b1Ad69512e5068Ae399db0",
+  "ShareToken": "0x27D40dA32537754f4D56A6A65150b767299ACECE",
+  "SimulateTrade": "0x8689FFa6153347B5dA82EA15ce5CD11Cd222599f",
+  "Trade": "0x0705b0dF22030a601F1DC8F35D9e55f4eBB1307B",
+  "ZeroXTradeToken": "0x0C031b9Ec3DBADB9a02862AdFA12e3C76dE82379"
  },
```
bthaile: we there a kovan deployment with this PR?

packages/augur-tools/src/libs/contract-api.ts
```diff
@@ -183,6 +186,38 @@ export class ContractAPI {
...
+    await this.faucet(cost);
+    await this.augur.zeroX.placeTrade(params);
+  }
+
+  async placeBasicYesNoZeroXTrade(direction: 0 | 1, market: ContractInterfaces.Market, outcome: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7, displayAmount: BigNumber, displayPrice: BigNumber, displayShares: BigNumber, expirationTime: BigNumber): Promise<void> {
+    await this.placeZeroXTrade({
+      direction,
+      market: market.address,
+      numTicks: await market.getNumTicks_(),
+      numOutcomes: await market.getNumberOfOutcomes_() as unknown as 3 | 4 | 5 | 6 | 7 | 8,
```
bthaile: should this be 3 | 4 ... 7 and not include 8?

packages/augur-ui/package.json
```diff
@@ -50,7 +50,7 @@
     "@ledgerhq/hw-transport-u2f": "4.48.0",
     "@portis/web3": "2.0.0-beta.30",
     "async-es": "2.6.3",
-    "bignumber.js": "6.0.0",
+    "bignumber.js": "8.0.2",
```
bthaile, line 53: this seems like a sig. upgrade. Any big changes we need to know about?

nuevoalex: Nothing that should affect us

packages/augur-sdk/src/state/db/ZeroXOrders.ts
```diff
@@ -0,0 +1,125 @@
...
+    let documents;
+    if (orders.length > 0) {
+      documents = _.map(orders, this.processOrder.bind(this));
+
+      success = await this.bulkUpsertUnorderedDocuments(
+        documents
+      );
+    }
+    if (success) {
+      augurEmitter.emit("ZeroXOrders", {});
```
justinbarry, line 102: Why not include the docs in the payload here like we are doing above?

justinbarry (approved):
One feedback item. Otherwise LGTM

packages/augur-sdk/src/state/db/ZeroXOrders.ts
nuevoalex: The context here is a full sync. This will always be every single order on Augur.

Above it is a single update.

Issue #3436, opened 2019-09-02
Finish remaining issues with syncing tracked users


adrake33:
TODO: 
- Replace Sync.createAPIAndController function with a better solution
- Make other endpoints (HTTP, WS, etc) respond to UserDataSynced event
- Enforce mixed case addresses

pgebheim:
@adrake33 Removing this from market list as its not necessary, but putting it in the general Sprint 16 bucket.

pgebheim:
We've removed `tracked users`

PR #3437, opened 2019-09-02
Sync tracked users separately from other data

adrake33:
Closes 3373. There are a few remaining issues, which I've created a ticket for here: https://github.com/AugurProject/augur/issues/3436.


bthaile (approved):
looks to be working.

PR #3463, opened 2019-09-03
update defaults, fix invalid filter

### LINKED ISSUE #3425: set default filters
Default filters to be set are:
fee: 0 - 5%
spread: less than 10%

https://www.figma.com/file/GKhft5ZOwuH3ANrupGNCLHUU/Market-list-page?node-id=187%3A28361

andrewdamelio:
closes #3425 

- sets fee filter to 0-5% as default
- sets spread filter to less than 10 as default
- fixes bug with invalid filter


ghost (commented):

 This link will always point to this PR's latest changes: https://peek.run/kq3hmop

packages/augur-ui/src/modules/app/components/inner-nav/markets-list-filters.tsx
```diff
@@ -19,7 +19,7 @@ import updateQuery from 'modules/routes/helpers/update-query';
 interface MarketsListFiltersProps {
   maxFee: string;
   maxLiquiditySpread: string;
-  showInvalid: string;
+  includeInvalidMarkets: 'show' | 'hide';
```
bthaile: use constants here

packages/augur-ui/src/modules/app/components/inner-nav/markets-list-filters.tsx
```diff
@@ -55,7 +55,7 @@ export default class MarketsListFilters extends React.Component<
         selectedFee: filterOptions.maxFee || this.props.maxFee,
         selectedSpread: filterOptions.spread || this.props.maxLiquiditySpread,
         showInvalidDefault:
-          filterOptions.showInvalid || this.props.showInvalid,
+          filterOptions.showInvalid || this.props.includeInvalidMarkets,
```
bthaile: why is there two properties here to set `shwInvalidDefault` ?

packages/augur-ui/src/modules/types.ts
```diff
@@ -425,7 +425,7 @@ export interface FilterSortOptions {
   marketSort: string;
   maxFee: string;
   maxLiquiditySpread: string;
-  includeInvalidMarkets: string;
+  includeInvalidMarkets: 'show' | 'hide';
```
bthaile: same comment as above, can we use constants here?

packages/augur-ui/src/modules/app/components/inner-nav/markets-list-filters.tsx
andrewdamelio: thanks for the catch Tom!

andrewdamelio: @bthaile correction.

filterOptions is coming from params. So if we have the param on the URL we use it, otherwise we fallback to includeInvalidMarkets (which is our reducer state).

I updated the var name to be more clear.

bthaile (approved):
lgtm

Issue #3489, opened 2019-09-06
Implement liquidity sort in getMarkets getter, make it the default, update tests


adrake33:

adrake33:
Closing, as this was completed in an old PR

PR #3491, opened 2019-09-06
Hourly liquidity snapshots & sorting by liquidity in getMarkets

### LINKED ISSUE #2997: Store hourly snapshots of MarketDB in new RankingMarketDB
The `getMarkets` getter needs to be able to sort markets by last liquidity-depleted. 

Discord conversation starts at https://discordapp.com/channels/378030344374583298/557226630293225472/605822297659408385

As per Alex on Discord: "Since this is naively a recalc required on every daily load for all markets regardless of changes having occurred I think some care should go into only recalcing what is needed"

adrake33:
Closes #2997 #3489 #3180


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🙌

 This link will always point to this PR's latest changes: https://peek.run/qo0hjqr

packages/augur-sdk/src/state/getter/Markets.ts
```diff
@@ -1365,3 +1375,129 @@ async function getMarketsSearchResults(
...
+  return marketsResults;
+}
+
+/**
+ * Sets the `liquidity` property for the last 24 hours of liquidity for all markets in `marketsResults`.
+ *
+ * @param {DB} db Database object to use for getting `liquidity` info
+ * @param {Array<Object>} marketsResults Array of market objects to add `liquidity` to
+ */
+async function setLiquidity(db: DB, marketsResults: any[]): Promise<Array<{}>>  {
```
bthaile, line 1425: why `any` here for `marketsResults` and not typed?
Also with return value why empty object `{}` and not typed?

packages/augur-sdk/src/state/getter/Markets.ts
```diff
@@ -1365,3 +1375,129 @@ async function getMarketsSearchResults(
...
   return Augur.syncableFlexSearch.where(whereObj);
 }
+
+/**
+ * Sets the `lastTradedTimestamp` property for all markets in `marketsResults`.
+ *
+ * @param {DB} db Database object to use for getting `lastTradedTimestamp` info
+ * @param {any[]} marketsResults Array of market objects to add `lastTradedTimestamp` to
+ */
+async function setLastTradedTimestamp(db: DB, marketsResults: any[]): Promise<Array<{}>>  {
```
bthaile, line 1387: same comment as below:
why any here for marketsResults and not typed?
Also with return value why empty object {} and not typed?

packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -59,26 +59,26 @@ export class MarketDB extends DerivedDB {
       },
     };
 
-    const result = await this.stateDB.findInDerivedDB(this.stateDB.getDatabaseName("CurrentOrders"), request);
+    const result = await this.stateDB.findInDerivedDB(this.stateDB.getDatabaseName('CurrentOrders'), request);
```
bthaile: nit pic, shouldn't database names have constants for look ups?

packages/augur-sdk/src/api/Liquidity.ts
```diff
@@ -152,11 +152,11 @@ export class Liquidity {
 
       // BIDS
       for (let outcome = 1; outcome < numOutcomes; outcome++) {
```
bthaile, line 154: need to add comment for future. outcome starts at 1 because invalid is outcome 0 and not included in liquidity calc. something like that.

bthaile (changes requested):
def didn't follow the logic all the way through, but generally looks good. suggestions for standards. made comments.

side not: super nit pic, I'd like to see field names as constants so there is tab completion. it' increases coding as well. 
larger task since there is a munch of field strings.

packages/augur-sdk/src/state/getter/Markets.ts
adrake33: I agree we should ideally be using an actual type here, but for some reason, FlexSearch is returning a different type than the expected Array<SearchResults<MarketFields>>, which in turn means I have to define the type as any[] as a workaround. This will take a fair amount of time to figure out why this is happening, so I've created a TODO item to come back to this.

packages/augur-sdk/src/state/db/MarketDB.ts
adrake33: True, this can be changed to `DB.findCurrentOrderLogs`. Will update.

bthaile (approved):
Ok, minor nit pics that were covered by todos

PR #3527, opened 2019-09-08
Only require specific 0x modules

nuevoalex:
It looks like the 0x/subproviders module still requires node-hid as an optional dependency. Not sure if that will work, though in theory it should I would hope. If not resolving this will require refactoring so that we don't use their provider API and just roll our own stuff or convincing them to break that out.


ghost (commented):

 This link will always point to this PR's latest changes: https://peek.run/qo0hjpr

packages/augur-ui/src/modules/alerts/reducers/alerts.ts
```diff
@@ -37,7 +37,7 @@ const DEFAULT_STATE: Array<Alert> = [];
  * @returns {Alert[]}
  *
  */
-export default function(alerts = DEFAULT_STATE, { data, type }: BaseAction): Array<Alert> {
+export default function alert(alerts = DEFAULT_STATE, { data, type }: BaseAction): Array<Alert> {
```
bthaile, line 40: is adding `alert` here a mistake?

nuevoalex: Nope, as it was written this will make a webpack build fail. Justin and I were unsure of how master was even passing with this.

bthaile: there are many `export default function` methods in the Ui. do they all need to change? 53 instances

nuevoalex: Doesn't seem so. @justinbarry discovered and fixed this so he can likely speak to it better

justinbarry: I really have no idea why this is the case. Possible the babel loader is pulling in a config from one of the required packages but that is just a guess.

justinbarry: Just confirmed master is building fine. This is def one of those things we can tackle later.

justinbarry (approved):
LGTM

bthaile (approved):
Lgtm

petong (approved):
I added export NODE_OPTIONS="--max-old-space-size=6144" before calling the tests since we were getting out of memory errors

PR #3530, opened 2019-09-09
use 2 decimals for DAI values

### LINKED ISSUE #3494: Use 2 decimal places for DAI display values
use 2 decimal places for all DAI values displayed. 
prices
volume
open interest

note: REP gets displayed using 4 decimal places

andrewdamelio:
closes #3494

### DIFF
--- a/packages/augur-sdk/src/api/Liquidity.ts
+++ b/packages/augur-sdk/src/api/Liquidity.ts
@@ -43,7 +43,7 @@ export interface GetLiquidityParams {
   reportingFeeDivisor: BigNumber;
   marketFeeDivisor: BigNumber;
   numOutcomes: number;
-  spread: number;
+  spread?: number;
 }
 
 export class Liquidity {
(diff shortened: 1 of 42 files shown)


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🥇

 This link will always point to this PR's latest changes: https://peek.run/0elhvoo

bthaile (commented):
can you track down why formatDai returns 0.0000 even though format of 2 decimals is passed 

as you mentioned, just a reminder. value of "0" acts differently with format-number, need to remove this

```
  if (value.eq(ZERO)) {
    if (zeroStyled) return formatNone();
    if (blankZero) return formatBlank();
  }
```

andrewdamelio:
@bthaile tracked it down.
see https://github.com/AugurProject/augur/pull/3530/commits/03f4049b85c0a6794cd35c29eb09c257be41e946#diff-e409b22a2e78488cbe6a7fa97986de97R361

Looks like for zero we would always pass in USUAL_NUMBER_DECIMAL_PLACES (4) before, this shouldn't be the case now. I adjusted so we use the provided decimal amount from the options if provided.

bthaile (approved):
formatDai should not need options passed in. right

```
export function formatDai(num: NumStrBigNumber, opts: FormattedNumberOptions = {}): FormattedNumber {
  return formatNumber(num, {
    decimals: 2,
    decimalsRounded: 2,
    denomination: " DAI",
    positiveSign: false,
    zeroStyled: false,
    blankZero: false,
    bigUnitPostfix: false,
    ...opts,
  });
}
```

unless they are override values.

bthaile (approved):
lgtm, will work

PR #3532, opened 2019-09-09
Robert/1812 implement get fork migration totals getter

robert-e-davidson3:


packages/augur-tools/src/flash/fork.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    // Contribute aka dispute. Opposing sides to keep raising the stakes.
+    const numerators = i % 2 === 0 ? conflictNumerators : payoutNumerators;
+    await user.contribute(marketContract, numerators, SOME_REP);
+  }
+
+  return false; // failed to fork
+}
+
+function getPayoutNumerators(market: MarketInfo, outcome: BigNumber): BigNumber[] {
```
bthaile: why not call util to get the payout numerators?

robert-e-davidson3: I got an error that looked like a cyclic import. I can make the sdk use these utils to avoid duplication.

packages/augur-sdk/src/state/getter/Universe.ts
```diff
@@ -13,8 +18,25 @@ export interface DisputeWindow {
...
+  outcomes: Outcome[];
+}
+
+interface Outcome {
+  outcomeName: string;
+  outcome: string; // non-scalar markets this is outcome id.
+  amount: string; // atto-rep given to this outcome
+  isMalformed: boolean;
+  payoutNumerators: string[]; // not needed by UI, but 3rd parties might want it
+}
```
bthaile, line 34: I missed this, we need a flag `isInvalid` for scalars when outcome is 0. 
0 outcome id can mean either a price between min and max or invalid
outcomeName for scalar markets is the denomination. see Markets.ts when it builds up outcomes

packages/augur-sdk/src/state/getter/Universe.ts
```diff
@@ -36,6 +58,65 @@ export class Universe {
...
+      selector: {
+        parentUniverse: params.universe,
+      },
+    });
+
+    const outcomes: Outcome[] = await Promise.all(children.map(async (child): Promise<Outcome> => {
+      const payoutNumerators = child.payoutNumerators.map((hex) => Number(hex).toString(10));
+
+      // TODO min and max prices probably need to be translated into display prices
+      const outcome = calculatePayoutNumeratorsValue(maxPrice, minPrice, numTicks, marketTypeName, payoutNumerators);
```
bthaile, line 104: there will be a new parameter to `calculatePayoutNumeratorsValue` called isInvalid it'll default to false

packages/augur-sdk/src/state/getter/Universe.ts
```diff
@@ -36,6 +58,65 @@ export class Universe {
...
+      const isMalformed = outcome === MALFORMED_OUTCOME;
+
+      const repTokenAddress = await (await augur.contracts.universeFromAddress(params.universe)).getReputationToken_();
+      const repToken = augur.contracts.reputationTokenFromAddress(repTokenAddress, augur.networkId);
+
+      return {
+        outcomeName: isMalformed ? 'malformed' : getOutcomeDescriptionFromOutcome(Number(outcome), forkingMarket),
```
bthaile: `malformed` needs to be a constant that UI can also use.

packages/augur-sdk/src/state/getter/Universe.ts
```diff
@@ -36,6 +58,65 @@ export class Universe {
...
+        outcome,
+        amount: String(await repToken.totalSupply_()),
```
bthaile: is this an off chain call? looks like an off-chain call,  if so that is frowned upon.

packages/augur-sdk/src/utils.ts
```diff
@@ -165,3 +166,52 @@ export function calculatePayoutNumeratorsArray(
...
+      return new BigNumber(market.prices[0]).toString(10);
+    } else {
+      return new BigNumber(market.prices[1]).toString(10);
+    }
+  } else {
+    return toAscii(market.outcomes[new BigNumber(outcome).toNumber()]);
+  }
+}
+
+export function getOutcomeFromPayoutNumerators(payoutNumerators: BigNumber[]): number {
```
bthaile: I see this has moved, it's not needed there is already a method that converts payout numerators to outcome. Use the existing one.

plus, this one doesn't check for malformed payout numerators.

packages/augur-sdk/src/utils.ts
```diff
@@ -165,3 +166,52 @@ export function calculatePayoutNumeratorsArray(
   }
   return payoutNumerators;
 }
+
+export function getOutcomeDescriptionFromOutcome(
```
bthaile, line 219: Markets.`getMarketOutcomes` needs to use this method. 

make constants for `Yes`, `No` and `Invalid`

packages/augur-tools/src/flash/fork.ts
bthaile: there is already a util that converts outcome to payout numerators in sdk. use that one. I have a PR up that fixes converting scalar outcome to payout numerators. 

don't use this one. we need the sdk one to be the only one

packages/augur-sdk/src/state/getter/Universe.ts
robert-e-davidson3: It is. I talked to Justin about it - there's no sensible way to avoid this call.

It **definitely** shouldn't be inside a loop though. We only need to query it once!

packages/augur-sdk/src/state/getter/Universe.ts
robert-e-davidson3: Can I just use `isMalformed`?

packages/augur-sdk/src/utils.ts
robert-e-davidson3: I agree but getting it integrated without breaking that getter is out of scope. I'll add a TODO so we get a ticket to do that refactor.

Constants added! I put them in an enum.

packages/augur-sdk/src/utils.ts
robert-e-davidson3: I'm not using this anywhere but I'm not going to refactor the Accounts getters in this PR either so I've left a TODO for that work.

packages/augur-sdk/src/state/getter/Universe.ts
bthaile: isMalformed isn't the same as isInvalid

packages/augur-sdk/src/state/getter/Universe.ts
```diff
@@ -36,6 +58,65 @@ export class Universe {
...
+      },
+    }))[0];
+    const marketTypeName = marketTypeToName(forkingMarket.marketType);
+    const numTicks = Number(forkingMarket.numTicks).toString(10);
+    const minPrice = Number(forkingMarket.prices[0]).toString(10);
+    const maxPrice = Number(forkingMarket.prices[forkingMarket.prices.length - 1]).toString(10);
+
+    const repTokenAddress = await (await augur.contracts.universeFromAddress(params.universe)).getReputationToken_();
+    const repToken = augur.contracts.reputationTokenFromAddress(repTokenAddress, augur.networkId);
+    const amount = (await repToken.totalSupply_()).toString();
```
bthaile, line 93: is this an off chain call ?

packages/augur-sdk/src/utils.ts
bthaile: if I'm reading this correct `getOutcomeFromPayoutNumerators` doesn't handle converting scalar market payout numerators into outcome price

packages/augur-sdk/src/state/getter/Universe.ts
robert-e-davidson3: NEVERMIND!

bthaile:
lots of refactoring needs to occur with the code in this PR, how that we have utils that convert payout numerators to outcome. The code was moved and consolidated and not core to the getter.

packages/augur-sdk/src/utils.ts
```diff
@@ -131,15 +137,23 @@ export function calculatePayoutNumeratorsValue(
...
   }
-  // test if stake payout is malformed
-  if (
-    payout.reduce((p, ticks) => (parseInt(ticks, 10) > 0 ? p + 1 : p), 0) > 1
-  ) {
-    return MALFORMED_OUTCOME;
-  }
+}
+
+function isInvalidCategorical(payout: string[]): boolean {
```
bthaile: this doesn't make sense. 

isInvalidCategorical but is testing for malformed payout numerators. 

Invalid and malformed are different concepts that don't mix

packages/augur-sdk/src/utils.ts
```diff
@@ -131,15 +137,23 @@ export function calculatePayoutNumeratorsValue(
...
+  // test if stake payout is malformed (has ticks in more than one outcome)
+  return payout.reduce((p, ticks) => (Number(ticks) > 0 ? p + 1 : p), 0) > 1;
+}
 
-  return payout.findIndex((item: string) => parseInt(item, 10) > 0).toString();
+function isInvalidScalar(payout: string[]): boolean {
+  // test if stake payout is malformed (has ticks in invalid _and_ another outcome)
+  return Number(payout[0]) > 0 && isInvalidCategorical(payout.slice(1));
```
bthaile: same comment as above

packages/augur-sdk/src/utils.ts
```diff
@@ -120,7 +121,12 @@ export function calculatePayoutNumeratorsValue(
   if (!payout) return null;
   if (payout.length === 0) return null;
 
+
   if (isScalar) {
+    if (isInvalidScalar(payout)) {
+      return MALFORMED_OUTCOME;
```
bthaile, line 133: nope, you are mixing two different concepts.

packages/augur-sdk/src/utils.ts
```diff
@@ -131,15 +137,23 @@ export function calculatePayoutNumeratorsValue(
       .dividedBy(new BigNumber(numTicks, 10))
       .plus(new BigNumber(displayMinPrice, 10))
       .toString();
+  } else {
+    if (isInvalidCategorical(payout)) { // or yes/no
+      return MALFORMED_OUTCOME;
```
bthaile, line 148: nope, same comment as above

bthaile (changes requested):
we need to go over the difference between invalid and malformed.

packages/augur-sdk/src/utils.ts
robert-e-davidson3: yikes fixed

packages/augur-sdk/src/utils.ts
robert-e-davidson3: :(

packages/augur-sdk/src/utils.ts
```diff
@@ -178,3 +209,51 @@ export function calculatePayoutNumeratorsArray(
   }
   return payoutNumerators;
 }
+
+export enum YesNoOutcomes {
```
bthaile: these should be in type file put NEXT to MarketTypeName:

```typescript
export enum MarketTypeName {
  YesNo = 'yesNo',
  Categorical = 'categorical',
  Scalar = 'scalar',
}
```

imo

bthaile (approved):
lgtm

packages/augur-sdk/src/utils.ts
robert-e-davidson3: I like it. Moved!

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 💪

 This link will always point to this PR's latest changes: https://peek.run/jvgh9qm

PR #3543, opened 2019-09-09
Namespace in memory pouchdb instances to avoid collision

justinbarry:
* Modify SDK factories to pass PouchDB factory vs just the configs


robert-e-davidson3 (approved):
ship it!

PR #3544, opened 2019-09-09
click on quantity to populate form

### LINKED ISSUE #1657: Make quantity button clickable
On trading pg in positions...we should make the quantity button clickable....so the quantity fileters into the order ticket...

If a user is Long an outcome...Sell shares should be highlighted and the quantity only fills into the order ticket

If a user is short an outcome and they click on their quantity, buy shares should be highlighted and the quantity only fills into the order ticket

phoebemirman:
closes https://github.com/AugurProject/augur/issues/1657

### DIFF
--- a/packages/augur-ui/src/modules/common/row-column.tsx
+++ b/packages/augur-ui/src/modules/common/row-column.tsx
@@ -62,7 +62,14 @@ function selectColumn(columnType: string, properties: Properties) {
                 location={properties.location}
               />
             }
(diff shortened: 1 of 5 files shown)


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 😄

 This link will always point to this PR's latest changes: https://peek.run/vv2h9xr

bthaile (approved):
lgtm, I like the formatting

PR #3546, opened 2019-09-10
LiquidityDB cleanup

### LINKED ISSUE #3542: LiquidityDB cleanup
During testing using Flash to push the timestamp, I noticed some markets aren't always showing when they should be. This was because the code needs to look at current liquidity in addition to the liquidity value in the DB for the last hour. (Previously, it was just using the last hour's liquidity value.)

Also need to change liquidity DB to be initialized using just the current data and then storing the last 24 hours' data going forward. (Previously, it was attempting to retroactively create the last 24 hours of data when being initialized.)

adrake33:
Closes #3542.


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 📈

 This link will always point to this PR's latest changes: https://peek.run/36whq3o

bthaile (approved):
lgtm

Issue #3549, opened 2019-09-10
Reporting page: scalar market reporting error


bthaile:
![image](https://user-images.githubusercontent.com/3970376/64576831-eaf78080-d33f-11e9-9532-6e430f2ad53e.png)

Getting this error when:
 * inputting valid scalar value between min and max Price.
 * adding additional stake when initially reporting

PR #3550, opened 2019-09-10
Fix reporting rep staked dispute window props

bthaile:
Still a bug with reporting on scalar markets even though price is between min and max. 
Created ticket <https://github.com/AugurProject/augur/issues/3549> to work that issue. 

initial reporting on yes no and category markets works. 

There is still some work with design to figure out best flow when reporting form closes to tell user their report is still being processed.


phoebemirman (approved):

ghost (commented):

 This link will always point to this PR's latest changes: https://peek.run/654h544

PR #3575, opened 2019-09-10
method for liquidity ranking during market creation

nuevoalex:
I made spread included as one of the params even though we originally had that left out. Seemed like that would be nicer for flexibility, also we don't have a default spread on the API side anyway currently.


packages/augur-sdk/src/state/getter/Liquidity.ts
```diff
@@ -0,0 +1,82 @@
...
+import { Getter } from './Router';
+import { Augur } from '../../index';
+import * as _ from 'lodash';
+
+export const Order = t.type({
+    price: t.string,
+    amount: t.string,
+});
+  
+export const OutcomeOrderBook = t.type({
```
bthaile, line 13: Is this OutcomeOrderBook different than in:

packages/augur-sdk/src/api/Liquidity.ts

packages/augur-sdk/src/state/getter/Liquidity.ts
```diff
@@ -0,0 +1,82 @@
...
+  @Getter('getMarketLiquidityRankingParams')
+  static async getMarketLiquidityRanking(augur: Augur, db: DB, params: t.TypeOf<typeof Liquidity.getMarketLiquidityRankingParams>): Promise<MarketLiquidityRanking> {
+    const liquidityScore = await augur.liquidity.getLiquidityForSpread({
+        orderBook: params.orderBook,
+        numTicks: new BigNumber(params.numTicks),
+        marketType: params.marketType,
+        reportingFeeDivisor: new BigNumber(params.reportingFeeDivisor),
+        marketFeeDivisor: new BigNumber(params.marketFeeDivisor),
+        numOutcomes: params.numOutcomes,
+        spread: params.spread,
```
bthaile, line 52: Can this be calculated instead of passed in? Since the orderbook is a parameter

packages/augur-sdk/src/state/getter/Liquidity.ts
```diff
@@ -0,0 +1,82 @@
...
+    });
+    const totalMarkets = (await db.getNumRowsFromDB("Markets", true)); // Includes this market. Would otherwise have to -1 to remove the meta row
+    if (liquidityScore.isZero()) {
+        return {
+            marketRank: 0,
```
bthaile, line 57: Zero here means not ranked? UI can assume not ranked. UI should capture if there is no orders in the book or if one side is missing.

packages/augur-sdk/src/state/getter/Liquidity.ts
```diff
@@ -0,0 +1,82 @@
...
+            totalMarkets,
+            hasLiquidity: false
+        };
+    }
+    const key = `liquidity.${params.spread}`;
```
bthaile, line 62: Is there any precision  limitations here? Like 9.1284726274859 as an example. Or is only used as unique identifier

packages/augur-test/src/tests/state/getter/Liquidity.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    // With no orders on the book the liquidity scores won't exist
+    await (await db).sync(john.augur, mock.constants.chunkSize, 0);
+    let marketData = await (await db).findMarkets({selector: { market: market.address }});
+    await expect(marketData[0].liquidity).toEqual(undefined);
+
+    // Place a Bid on A and an Ask on A
+    const bid = new BigNumber(0);
+    const ask = new BigNumber(1);
+    let numShares = new BigNumber(10**18);
```
bthaile, line 63: Do orders (orderBook) passed from UI need to be in on-chain format?

packages/augur-test/src/tests/state/getter/Liquidity.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    liquidityRanking = await api.route('getMarketLiquidityRanking', liquidityRankingParams);
+    await expect(liquidityRanking.marketRank).toEqual(0);
+    await expect(liquidityRanking.totalMarkets).toEqual(2);
+    await expect(liquidityRanking.hasLiquidity).toEqual(false);
+
+    // Place lesser liquidity. Ranks second place
+
+    liquidityRankingParams.orderBook[1] = {
+        bids: [{
+            price: "51",
```
bthaile, line 88: And price? Does it need to be in on-chain values?

packages/augur-sdk/src/state/getter/Liquidity.ts
adrake33: They are the same data shape, but one is an interface, and the other is defined for type-checking when passed into the getter.

packages/augur-sdk/src/state/getter/Liquidity.ts
nuevoalex: I don't understand what you mean. The provided spread is a parameter used to evaluate an orderbook's liquidity. It cant be derived

packages/augur-sdk/src/state/getter/Liquidity.ts
nuevoalex: Zero is not ranked, correct

packages/augur-sdk/src/state/getter/Liquidity.ts
nuevoalex: we only use the values 0,10,15,20

packages/augur-test/src/tests/state/getter/Liquidity.test.ts
nuevoalex: Yes

packages/augur-test/src/tests/state/getter/Liquidity.test.ts
nuevoalex: Yea

packages/augur-sdk/src/state/getter/Liquidity.ts
adrake33: This is just a value that has to be specified. `getLiquidityForSpread` returns a different value, depending on the spread that's passed in.

bthaile: I get it. UI knows the default spread that needs to be passed. Gotcha

packages/augur-sdk/src/state/getter/Liquidity.ts
```diff
@@ -0,0 +1,82 @@
...
+        if (market.liquidity) {
+            return numHigher + (liquidityScore.lt(market.liquidity[params.spread]) ? 1 : 0);
+        }
+        return numHigher;
+    }, 0);
+    return {
+        marketRank: higherRankingMarkets + 1,
+        totalMarkets,
+        hasLiquidity: true 
+    };
```
adrake33, line 78: Maybe we should also add `hasRecentlyDepletedLiquidity` as a value for this? I created Markets.setHasRecentlyDepletedLiquidity for depleted-liquidity filtering, but maybe this getter could be used for that instead with a few tweaks.

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🎉

 This link will always point to this PR's latest changes: https://peek.run/36whqeo

packages/augur-sdk/src/state/getter/Liquidity.ts
nuevoalex: I dont think that would make sense here since this is for a market that doesnt even exist yet

bthaile (approved):
Lgtm

packages/augur-sdk/src/state/getter/Liquidity.ts
adrake33: Ah right, never mind.

adrake33 (approved):
👍

Issue #2498, opened 2019-06-17
Market Creation Liquidity integration


bthaile:
Here is the backend PR, it's informative <https://github.com/AugurProject/augur/pull/3575>

impl design for market creation liquidity page. This includes calling middleware for getting ranking and results of liquidity passing default spread filter.

UI knows the default liquidity spread. Market creation liquidity will inform the user about:
 * tell user if their liquidity spread won't pass default spread filter.
 * tells user they need to add a bid or ask if one doesn't exist.

Need sdk method/getters to call to get two pieces of UI data:
 * given the user's current liquidity (order book). What ranking will the market be given.

See PR linked above to exact inputs and outputs
```typescript
export interface Order {
  price: string;
  amount: string;
}

export interface OutcomeOrderBook {
  bids: Order[];
  asks: Order[];
}

export interface OrderBook {
  [outcome: number]: OutcomeOrderBook;
}

interface getMarketLiquiditySortPositionParams {
  orderBook: OrderBook;
  numTicks: BigNumber;
  marketType: MarketType;
  reportingFeeDivisor: BigNumber;
  marketFeeDivisor: BigNumber;
  numOutcomes: number;
}

result: 

export interface MarketRanking {
  marketRank: number;
  totalMarkets: number;
}
```
design area of mocks
<https://www.figma.com/file/IpIAYVQPeQfuT5TTUboQpo4S/Market-creation?node-id=781%3A960>

![image](https://user-images.githubusercontent.com/3970376/64392656-6214d800-d013-11e9-96fe-c0aba6ca2a63.png)

![image](https://user-images.githubusercontent.com/3970376/64392662-680ab900-d013-11e9-9f36-0c37056676ab.png)

![image](https://user-images.githubusercontent.com/3970376/64392678-70fb8a80-d013-11e9-8942-2e22f4e24e05.png)

pacif1c:
One thing I’ve been thinking for a while would be useful, and maybe this is already the plan, is that on the market creation page you can plug in different orders with different spreads + volume and see instantly where the market would rank in the sort. 

So by default it may say something like “Currently your market will not be visible to most users. Add liquidity to make your market more prominent.” Then as soon as I add some orders it will update to say something like, “your market will now rank 5th.” Ideally it could even say “add X amount to rank 4th.” Imo we want market creators competing for the top few spots.

[...]

PR #3618, opened 2019-09-12
market-liquidity-integration

JohnDanz:
https://github.com/AugurProject/augur/issues/2498

This is WIP, still needs to be refactored and tested -- want to get styling fix in and exposure on the augurSDK object for getMarketLiquidityRanking.

started work on liquidity integration, added getMarketLiquidityRanking to augurSDK object, updated some styling for tablets on dispute View


packages/augur-ui/src/modules/create-market/components/visibility.tsx
```diff
@@ -1,40 +1,178 @@
...
+      settlementFee,
+    } = newMarket;
+    const tickSizeBigNumber = createBigNumber(tickSize);
+    const numTicks = tickSizeToNumTickWithDisplayPrices(
+      tickSizeBigNumber,
+      minPriceBigNumber,
+      maxPriceBigNumber
+    );
+    let marketTypeNumber = 0;
+    switch (marketType) {
```
bthaile, line 70: add a new method in sdk utils. to convert from marketTypeName to marketType and use their constants

packages/augur-ui/src/modules/create-market/components/visibility.tsx
```diff
@@ -1,40 +1,178 @@
...
+      });
+    });
+		const params = {
+			orderBook: formattedOrderBook,
+      numTicks: numTicks.toFixed(),
+      numOutcomes: outcomesFormatted.length,
+      marketType: marketTypeNumber,
+      marketFeeDivisor: `${settlementFee}`,
+      reportingFeeDivisor: '0',
+			spread: 10,
```
bthaile, line 112: let's not hard code this here. and we have constants for that value

packages/augur-ui/src/modules/create-market/components/visibility.tsx
```diff
@@ -1,40 +1,178 @@
...
+          formattedOrderBook[outcome].asks.push(formattedOrder);
+        }
+      });
+    });
+		const params = {
+			orderBook: formattedOrderBook,
+      numTicks: numTicks.toFixed(),
+      numOutcomes: outcomesFormatted.length,
+      marketType: marketTypeNumber,
+      marketFeeDivisor: `${settlementFee}`,
```
bthaile, line 110: can prob just have this 0, not sure its used in calculation and it would be different than just settlementFee

packages/augur-ui/src/modules/create-market/actions/get-market-liquidity-ranking.ts
```diff
@@ -0,0 +1,13 @@
+import { augurSdk } from 'services/augursdk';
+import logError from 'utils/log-error';
+import { NodeStyleCallback } from 'modules/types';
+
+export const getMarketLiquidityRanking = async (
```
bthaile, line 5: I don't think pulling this out into a separate file gives us anything. it's only going to be called from one place. and this is a lot of boilerplate for no reason. 
and we are trying to move away from callbacks for simple calls like this

bthaile (changes requested):
a few comments

bthaile (approved):
these comments can be addressed in follow up PR

packages/augur-ui/src/modules/create-market/actions/get-market-liquidity-ranking.ts
JohnDanz: for sure, This is all a WIP, I'll refactor/refine all of this, I was on the fence on if this needed it's own file but it doesn't so ill remove this and put it into visibility or the container.

packages/augur-ui/src/modules/create-market/components/visibility.tsx
JohnDanz: settlementFee here is the marketCreatorFee so I gotta include it, I'll double check on if it effects the calcs in the getter in a follow up PR

packages/augur-ui/src/modules/create-market/components/visibility.tsx
JohnDanz: I'll adjust in follow up PR

packages/augur-ui/src/modules/create-market/components/visibility.tsx
JohnDanz: Will do, I think we might already have this somewhere, but i was just trying to get my quick and dirty implementation in place to prove out what needed to be done before i refactor for the final product

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🚀

 This link will always point to this PR's latest changes: https://peek.run/5m3hgk7

PR #3620, opened 2019-09-12
Alert initial work

phoebemirman:
set up toast for four notification types, and some other alerts


bthaile (approved):
lgtm, some issues will be addressed in next PR

PR #3625, opened 2019-09-13
values from getter are in atto value, need to convert to display values

bthaile:
image

should be this:

image

same for dispute window participation token balance.


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 😎

 This link will always point to this PR's latest changes: https://peek.run/zr5hv4z

andrewdamelio (approved):
looks good!
image

PR #3668, opened 2019-09-16
FB comments

### LINKED ISSUE #3660: market trading chat
use facebook comments for each market 

https://developers.facebook.com/docs/plugins/comments/

Create a standalone component that takes in marketId. It'll be added to market trading page as an integration task. Of course you can add it to market trading page to ease development but needs to have it's own container. It'll need to be responsive. 

<https://www.figma.com/file/fLWVwmanAwetVZbujQquEi/Market-Page?node-id=390%3A134>

use `market://marketId` as the href

use dark theme

E0E5F926-F3FC-48DF-89FB-C39E25F646FF

adrake33:
Closes #3660.


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🏆

 This link will always point to this PR's latest changes: https://peek.run/kq3hzpo

packages/augur-ui/src/modules/market/components/common/comments/market-comments.tsx
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  static defaultProps = {
+    marketId: PropTypes.string.isRequired,
+    colorScheme: PropTypes.string.isRequired,
+    dataWidth: PropTypes.string.isRequired,
+    networkId: PropTypes.string.isRequired,
+  };
+
+  constructor(props) {
```
bthaile: constructor isn't needed

packages/augur-ui/src/modules/market/components/common/comments/market-comments.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export default class MarketComments extends Component {
+  static propTypes = {
+    marketId: PropTypes.string.isRequired,
+    colorScheme: PropTypes.string.isRequired,
+    dataWidth: PropTypes.string.isRequired,
+    networkId: PropTypes.string.isRequired,
+  };
+
+  static defaultProps = {
```
bthaile: wrong syntax for defaultProps.

packages/augur-ui/src/modules/market/components/common/comments/market-comments.tsx
```diff
@@ -0,0 +1,60 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+export default class MarketComments extends Component {
+  static propTypes = {
```
bthaile: we don't use propTypes anymore we use typescript style. There are a lot of examples.

image

image

packages/augur-ui/src/modules/market/components/common/comments/market-comments.tsx
```diff
@@ -0,0 +1,60 @@
...
+      colorScheme,
+      dataWidth,
+      networkId,
+    } = this.props;
+
+    let content = null;
+
+    content = (
+      <div id='fb-comments-container'>
+        <div id='fb-comments' className='fb-comments' data-colorscheme={colorScheme} data-href={'www.augur.net/comments/' + networkId + '/' + marketId} data-width={dataWidth} data-numposts='10'></div>
```
bthaile: don't use `+` to concat strings use ` ${...}` it's cleaner code.

packages/augur-ui/src/modules/market/components/common/comments/market-comments.tsx
```diff
@@ -0,0 +1,60 @@
...
+      </div>
+    );
+
+    if (!window.document.getElementById('fb-root')) {
```
bthaile: really don't like using window.document.  why need to find `fb-root` ?

packages/augur-ui/src/modules/market/components/common/comments/market-comments.tsx
```diff
@@ -0,0 +1,60 @@
...
+      const fbRoot = window.document.createElement('div');
```
bthaile: Is this needed to create div like this? can just using React work

like const content {
<div id='fb-root'>
</div>
}

return (<> {content}</>)


also where is the marketId ? seems this will only work once? 

also need a constant for the src url since it looks like a magic string to me.

adrake33: I was initially doing it that way and then was trying some different things to see what would allow the comments to load, but I forgot to change it back. Will do that.

bthaile (approved):
lgtm, moved some stuff around looks like tiny bit cleaner

PR #3676, opened 2019-09-17
fix type issue when using max liquidity spread filter

bthaile:
fixes type error in code, also was seeing  console error.

image
 

image

backend didn't like the string passed in for maxLiquiditySpread. it needed to be typed.


pgebheim (commented):
Id split out that Union and make a med type and use that in both the interface as well as when casting it.

bthaile:
Tried doing that I couldnt figure out the syntax. Help would greatly be appreciated

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🍕

 This link will always point to this PR's latest changes: https://peek.run/xjlhk55

pgebheim (approved):
Yes like that!

PR #3682, opened 2019-09-17
Add inline "forking" to ganache provider

justinbarry:
This allows us to reuse a base state for use multiple tests.


nuevoalex (approved):
fork in this sense being a forked chain, not an Augur fork ;p

robert-e-davidson3 (approved):
lgtm

justinbarry:
Maybe we call a “spork” or something.

PR #3689, opened 2019-09-17
Update getUserTradingPositions for claiming proceeds

### LINKED ISSUE #1800: Changes to user positions getter to include winning amount
Add new property `winnings` to `TradingPosition` type.

`TradingPosition` is is served from `getUserTradingPositions` getter. 

This property would get populated when the user can claim winnings based on a profitable position. This is determined when current block time is greater than market's dispute windows end time and user has a position on the current tentative winning outcome.

adrake33:
Closes #1800 and #3656 and #3725


packages/augur-sdk/src/state/getter/Users.ts
```diff
@@ -473,6 +472,28 @@ export class Users {
       }
     );
 
+    const derivedDbMarketsInfos = await db.findMarkets({
+      selector: { market: { $in: marketIds } },
+    });
+    for (const derivedMarketDbInfo of derivedDbMarketsInfos) {
+      if (derivedMarketDbInfo.reportingState === MarketReportingState.Finalized || MarketReportingState.AwaitingFinalization) {
+        const derivedDbDisputeInfo = await db.findDisputeDocs({
```
nuevoalex: This could be one query outside of the loop to get all records for the markets in marketIds

packages/augur-sdk/src/state/getter/Users.ts
```diff
@@ -473,6 +472,28 @@ export class Users {
...
+          selector: { market: { $eq: derivedMarketDbInfo.market } },
+        });
+        const reportingFeeDivisor = derivedMarketDbInfo.feeDivisor;
+        // TODO: set numberOfShares based on winning/tentative payout
+        const reportingFee = numberOfShares.div(reportingFeeDivisor);
+        const contracts = augur.contracts;
+        const totalUnclaimedProceeds = (await contracts.claimTradingProceeds.calculateProceeds_(derivedMarketDbInfo.market, new BigNumber(tentativeOutcome), numberOfShares))
```
nuevoalex: This is going to do a call per market. We should add this information to the logs instead

packages/augur-sdk/src/state/getter/Users.ts
adrake33: I think I tried that, but it was giving me an error when I tried to use `$in` in the query. Will try again to see what the exact error was.

packages/augur-sdk/src/state/getter/Users.ts
nuevoalex: Alternatively you can calculate this with the existing data in the market using the winningPayoutNumerators pretty easily.

packages/augur-sdk/src/state/getter/Users.ts
adrake33: Actually I don't think I need to get the dispute docs, do I? I should be able to get the winning/tentative payout now without doing that.

packages/augur-sdk/src/state/getter/Users.ts
adrake33: True, but if a market is awaiting finalization, it won't be set, and I'll need to use tentativeWinningPayoutNumerators  instead, correct?

nuevoalex: thats correct. Actually since these are finalized or awaiting finalization you could always just use that

packages/augur-sdk/src/state/getter/Users.ts
nuevoalex: yup, true

packages/augur-sdk/src/state/getter/Users.ts
adrake33: I'm still unclear on how to get the number of shares in certain cases. For example, if the user is short a particular outcome, they will have a net position that's negative, so that should be considered a positive number of shares in all other outcomes? How should scalar markets be handled?

nuevoalex: You can get balances from the users TokenBalanceChanged db. The records have a token type and a market so you can get the current balance of all shares for a market.

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 😎

 This link will always point to this PR's latest changes: https://peek.run/qo0h6lp

packages/augur-sdk/src/state/getter/Users.ts
```diff
@@ -473,6 +472,41 @@ export class Users {
       }
     );
 
+    // Set unclaimedProceeds & unclaimedProfit
+    const marketsData = await db.findMarkets({
```
nuevoalex: really we should just replace the `findMarketCreatedLogs` way above with this

adrake33: Yeah it's not needed anymore. I'll remove it.

packages/augur-sdk/src/state/getter/Users.ts
```diff
@@ -473,6 +472,41 @@ export class Users {
...
+      selector: { market: { $in: marketIds } },
+    });
+    for (const marketData of marketsData) {
+      marketTradingPositions[marketData.market].unclaimedProceeds = '0';
+      marketTradingPositions[marketData.market].unclaimedProfit = '0';
+      if (marketData.reportingState === MarketReportingState.Finalized || MarketReportingState.AwaitingFinalization) {
+        if (marketData.tentativeWinningPayoutNumerators) {
+          const outcome = marketData.tentativeWinningPayoutNumerators.findIndex((item: string) => Number(item) > 0);
```
nuevoalex: This won't work for scalars

packages/augur-sdk/src/state/getter/Users.ts
```diff
@@ -473,6 +472,41 @@ export class Users {
...
+          const tokenBalanceChangedLogs = await db.findTokenBalanceChangedLogs(
```
nuevoalex: We should be doing a single query for all markets above are arranging it into a mapping so we can get share balances for each non zero winning outcome

nuevoalex (changes requested):
Made some comments on the algo

packages/augur-sdk/src/state/getter/Users.ts
adrake33: Yeah I'll use calculatePayoutNumeratorsValue instead

nuevoalex: No, I mean in scalars there are multiple non zero outcome values

adrake33: calculatePayoutNumeratorsValue won't return an outcome that can be used to look up the TokenBalanceChangedLog? How do you suggest getting the outcome to use for looking up the logs?

nuevoalex: You will have multiple outcomes, not a single outcome. For any non zero outcome in `tentativeWinningPayoutNumerators` you would retrieve the associated log for the market and that outcome. I describe this below

adrake33: Ok, I see now. Makes sense.

nuevoalex (approved):
👍

bthaile (approved):
lgtm

packages/augur-ui/src/modules/positions/selectors/login-account-claimable-winnings.ts
```diff
@@ -0,0 +1,77 @@
...
+};
+
+// need to add marketInfos in case positions load before markets
+const getLoginAccountPositionsMarkets = createSelector(
+  selectAccountPositionsState,
+  selectMarketInfosState,
+  (positions, markets) => {
+    console.log(
+      'selectLoginAccountClaimablePositions -> getLoginAccountPositionsMarkets'
+    );
```
adrake33: @bthaile Should this console.log be left here?

adrake33 (commented):
@bthaile Your changes look good. 👍

PR #3713, opened 2019-09-18
color-theme-preparation

JohnDanz:
This is another step toward cleaning up the colors in preparation for the styles.

refactored out some colors, removed some unused icons, adjusted webpack config to ignore `*.d.ts` files as "unused", did a full audit of colors and added comments for number of files for each color to help us narrow things down, need to discuss with design. fixed some file names from style.less to styles.less to fit conventions. removed _inputs.less and put it directly into shared, removed 2 classes in shares that were unused.

More discussion is needed with design to come to an agreement on certain colors we will need which haven't been outlined yet here: https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=10500%3A302591


bthaile (approved):
lgtm, moving right along. we'll get there

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 😄

 This link will always point to this PR's latest changes: https://peek.run/654h22z

PR #3715, opened 2019-09-18
Force refresh reporting

### LINKED ISSUE #3561: Reporting Page: trigger reporting page refresh
Need to signal to reporting page (if user is currently on this page) when markets move into designated reporting or open reporting. This way searches can be triggered and deliver results. 

this will provide dynamic UX. Need to figure out this arch.

### LINKED ISSUE #3642: Disputing and Reporting Page: log handlers, pagination refresh annoyance
Currently arch. Reporting page and Dispute page use `getMarkets` getter pagination. 

Then markets are stored in state (MarketInfos) the marketIds are stored in state (ReportingList) and selectors are used to populate Viz components. This way the market cards will update if a log-handler comes in and forces a refresh of MarketInfo.

Issue is:

If the market reporting state updates and the UI pulls the new marketInfo, the market card will update on disputing and reporting page but the market will stay in the list. It won't leave the list unless the user navigates off that page or goes to the next page in pagination. 

I think minor annoyance but wanted to capture it here.

approach: add same parameters that were used to paginate and simply re-call getMarkets.

bthaile:
fixes https://github.com/augurproject/augur/issues/3642
fixes https://github.com/augurproject/augur/issues/3561

verify that reporting page and dispute page refreshes when the user is on the page and a dispute or report was performed. 

still working through sorting issues on reporting market cards. the changes to market.ts is an ongoing task.


packages/augur-sdk/src/utils.ts
```diff
@@ -134,32 +133,43 @@ export function calculatePayoutNumeratorsValue(
     }
```
robert-e-davidson3: The line above this one also needs `outcome: null`

bthaile: updated

robert-e-davidson3 (approved):
I can't attest to the UI changes but the sdk changes LGTM

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). ✨

 This link will always point to this PR's latest changes: https://peek.run/p37hvvp

PR #3757, opened 2019-09-20
fix issue by removing weird log item we dont use

nuevoalex:
This log param was nonsense (adding ETH balance and DAI gained) and caused gas estimation to fail, probably because of checking the users balance (bug in geth perhaps)


bthaile (approved):
Lgtm

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🍭

 This link will always point to this PR's latest changes: https://peek.run/l74hxpm

PR #3759, opened 2019-09-21
Fix dispute modal values

### LINKED ISSUE #3629: Disputing Page: `Fill Dispute Bond` amount causes dispute to fail
Need to use `stakeRemaining` to populate REP amount.

clicking on `fill dispute bond` fills in REP amount but that amount causes contract call to fail. 

needs investigation, sending more than bond shouldn't cause failure, but does.

adrake33:
Closes #3629


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 💙

 This link will always point to this PR's latest changes: https://peek.run/o9phx3e

packages/augur-ui/src/modules/common/form.tsx
```diff
@@ -940,17 +940,21 @@ export class ReportingRadioBar extends Component<ReportingRadioBarProps, {}> {
...
     if (scalar) {
       stake = {
-        preFilledStake: formatRep("0"),
-        bondSizeCurrent: formatRep("1"),
-        bondSizeCurrent: formatRep("1"),
-      }
+        outcome: null,
+        stakeCurrent: '0',
+        bondSizeCurrent: '1',
+        stakeRemaining: '1',
```
bthaile: These are hardcoded values, used for dev. Replace with actual values from getter

adrake33: Ok, will update later today.

bthaile (changes requested):
already conveyed needed changes

packages/augur-ui/src/modules/common/form.tsx
adrake33: Actually, I think this is only used for the "Other Outcomes radio button":

Screen Shot 2019-09-22 at 11 40 41 AM

For these outcomes, isn't it ok to use hard-coded values since they won't have any stake on them yet?

adrake33: Never mind, I think you mean for the bondSizeCurrent/stakeRemaining, not the stakeCurrent. I think this will require adding a new prop for ReportingRadioBar?

bthaile: Really we need to use StakeDetails and not have the UI create its own internal object and pass it around.  These UI components were created before the new StakeDetails properties rename

bthaile: Also the UI dispute components will need to know bond size new outcome value that is on the parent object of stakeDetails. Its used when an outcome hasnt been staked on.

adrake33: Isn't this info already in state.marketInfos[marketId].disputeInfo?

bthaile: Yes thats what I’m referring to. Better if I would have said use the same types from the getMarkets/getMarketsInfo getters.

bthaile: For example, sdk doesnt diff between pre-filled and stakeCurrent. imo, the UI shouldnt in its variables. We do in the viz components.

adrake33: Ok, seems like ModalReporting should be passing in the disputeInfo for the market into RadioBarGroup as a prop, which should then be passed around similarly from there?

bthaile: Phoebe and talked about how the components are put together. Really containers should be used and only minimum props are passed. Can you organize the components? Good op to learn child prop stuff and containers.

adrake33: Sure, I can do that. My wife and I have plans today, so I probably won't be able to finish by tomorrow, but I'll start working on that later today.

bthaile (approved):
lgtm, we'll keep refining as we fix issues that pop up.

PR #3830, opened 2019-09-25
color-adjustments

JohnDanz:
fixed colors for filter notification on market-list and updated invalid outcome name on trading page to be the invalid color


bthaile (approved):
approved

PR #3849, opened 2019-09-26
add a unique id that we compare when updating and deleting

### LINKED ISSUE #3847: alerts have overlapping ids which makes it so you can't edit them

phoebemirman:
closes https://github.com/AugurProject/augur/issues/3847

now alerts have an id/txHash and a uniqueId 
- in most cases uniqueId is the same as the id/txHash, except for orders and whatever else we need down the line, because multiple orders can have the same txHash
- uniqueId is used to compare for updating and deleting and to tell if the alerts are the same
- id/txhash isn't always the txHash, for market alerts it is the marketId and cancel order alerts it is the orderId

also removed some calls to update and add alert that we don't need anymore

### DIFF
--- a/packages/augur-ui/src/modules/account/actions/get-rep.ts
+++ b/packages/augur-ui/src/modules/account/actions/get-rep.ts
@@ -1,4 +1,3 @@
-import { updateAlert } from "modules/alerts/actions/alerts";
 import { selectCurrentTimestampInSeconds as getTime } from "store/select-state";
 import logError from "utils/log-error";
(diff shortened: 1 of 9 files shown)


adrake33 (approved):
lgtm

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🧡

 This link will always point to this PR's latest changes: https://peek.run/vv2hzr1

PR #3854, opened 2019-09-26
sets npm memory usage so we don't OOM error

petong:
fixes OOM error in CI


bthaile (approved):
seems we are putting that every where. lgtm

PR #3882, opened 2019-09-27
get categories back into the UI

andrewdamelio:
Quick and dirty fix to get categories back in the UI -- there may be a better way to go about this but I'd rather have categories in the UI in the meantime.


bthaile (approved):
lgtm

PR #3885, opened 2019-09-27
gnosis relay test against local relay

robert-e-davidson3:


packages/augur-core/source/libraries/ContractDeployer.ts
```diff
@@ -245,7 +245,7 @@ Deploying to: ${networkConfiguration.networkName}
         return contract.address;
     }
 
-    private async uploadAllContracts(serial=false): Promise<void> {
+    private async uploadAllContracts(serial=true): Promise<void> {
```
robert-e-davidson3, line 248: @nuevoalex how do you feel about defaulting to serial uploads? I haven't been able to deploy to ganache from a separate process (like flash run deploy) when this is run in parallel because of nonce errors. I added the `serial` parameter and code so I could easily tweak it for that kind of test but I haven't worked out how to actually pass in the right argument here.

I assume running in parallel is fantastic with geth and such because that fits many txs into one block.

packages/gnosis-relay-api/src/tests/GnosisRelayAPI.test.ts
```diff
@@ -125,6 +132,12 @@ test('Gnosis Relay API:: Make safe and do transactions', async () => {
     await expect(txHash).not.toEqual(undefined);
 
     console.log(`Waiting on TX: ${txHash}`);
-    await provider.getTransactionReceipt(txHash);
+    const receipt = await provider.getTransactionReceipt(txHash);
+    expect(receipt).toMatchObject({
+      to: safeAddress,
+      status: 1,
+      contractAddress: null,
```
robert-e-davidson3, line 139: I don't know what `contractAddress` is supposed to be so I nailed it down to the value I get. If this should be non-null then there's an issue.

packages/augur-core/source/libraries/ContractDeployer.ts
nuevoalex: as long as its not extremely painfully slow thats fine. The purpose of the Fast Wallet is supposed to be to allow non serial uploads but I've also seen the occasional nonce error so probably better to just suck it up. Not doing too much iteration with the integration tests anyway.

packages/gnosis-relay-api/src/tests/GnosisRelayAPI.test.ts
nuevoalex: That would be if a contract was deployed with the tx so yea should be null here!

nuevoalex (approved):
👍

packages/gnosis-relay-api/src/tests/GnosisRelayAPI.test.ts
robert-e-davidson3: sweet!

Issue #3902, opened 2019-09-28
Market Card: long market title is behind outcomes


bthaile:
market title is behind the outcomes:
![image](https://user-images.githubusercontent.com/3970376/65819732-a431d380-e1e5-11e9-9400-39126aa64f72.png)

you can see the market title is very long

![image](https://user-images.githubusercontent.com/3970376/65819749-d2afae80-e1e5-11e9-8446-d65abfd92cf7.png)

PR #3934, opened 2019-09-30
Should only increase overall height when there are three lines of text

bconfortin:
With auto in the middle row, the market title only goes as far as 77px when the title has 3 rows of text.
When collapsed, 1 row of text should only occupy the height of 1 row, and so forth.
https://github.com/AugurProject/augur/issues/3902


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 💖

 This link will always point to this PR's latest changes: https://peek.run/n5xh5kn

bthaile (approved):
lgtm

PR #3971, opened 2019-10-01
added getCategoryStats getter

robert-e-davidson3:


bthaile (approved):
lgtm, well need to take it out for a spin. Will do when we wire up the template landing page

PR #3973, opened 2019-10-01
contract changes. Back to single trade contract

nuevoalex:
The branch name is misleading somewhat since zeroX was already using mesh and we're not really changing anything with it on the middleware side.

This is just packing more data into the token id so that we only have one contract to manage and later filter on as well as adding the exchange as a part of the signed order data so that it can be changed in the future.


packages/augur-core/source/contracts/trading/ZeroXTrade.sol
```diff
@@ -86,6 +84,89 @@ contract ZeroXTrade is Initializable, IZeroXTrade {
...
+        for (uint256 _i = 0; _i < _numberOfOutcomes; _i++) {
+            if (_i != _outcome) {
+                uint256 _creatorShareTokenBalance = _market.getShareToken(_i).balanceOf(_owner);
+                _attoSharesOwned = _creatorShareTokenBalance.min(_attoSharesOwned);
+            }
+        }
+
+        uint256 _attoSharesPurchasable = cash.balanceOf(_owner).div(_price);
+
+        return _attoSharesOwned.add(_attoSharesPurchasable);
```
robert-e-davidson3, line 139: Can this overload?

nuevoalex: Doubtful since youd have to have an impossibly large amount of cash or shares, but if it did it would just make the tx fail as we're using safe math

robert-e-davidson3 (approved):
lgtm!

Issue #3954, opened 2019-10-01
UTC calculation seems broken on both market creation and market trading pg


Chwy5:
![10-1UTCbroken.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/a67ed23e-37da-443b-ad82-1557a099a50b)

I put in 11:30 pm EST...shows EDT...I think UTC should be 4 hours forward and it looks to be displaying it 4 hours backwards

PR #4001, opened 2019-10-02
fix timezone offset calc 

bthaile:
was applying offset to local time should have been inverting offset and applying to get utc time.

fixed issue with applying offset incorrectly. convert to use `useState` and `userEffect`

https://github.com/AugurProject/augur/issues/3954
 * missed checking in some checks for reporting form
 * fixed issue where auto complete didn't open list when user clicked on input.
 * fixed a few color issues
image

image


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🖖

 This link will always point to this PR's latest changes: https://peek.run/0elhp69

phoebemirman (approved):

PR #4009, opened 2019-10-02
various contract changes

nuevoalex:
- Add some validation to zeroX trades and processing in middleware
- Add events for when fee rates change
- Fix OICash to track fees paid rather than OI which is no longer subject to fees
- Make Gnosis Registry call do approval to save a tx for the user


packages/augur-core/source/contracts/Augur.sol
```diff
@@ -90,6 +90,10 @@ contract Augur is IAugur {
     event ProfitLossChanged(address indexed universe, address indexed market, address indexed account, uint256 outcome, int256 netPosition, uint256 avgPrice, int256 realizedProfit, int256 frozenFunds, int256 realizedCost, uint256 timestamp);
     event ParticipationTokensRedeemed(address indexed universe, address indexed disputeWindow, address indexed account, uint256 attoParticipationTokens, uint256 feePayoutShare, uint256 timestamp);
     event TimestampSet(uint256 newTimestamp);
+    event ValidityBondChanged(address indexed universe, uint256 validityBond);
+    event DesignatedReportStakeChanged(address indexed universe, uint256 designatedReportStake);
+    event NoShowBondChanged(address indexed universe, uint256 noShowBond);
+    event ReportingFeeChanged(address indexed universe, uint256 reportingFee);
```
bthaile, line 96: what is the expected usage for these event? In UI, the reporting fee, validity bond and no-show bond amounts are retrieved when the user goes into market creation workflow. 

Would the UI need to take any action when these logs get emitted?

bthaile (approved):
lgtm

PR #4013, opened 2019-10-03
added conv method to deploy and create markets

bthaile:
one command to deploy contracts with fake time and create canned markets.


adrake33 (approved):
Lgtm

PR #4036, opened 2019-10-04
Convert `propTypes` to typescript version of props

gitstart-augur:
Converted the old way of handling props to 
interfaces.


bthaile (changes requested):
Files should not be deleted. Only changes are in packages/augur-ui

gitstart-augur:
@bthaile fixing..Reopening a new PR.

PR #4057, opened 2019-10-05
Fix participation token dispute markets

bthaile:
changed participation tokens to use new structure from getter. 

so these values show
image

using isClaimable it auto filters out current dispute window and markets not resolved.
image

### DIFF
--- a/packages/augur-ui/src/modules/reporting/components/markets-in-dispute.tsx
+++ b/packages/augur-ui/src/modules/reporting/components/markets-in-dispute.tsx
@@ -22,17 +22,25 @@ const TAB_CURRENT = 'current';
 const TAB_AWAITING = 'awaiting';
 const SORT_REP_STAKED = 'totalRepStakedInMarket';
 const SORT_DISPUTE_ROUND = 'disputeRound';
+const defaultTabs = {
+  [TAB_CURRENT]: [],
+  [TAB_AWAITING]: [],
+};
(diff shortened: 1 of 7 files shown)


packages/augur-ui/src/modules/reporting/components/markets-in-dispute.tsx
```diff
@@ -125,6 +133,10 @@ export default class MarketsInDispute extends Component<
     }
   }
 
+  componentWillUnmount() {
+    console.log("component has unmounted");
+  }
```
adrake33, line 138: Should this be removed?

adrake33 (approved):
👍

packages/augur-ui/src/modules/reporting/components/markets-in-dispute.tsx
bthaile: I need it for troubleshooting. every once in a while markets-in-dispute blows up.

bthaile: I think something is unmouting it while a load markets call is returning values. still thinking about it.

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🍉

 This link will always point to this PR's latest changes: https://peek.run/0elhz9r

PR #4062, opened 2019-10-06
Fix Universe Picker console error

adrake33:
Fixes a console error saying that keys must be specified when using map.


bthaile (approved):
Lgt

PR #4064, opened 2019-10-06
added fake-all and normal-all scripts

bthaile:
short cuts for deploying contracts and canned markets.
fake time and normal time

`fake-all`
`normal-all`

scripts added to flash


adrake33 (approved):
Lgtm

PR #4065, opened 2019-10-07
add the various notices to market cards with button actions

bthaile:
Message on market cards when they are in pre-reporting
image

User has REP staked on market
image

disputing markets during fork. text is a little off. didn't figure it out.
image

labeled forking market (wasn't in design)
image

User needs REP on winning universe to migrate
image


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 💫

 This link will always point to this PR's latest changes: https://peek.run/vv2h2v5

packages/augur-ui/src/modules/market-cards/containers/migrate-market-notice.ts
```diff
@@ -16,48 +16,107 @@ import { DISMISSABLE_NOTICE_BUTTON_TYPES } from 'modules/reporting/common';
...
-      state.universe.forkingInfo &&
-      (state.universe.forkingInfo.winningChildUniverseId || hasReleaseRep)
-    ) &&
-    (reportingState !== REPORTING_STATE.FINALIZED &&
-      reportingState !== REPORTING_STATE.AWAITING_FINALIZATION);
 
+  if (
+    hasReleaseRep &&
+    releasableRep.claimableMarkets &&
+    releasableRep.claimableMarkets.unclaimedRep
```
bconfortin, line 65: Shouldn't this be testing `releasableRep.claimableMarkets.marketContracts` instead of `releasableRep.claimableMarkets.unclaimedRep`?

bthaile: only need to know if the user has unclaimedRep. The getter does have contracts in collection that the user has already claimed so the unclaimedRep would be "0". we don't want those.

bconfortin (approved):

bthaile:
Lets stir away from the mocks on this banner and put the banner in app.tsx only

PR #4083, opened 2019-10-07
Improvised solution

bconfortin:
This is a work in progress

ForkingBanner's wrapper was modified to <header> in order to avoid conflicts with `.app-styles_Main__content > section` and `<div>` styling in some pages.

A fix using `!important` was considered but discarded.


packages/augur-ui/src/modules/markets-list/components/markets-view.tsx
```diff
@@ -230,117 +230,119 @@ export default class MarketsView extends Component<
...
-          selectedCategory={selectedCategories}
-          search={search}
-          updateMobileMenuState={updateMobileMenuState}
-        />
-
-        <div className={classNames({
-          [Styles.Disabled]: isSearching,
-        })}>
-          <MarketTypeFilter
+      <>
```
bconfortin: Will remove this fragment.

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 💫

 This link will always point to this PR's latest changes: https://peek.run/exoh0qj

bthaile:
Chatted with @phoebemirman about the banner. She suggested putting the banner in App.tsx. It does make logical sense but I think there is an issue with the responsiveness stuff pushing the banner under the header and navigation. Is that what you were seeing?

bconfortin:
> Chatted with @phoebemirman about the banner. She suggested putting the banner in App.tsx. It does make logical sense but I think there is an issue with the responsiveness stuff pushing the banner under the header and navigation. Is that what you were seeing?

@bthaile I'm testing it right now. First attempt: it breaks the scrolls.

PR #4088, opened 2019-10-07
starting template strings for market creation templates

bthaile:
here is the first cut at template structure. 

structure describes user created outcomes for categorical markets
outcomes names get substituted by replace by other text input
added resolution rules the user can add.


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 😁

 This link will always point to this PR's latest changes: https://peek.run/0elhzvm

phoebemirman (approved):

packages/augur-ui/src/modules/create-market/get-template.ts
```diff
@@ -0,0 +1,158 @@
...
+                tooltip: `A team not listed as an outcome wins the event`,
+              },
+            ],
+            resolutionRules: [
+              ` If the game is NOT played or is not deemed an official game, meaning, less than 90% of the scheduled match had been completed, or ends in a tie, the market should resolve as "Draw/No Winner".`,
+            ],
+          },
+          {
+            marketType: CATEGORICAL,
+            question: `[0] vs [1]: Total goals scored; Over/Under [2].5`,
```
phoebemirman, line 130: Can you use a key instead of index

phoebemirman (changes requested):

packages/augur-ui/src/modules/create-market/get-template.ts
bthaile: added an id to the inputs. like a key but called `id`

phoebemirman (approved):

PR #4089, opened 2019-10-08
margin change and remove background color for bottom bar

bthaile:
small tweaks to market card style. seems we are caught between the mocks being updated and market cards spacing looking a bit different on desktop and mobile. 

desktop, outcomes go to the edges
image

mobile, outcomes have a gap to the edge to line up with bottom bar stats.
image


andrewdamelio (approved):
looks good

PR #4100, opened 2019-10-08
style-adjustments

JohnDanz:
corrected some smaller style issues i noticed while checking notifications

reporting market card rows now always have the button on the far right of the card instead of sometimes rendering in a non consistent way.

corrected transactions content box in account summary section that would cause the buttons to move around on wider screens.

moved the top icons over slightly on the market cards because they were not aligned with the text below. also corrected an extra bit of space on the `open reporting`, etc tag on market cards


bthaile (approved):
lgtm

PR #4111, opened 2019-10-09
only show template category cards where there is an actual template

bthaile:
filter down the template cards and only show ones where there is actually a template for.

image

image

image

### DIFF
--- a/packages/augur-ui/src/modules/create-market/components/form.tsx
+++ b/packages/augur-ui/src/modules/create-market/components/form.tsx
@@ -624,6 +624,7 @@ export default class Form extends React.Component<FormProps, FormState> {
                 <MarketType
                   updateNewMarket={updateNewMarket}
                   marketType={marketType}
+                  categories={newMarket.categories}
                 />
               )}
               {saveDraftError && (
--- a/packages/augur-ui/src/modules/create-market/components/form.tsx
+++ b/packages/augur-ui/src/modules/create-market/components/form.tsx
@@ -625,6 +625,7 @@ export default class Form extends React.Component<FormProps, FormState> {
                   updateNewMarket={updateNewMarket}
                   marketType={marketType}
                   categories={newMarket.categories}
+                  nextPage={this.nextPage}
                 />
               )}
               {saveDraftError && (
--- a/packages/augur-ui/src/modules/create-market/components/form.tsx
+++ b/packages/augur-ui/src/modules/create-market/components/form.tsx
@@ -619,7 +619,7 @@ export default class Form extends React.Component<FormProps, FormState> {
               )}
               {mainContent === REVIEW && <Review />}
               {mainContent === TEMPLATE_PICKER && <TemplatePicker />}
-              {mainContent === SUB_CATEGORIES && <SubCategories />}
+              {mainContent === SUB_CATEGORIES && <SubCategories nextPage={this.nextPage}/>}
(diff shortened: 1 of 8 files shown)


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🌈

 This link will always point to this PR's latest changes: https://peek.run/r9zh33m

packages/augur-ui/src/modules/create-market/components/form.tsx
```diff
@@ -619,11 +619,12 @@ export default class Form extends React.Component<FormProps, FormState> {
               )}
               {mainContent === REVIEW && <Review />}
               {mainContent === TEMPLATE_PICKER && <TemplatePicker />}
-              {mainContent === SUB_CATEGORIES && <SubCategories />}
+              {mainContent === SUB_CATEGORIES && <SubCategories nextPage={this.nextPage}/>}
```
phoebemirman, line 622: its weird that sub category page goes to next page automatically but the market type page or template picker one doesn't

packages/augur-ui/src/modules/create-market/get-template.ts
```diff
@@ -1,4 +1,19 @@
-import { CATEGORICAL } from 'modules/common/constants';
+import { CATEGORICAL, YES_NO } from 'modules/common/constants';
+import {
+  SOCCER,
+  AMERICAN_FOOTBALL,
+  SPORTS,
+  MARKET_TEMPLATES,
+  MARKET_SUB_TEMPLATES,
+  MARKET_TYPE_TEMPLATES,
+} from './constants';
```
phoebemirman: can you use the full path. same in other spot

phoebemirman (changes requested):

packages/augur-ui/src/modules/create-market/get-template.ts
```diff
@@ -71,45 +87,92 @@ export interface TemplateInput {
...
   categoryTemplates: Template[],
   marketType
 ) => {
   const values = categoryTemplates.filter(t => t.marketType === marketType);
-  return values;
+  return values.map(v => ({ ...v, inputs: getInputsByType(v.inputsType) }));
 };
 
+const getInputsByType = (inputName: TemplateInputTypeNames) => {
+  return inputes[inputName];
```
phoebemirman: this is spelled wrong

packages/augur-ui/src/modules/create-market/get-template.ts
```diff
@@ -123,87 +186,172 @@ export const getTemplateReadableDescription = (template: Template) => {
...
+            ],
+          },
+        ],
+      },
     },
   },
 };
+
+const inputes = {
+  [TemplateInputTypeNames.TEAM_VS_TEAM_POINTS_BIN]: [
```
phoebemirman, line 264: why are these types if they are only used once

bthaile: I suspect they will be used multiple times.

packages/augur-ui/src/modules/create-market/components/form.tsx
bthaile: I'd like the market type page to auto page as well. Template picker does auto-page

phoebemirman (approved):

Issue #4043, opened 2019-10-04
Disputing: Event emitter issue (in slow disputing)


bthaile:
Here is the scenario, for disputed market doesn't update reportingState:

1. market is in CrowdsourcingDispute
2. dispute to fill the dispute bond in slow disputing
3. UI requests the market and gets CrowdsourcingDispute as reportingState
4. Switch tabs in disputing, just to make sure and a new getMarkets request is made for markets in CrowdsourcingDispute and the disputed market is returned
5. Switching tabs to request markets in AwaitingNextWindow and no markets come back.
6. refresh UI
7. disputed market has the correct reportingState.

robert-e-davidson3:
It's a race condition.

I added a bunch of logging. When the race condition manifests itself, here's the order of execution:

1. Derived DisputeDB begins upserting its docs into the database.
2. The processing of the relevant log, DisputeCrowdsourcerCompleted, is executed in its entirety and returns a log to be bulk-upserted into the derived MarketDB. The method is called `processDisputeCrowdsourcerCompleted`.
3. Derived MarketDB begins upserting its docs into the database.
4. The processing of the timestamp from the new block* is executed up until it begins upserting docs. The execution at this point includes a database query of the Market logs. The method is called `processTimestamp`.
5. The bulk upserts for DisputeDB and MarketDB finish processing.
6. The timestamp upsert finishes processing.

Explanation:

The `processTimestamp` method depends on changes made in `processDisputeCrowdsourcerCompleted`, specifically the update to `nextWindowStartTime`. That update isn't made until `processTimestamp` has begun executing, specifically after it makes a database query.

When the race condition manifests, `processTimestamp` finishes upserting *after* `processDisputeCrowdsourcerCompleted` finishes upserting, thereby overwriting the correct `reportingState` that was derived from the DisputeCrowdsourcerCompleted log.

When the race condition does **not** manifest, this is the order of events:

1. Derived DisputeDB begins upserting its docs into the database.
2. The processing of the relevant log, DisputeCrowdsourcerCompleted, is executed in its entirety and returns a log to be bulk-upserted into the derived MarketDB.
3. Derived MarketDB begins upserting its docs into the database.
4. The processing of the timestamp from the new block begins execution. It doesn't yet get to the database query, let alone upserting.
5. The bulk upserts for DisputeDB and MarketDB finish processing.
6. The timestamp processing does the database query and fully upserts without interruption.

[...]

PR #4117, opened 2019-10-09
Disputing: Event emitter issue (in slow disputing) #4043

robert-e-davidson3:
See <https://github.com/AugurProject/augur/issues/4043> for details.


packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -270,6 +270,7 @@ export class MarketDB extends DerivedDB {
   }
 
   private processDisputeCrowdsourcerCompleted(log: ParsedLog): ParsedLog {
+    this.locks['processDisputeCrowdsourcerCompleted'] = true;
```
justinbarry, line 272: @robert-e-davidson3 Where is this lock being "cleared"? Does it only happen on DB creation?

packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -22,7 +21,8 @@ import { OrderBook } from '../../api/Liquidity';
 import { ParsedLog } from '@augurproject/types';
 import { MarketReportingState, SECONDS_IN_A_DAY } from '../../constants';
 import { QUINTILLION, padHex } from '../../utils';
-import { Block } from "ethereumjs-blockstream";
+import { Block } from 'ethereumjs-blockstream';
+import { sleep } from '../utils/utils';
```
bthaile: code smell having to sleep

packages/augur-sdk/src/state/db/MarketDB.ts
justinbarry: Disregard. The diff made it look like it was clearing in the wrong place. Looking at the full file It is fine.

packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -296,15 +297,20 @@ export class MarketDB extends DerivedDB {
...
   processTimestampSet = async (log: TimestampSetLog): Promise<void> => {
     const timestamp = new BigNumber(log.newTimestamp).toNumber();
     await this.processTimestamp(timestamp, log.blockNumber)
-  }
+  };
 
   private async processTimestamp(timestamp: number, blockNumber: number): Promise<void> {
+
+    while (this.locks['processDisputeCrowdsourcerCompleted']) {
+      await sleep(50);
```
bthaile: prob need ability to just break out and die. We don't want to sleep forever. I suggest putting a setTimeout or some way to break out after 1 second

robert-e-davidson3: Since I'm adding this logic to DerivedDB, I'll put the helper function there to do that.

packages/augur-sdk/src/state/db/MarketDB.ts
robert-e-davidson3: Very true. There's a discussion about this in discord #dev. We'll need to rearchitect a good deal of our state management code to fix this without sleep and locking.

bthaile (approved):
lgtm

PR #4125, opened 2019-10-10
gnosis docker

petong:
`yarn docker:gnosis` brings up the safe-relay service


robert-e-davidson3 (approved):
Local gnosis safe test worked. lgtm!

PR #4144, opened 2019-10-11
Est start time input

bthaile:
wire up using date time selector for estimate start time. It sets userInputs on template input object. 
Wasn't sure how to replace [#] in the market question

Default value in market question if user hasn't selected hour and minute (opened up time selector)
image

image

### DIFF
--- a/packages/augur-ui/src/modules/common/form.tsx
+++ b/packages/augur-ui/src/modules/common/form.tsx
@@ -1337,6 +1337,7 @@ interface TimeSelectorProps {
   onDateChange: Function;
   focused?: boolean;
   errorMessage?: string;
+  uniqueKey?: string;
 }
 
 export class TimeSelector extends React.Component<TimeSelectorProps, {}> {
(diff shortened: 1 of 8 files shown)


phoebemirman (changes requested):
you need to update userInput on the DATETIME input for it to show in the question. not seeing that happen in state and if it were the question should be updating

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🎊

 This link will always point to this PR's latest changes: https://peek.run/1reh0w4

bthaile:
fixed updating market question, fixed offset save issue.

phoebemirman (changes requested):
put this as offset:
Screen Shot 2019-10-10 at 9 24 55 PM

confirmed in review:
Screen Shot 2019-10-10 at 9 25 18 PM

went back to first page
Screen Shot 2019-10-10 at 9 25 38 PM

offset i selected is gone and time is different now

phoebemirman (approved):
lg, just need to fix that thing for the offset

PR #4145, opened 2019-10-11
pass in timezone if its available, pass it into timezone if its avail…

bthaile:
…able on parent component.

found bug with timezone not getting passed the value on newMarkets


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🤩

 This link will always point to this PR's latest changes: https://peek.run/o9ph75x

phoebemirman (changes requested):
offset isn't being applied
Screen Shot 2019-10-10 at 10 20 47 PM

bthaile:
removed the `useEffect` that was resetting timezone offset and offsetName. works now.

phoebemirman (changes requested):
Screen Shot 2019-10-10 at 11 19 17 PM
getting this crash on review step

maybe cause end time is null? shouldn't happen when we add validations

bthaile:
agreed, we'll need to add validations

phoebemirman (approved):

PR #4157, opened 2019-10-11
support for Market creation categories round 2/3

robert-e-davidson3:
Does NOT include the addition of a derived db. This PR is for interfacing with the UI but performance improvement is needed.


robert-e-davidson3:
Going to need to refactor to make use of `MarketListMetaCategories` in the `getCategoryStats` getter but I rewrote it a bit in expectation that I'd use it. I think it's more readable now.

packages/augur-sdk/src/state/getter/Markets.ts
```diff
@@ -749,47 +765,96 @@ export class Markets {
...
+          if (secondaryCategory) {
+            let secondaryStats = stats.categories[secondaryCategory];
+            if (!secondaryStats) {
+              stats.categories[secondaryCategory] = makeDefaultCategoryStats(secondaryCategory);
+              secondaryStats = stats.categories[secondaryCategory];
+            }
+            secondaryStats.numberOfMarkets++;
+            secondaryStats.volume = new BigNumber(secondaryStats.volume).plus(market.volume).toString();
+            secondaryStats.openInterest = new BigNumber(secondaryStats.openInterest).plus(market.marketOI).toString();
+          }
```
bthaile, line 832: ~~am I missing the place where primary's secondary stats (vol and oi) get added to the primary (vol and oi)? The vol, oi and num markets needs to roll up to the primary category stat~~

i see it gets added before the secondaryCategory condition

bthaile (approved):
lgtm

Issue #4181, opened 2019-10-14
Current Wallet SDK issues


andrewdamelio:
```
AS OF~~~
"fortmatic": "1.0.0",
"@portis/web3": "2.0.0-beta.53",
"@toruslabs/torus-embed": "0.2.6",
```

- Chrome ✅
- Opera: ✅
- Safari (13)  ✅
- Firefox ✅
- Brave ⚠️*

*Portis/Torus require the user to accept cookies on Brave
**Torus has pop-ups issues, pop-ups may need approval

### Tor.us
- OAuth pop-ups getting blocked
- Permissions menu doesn't save user state, asks for permissions every time
- Session cookies don't persist in all cases (tab closed)
- ~~On Brave, getting the below errors in regards to localStorage~~
- ~~On Brave, getting the below errors in regards to sessionStorage~~
- On Signup flow, we shouldn't load the previous session 
- ~~On Safari (Mobile/Desktop) had to disable the permissions modal since its getting blocked on Safari.~~

- iOS 12 Error 

### Portis

- <s>Permission window doesn't block/wait for users' selection </s>

- <s>On Portis login, the wallet remains open until the user closes it</s>

- On Portis load we get ```https://api.portis.io/auth/is-dapp-authorized 401```

- <s>On Brave, getting the below errors in regards to localStorage</s>

- On Signup flow, we shouldn't load the previous session

- ~~On desktop Safari getting the following errors on login~~

- iOS 12 Error

joeykrug:
Status of this?

andrewdamelio:
In progress. Wallets have been updated. Next step is to re-test above scenario and bring up issues.

joeykrug:
bump

andrewdamelio:
Closing this issue. #6389, #6700 are still open and relate.

PR #4183, opened 2019-10-14
use portis.openWallet

andrewdamelio:
relates to #4181

Fixes an issue where Portis wouldn't keep the permissions modal up long enough for the user to approve/decline.

Screen Shot 2019-10-14 at 9 23 04 AM

To Test:
1) Signup with a new Portis account
2) You should see a modal about approving/declining email permissions
3) You should be able to select an option, the window **should not** disappear after a few seconds.


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🌻

 This link will always point to this PR's latest changes: https://peek.run/exoh0mo

bthaile (approved):
nice!! lgtm

PR #4194, opened 2019-10-14
locking MarketDB on all log processing

robert-e-davidson3:


packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -229,6 +230,7 @@ export class MarketDB extends DerivedDB {
   }
 
   protected processDoc(log: ParsedLog): ParsedLog {
+    this.lock(this.PROCESS_DOC_LOCK);
```
nuevoalex: I think this should actually go in DerivedDB.handleMergeEvent and should be unlocked at the end of it.

robert-e-davidson3: ah so do it for all of them. will do

nuevoalex: The important thing is that that will make the lock persist during all of the event handling and only be released once its over. If its in process docs then it will be toggled for each doc in a large set

robert-e-davidson3: It was only being toggled on, never off, so it would have persisted. No reason to toggle in on a bunch of times though when once will do.

packages/augur-sdk/src/state/db/DerivedDB.ts
```diff
@@ -195,6 +197,10 @@ export class DerivedDB extends AbstractDB {
     return log;
   }
 
+  protected lock(name: string) {
+    this.locks[name] = true;
+  }
+
   protected async waitOnLock(lock: string, maxTimeMS: number, periodMS: number): Promise<boolean> {
```
nuevoalex: Just reading this, should we actually let it write if the thing its waiting for exceeds some timeout? I feel like this should actually crash or something before taking an action that would corrupt the DB

robert-e-davidson3: Great point. Setting it to throw an error instead of returning a bool.

nuevoalex (approved):
LGTM. Maybe up the timeout a bit though? Could potentially see a huge sync taking longer than a second

packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -305,7 +304,7 @@ export class MarketDB extends DerivedDB {
   };
 
   private async processTimestamp(timestamp: number, blockNumber: number): Promise<void> {
-    await this.waitOnLock('processDisputeCrowdsourcerCompleted', 1000, 50);
+    await this.waitOnLock(this.HANDLE_MERGE_EVENT_LOCK, 1000, 50);
```
robert-e-davidson3:
```suggestion
    await this.waitOnLock(this.HANDLE_MERGE_EVENT_LOCK, 2000, 50);
```

PR #4198, opened 2019-10-14
fix disputing markets counts, added check for calling update child un…

bthaile:
…iverse data when forking only when rep is minted.

updated disputing markets counts correctly when market dispute bond is filled and when disputing window is created. 

also fixes issue when nav off disputing page and getting react error that component was unmounted but then updated state.


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🍦

 This link will always point to this PR's latest changes: https://peek.run/4kghv7p

bconfortin (approved):
LGTM

PR #4204, opened 2019-10-14
Forking issues & nits

nuevoalex:
- Adds "disavowed" boolean to marketInfo
 - Normalize categories (and while in there reduce the burden of extraInfo processing in general by doing it once on processing)
 - Add proper validation / restriction to some PL contract methods


bthaile (approved):
lgtm, def more clean and readable.

PR #4212, opened 2019-10-15
Update market review seen

andrewdamelio:
Chatted with chwy - 
>I'm thinking we show it [Market Review Details modal]  only the very first time a user enters their first market and that's it.

Testing:
- If the ```marketReviewSeen``` localStorage var is set you will never see the modal.
- If it isn't and...
  - You load the markets page directly you should see it AFTER the market loads
  - You jump to a market from the markets list page you should see it upon loading the market


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 😍

 This link will always point to this PR's latest changes: https://peek.run/5m3heov

bthaile (approved):
lgtm

PR #4213, opened 2019-10-15
auto login wallets when possible

### LINKED ISSUE #3918: Onboarding - Signup/login/connect - auto login
Autologin: (show modal + login account)

- [x] Add a way to close modal if taking too long / issues
- [x] MetaMask
- [x] Portis
- [x] Fortmatic
- [x] Tor.us

andrewdamelio:
closes #3918
- login previous account on return
- better error handling


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🤙

 This link will always point to this PR's latest changes: https://peek.run/g7ohvq0

packages/augur-ui/src/modules/app/actions/init-augur.ts
```diff
@@ -38,56 +47,75 @@ function pollForAccount(
...
-        loggedInAccount = loginAccount.address;
-      }
-      if (!authStatus.isLogged && usingMetaMask && loggedInAccount && loggedInAccountType === ACCOUNT_TYPES.WEB3WALLET) {
-        autoLoginAccount(dispatch, loggedInAccount);
       }
     }
+  }
+
+  intervalId = setInterval(() => {
+    attempLogin();
```
bthaile: Is this spelt incorrectly on purpose? Missing t in attempt

andrewdamelio: my mistake

packages/augur-ui/src/modules/app/actions/init-augur.ts
```diff
@@ -38,56 +47,75 @@ function pollForAccount(
...
-      let loggedInAccountType: string = undefined;
-      if (!loginAccount.address) {
-        if (windowApp.localStorage && windowApp.localStorage.getItem) {
-          loggedInAccount = windowApp.localStorage.getItem('loggedInAccount');
-          loggedInAccountType = windowApp.localStorage.getItem('loggedInAccountType');
+
+  let attemptedLogin = false;
+  let intervalId = null;
+
+  async function attempLogin() {
```
bthaile: Spelling of attempt

packages/augur-ui/src/modules/app/actions/init-augur.ts
```diff
@@ -38,56 +47,75 @@ function pollForAccount(
...
+    const { connection } = getState();
+    if (attemptedLogin) {
```
bthaile, line 58: Minor thing, should this only process if UI isConnected?
It would make the code a little more readable

andrewdamelio: good call didn't notice that

packages/augur-ui/src/modules/app/actions/init-augur.ts
```diff
@@ -121,6 +149,47 @@ export function connectAugur(
...
+          signer: null,
+          email: null,
+          profileImage: null,
+          openWallet: null,
+          accountType,
+          isWeb3: true,
+          preloaded: true,
+        },
+      };
+      dispatch(updateAuthStatus(IS_LOGGED, true))
```
bthaile: Can’t set logged in here. The sdk needs to be synced in order to be officially “logged in”.  See how logged in is set in log-handler when user synced event comes in

bthaile: Can just remove setting logged in true here. Should all still work. If the sdk is told about the account

andrewdamelio: Didn't seen any issues during testing, but I think you are right here.

Yeah, in this case, I'll have to create another authStatus I think, something like PRELOADED, to control showing things like Account details and avatar,  while we wait for syncing to complete

Does that make sense?

andrewdamelio: > Can just remove setting logged in true here. Should all still work. If the sdk is told about the account

If I do this, currently, the user won't appear logged in till the SDK is loaded + markets. Which is a long delay vs being able to show that right on page load.

bthaile (changes requested):
made comments, some nit pics and don't set is logged in on state.

bthaile (approved):
Lgtm

PR #4228, opened 2019-10-16
0x work

nuevoalex:
- Integrate mesh browser via DI and a mock in tests
 - Add cancel methods to exchange


bthaile (approved):
lgtm, I especially like `Panic` verbosity. I think that will come in handy

Issue #4231, opened 2019-10-16
Disputing Reporting: claim reporting modal doesn't update after claiming


bthaile:
After user claims individual markets / participation tokens the modal doesn't dynamically update.

![image](https://user-images.githubusercontent.com/3970376/66883277-fd925480-ef92-11e9-867e-a9680a21f7c4.png)

claim all works since it closes the modal.

PR #4232, opened 2019-10-16
Add func releasing rep modal

### LINKED ISSUE #3928: Forking: Release All REP
need to modify select-reporting-winnings-by-market selector to pass in `isForking` boolean.
if during a fork, don't filter out any markets or dispute windows. The contract call doesn't need to change.

if `isForking` change the header and the primary button text. Need to add property on state.universe.isForking this is set by log-handler `handleUniverseForkedLog` method (needs to be done)

There is a flash script to force local node into forking state. 

<https://www.figma.com/file/aAzKHh4cA6OT2t7WFv2BQ7fB/Reporting-and-Disputing?node-id=2157%3A283>

image

### LINKED ISSUE #4066: Forking: contract call failures
Need to investigate, seems there is conditions not met so user can't perform these actions:

### issue 1
forking has already started, user needs to release their REP so they can migrate:
 * Can't release REP staked on market. markets are non-forking market
```typescript
    contracts.redeemStake.redeemStake(
      reportingParticipantsContracts,
      disputeWindows
    );
```

Note: `disavowCrowdsourcers()` needs to be called on all market before user can release all their REP. It would be advantageous if MarketsInfo had a property to indicate if the market's crowdsourcers have been disavowed already. Then the UI would know if 2 transactions are needed or just one per market.

### issue 2
fork has resolved there is a winning universe, market that haven't expired can't be migrated:
 * Can't migrate markets where their endTime hasn't passed.
```typescript
    market.migrateThroughOneFork(payoutNumerators, description);
```

### ~~issue 3~~ 
(has been addressed)
not forking related, but on the way to a fork, more info on ticket
 * Once markets move to slow disputing their reportingState doesn't update
<https://github.com/AugurProject/augur/issues/4043> existing ticket

### LINKED ISSUE #4199: Forking: Special logic needed for redeeming REP for forking market
`getAccountRepStakeSummary` results says REP staked on forking market isClaimable. 

this is correct. it doesn't "release" the REP so the user can migrate it. `forkAndRedeem` migrates the REP to the correct child universe based on how the user reported/disputed. 

need to have a special section in release modal to figure out when to call `forkAndRedeem` and add verbiage to tell the user what is going on

bthaile:
fixes https://github.com/AugurProject/augur/issues/3928
fixes https://github.com/AugurProject/augur/issues/4199
fixes https://github.com/AugurProject/augur/issues/4066

and verifies the above backend ticket. Tested regular non-fork claim REP and it works. Found one issue that might have already existed: <https://github.com/AugurProject/augur/issues/4231>

added special logic to allow user to migrate their staked REP on forking market
image

added special logic to allow the user to disavow market before releasing REP. 
image

the veribage will prob need to change but the functionality is there.


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 💜

 This link will always point to this PR's latest changes: https://peek.run/0elhze6

packages/augur-ui/src/modules/modal/containers/modal-claim-fees.ts
```diff
@@ -33,14 +34,18 @@ const mapStateToProps = (state: AppState, ownProps) => ({
...
 
 const mapDispatchToProps = (dispatch: ThunkDispatch<void, any, Action>) => ({
   closeModal: () => dispatch(closeModal()),
   redeemStake: (options, callback) => redeemStake(options, callback),
+  disavowMarket: marketId => disavowMarket(marketId),
 });
 
 const mergeProps = (sP: any, dP: any, oP: any) => {
+  const isForking = !!sP.forkingInfo;
+  const forkingMarket = isForking ? sP.forkingInfo.forkingMarket : undefined;
```
bconfortin: Does this have to be undefined in case it's not forking?

Alternative method (which would make forkingMarket = false instead of undefined) would be:
`const forkingMarket = isForking && sP.forkingInfo.forkingMarket;`

Not sure which one you find more readable.

because of the conditional down below
const isForkingMarket = market.id === forkingMarket;

bthaile: forkingMarket is a string, I didn't want to use it as a boolean. isForking is a boolean

packages/augur-ui/src/modules/modal/containers/modal-claim-fees.ts
```diff
@@ -59,15 +64,49 @@ const mergeProps = (sP: any, dP: any, oP: any) => {
     const market = marketObj.marketObject;
     if (market) {
       const marketRep = formatAttoRep(marketObj.totalAmount);
+      const isForkingMarket = market.id === forkingMarket;
```
bconfortin, line 67: > forkingMarket is a string, I didn't want to use it as a boolean. isForking is a boolean

But if `market` doesn't have an `id`, it would be undefined.
`undefined === undefined` would be true here, as opposed to `undefined === false`.

bthaile: if market doesn't have an `id` we have bigger problems. much much bigger problems

bthaile: but I see your point. I can make it null.

bconfortin (approved):
LGTM

PR #4251, opened 2019-10-16
unsubmodulize orbit-web

robert-e-davidson3:
Having orbit-web as a submodule wrecked our git workflow. This PR makes it a regular package.

In a subsequent PR, orbit-web will be added to augur-ui's webpack.


bthaile (approved):
lgtm

PR #4261, opened 2019-10-17
Building Orbit chat within our app instead of making it separate

robert-e-davidson3:


packages/augur-ui/package.json
```diff
@@ -101,6 +102,19 @@
   },
   "devDependencies": {
     "@augurproject/tools": "^2.0.0-alpha.1",
+    "@babel/core": "^7.6.4",
```
bthaile, line 105: why are all these deps needed?

packages/augur-ui/config/webpack.common.config.js
```diff
@@ -103,12 +107,71 @@ module.exports = {
         ]
       },
       {
-        test: /\.css$/,
+        test: /\.s?css$/,
```
bthaile, line 110: why are all these rules needed

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🥳

 This link will always point to this PR's latest changes: https://peek.run/zr5h7pv

packages/augur-ui/package.json
robert-e-davidson3: They're needed by webpack to process orbit-web. I tried not including them here since they're already in orbit-web/package.json but I got errors.

packages/augur-ui/config/webpack.common.config.js
robert-e-davidson3: They're to allow webpack to process orbit-web's code.

packages/augur-ui/package.json
bthaile: we can't build orbits in it's own dir and have augur-ui reference it?

packages/augur-ui/config/webpack.common.config.js
bthaile: why isn't orbits already built and available as a package?

packages/augur-ui/package.json
robert-e-davidson3: Ah. That should work. Rely on the top-level `build` also building orbit-web then just copy everything over to its own directory in the output.

robert-e-davidson3: Actually, that would involve changes to root's package.json.  The same change I removed from augur-ui, where we specifically build orbit-web.

packages/augur-ui/config/webpack.common.config.js
robert-e-davidson3: We would have to build it separately. That won't happen from our usual build process so we'd either need to:
1) Add a step to our usual process. Before someone does UI work, they'd need to run `yarn workspace orbit-web build`.
2) Add that command to `augur-ui/package.json` under the `build` commands.

bthaile: I’m confused. Doesnt the top level build build everything then build augur-ui?

Could orbits-web just get built by the top level build process. 
How does augur-sdk make it in augur-ui’s node_modules?

Could orbits-web follow the same process?
Would we basically have to make a node module out of orbits-web like augur-sdk in order for augur-ui to use it?

robert-e-davidson3: > I’m confused. Doesnt the top level build build everything then build augur-ui?

The top-level `build` only builds typescript, it doesn't run webpack or call the `build` script of any of its packages. We have to run `build` for augur-ui ourselves, though usually we use the `dev` wrapper instead.

> Could orbits-web just get built by the top level build process.

It can but that means changing the package.json script for `build` from just building typescript to also calling orbit-web's `build` command. This is what I had augur-ui's `build` script doing in an earlier PR.

> How does augur-sdk make it in augur-ui’s node_modules?
> Could orbits-web follow the same process?
> Would we basically have to make a node module out of orbits-web like augur-sdk in order for augur-ui to use it?

This would involve importing orbit-web into our project. That would violate our isolation policy of keeping 3rd party software in iframes.

We'd also need to turn it into a typescript-compatible project. When I tried to get around using babel by using our typescript loader, I got errors even with `transpileOnly` set to `true`. I believe this mean that they use features provided by babel but not typescript. Their .babelrc has enough plugins that this wouldn't surprise me.

robert-e-davidson3: Oh! Another possibility is to build orbit-web completely separately but commit its output somewhere, probably orbit-web/dist. Our webpack file would still need to copy things over but that's probably the only change it would need.

PR #4270, opened 2019-10-17
fix issue with custom market creation blowing up. rename method

bthaile:
custom market creation blowing up


packages/augur-ui/src/modules/create-market/get-template.ts
```diff
@@ -370,8 +370,9 @@ export const buildResolutionDetails = (
   return details;
 };
 
-export const findIfSubCats = category => {
-  if (TEMPLATES[category].children) return false;
+export const hasNoTemplateCategoryChildren = category => {
```
phoebemirman, line 373: not working for templates right

bthaile:
closing this PR

PR #4285, opened 2019-10-17
Warp sync

nuevoalex:
Warp Sync contract.

Allows for the creation of recurring markets on a per universe basis which are used to determine the latest IPFS hash of Augur log data.


packages/augur-core/source/contracts/WarpSync.sol
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    function getCreationReward(IUniverse _universe) public view returns (uint256) {
+        return getRepReward(_universe.creationTime());
+    }
+
+    function getRepReward(uint256 _theoreticalTime) private view returns (uint256) {
+        uint256 _currentTime = augur.getTimestamp();
+        uint256 _timeSinceTheoreticalCreationInSeconds = _currentTime - _theoreticalTime;
+        return (_timeSinceTheoreticalCreationInSeconds ** 3) * 1000;
```
robert-e-davidson3, line 77: Are we awarding more REP the longer we go without this market? That's awesome.

robert-e-davidson3 (approved):
lgtm

packages/augur-core/source/contracts/WarpSync.sol
nuevoalex: hehe yup

bthaile (approved):
Lgtm, some bits I dont get. Will patch wholes of understanding over Time

packages/augur-core/source/contracts/WarpSync.sol
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    IAugur public augur;
+    mapping(address => address) public markets;
+    mapping(address => Data) public data;
+
+    uint256 private constant MARKET_LENGTH = 1 days;
+    uint256 private constant MAX_NUM_TICKS = 2 ** 256 - 2;
+    int256[] private PRICES = [0, 1 ether];
+    string private constant EXTRA_INFO = '{"description":"What will the next Augur Warp Sync hash be?","longDescription":"What will the Augur SDK warp sync hash be for the latest block time before or equal to the end date of this market?"}';
```
justinbarry, line 24: Two things:
1. Could we go this with a blocknumber vs a time for the hash?
2. Should it be the hash at end of the *last* market?

nuevoalex: 1. The problem with blockNumber is that it makes wording the market more difficult and potentially could in theory result in an invalid market if the corresponding block number hasn't been mined.
2. The stored hash is for the last resolved market which will give us the latest hash available.

justinbarry (approved):
LGTM

PR #4289, opened 2019-10-18
0x v3 adjustments and API fixes

nuevoalex:
v3 0x requires a protocol fee be paid by takers in ETH, so we support that now.

Also modified the browser mesh API interface to match what it will actually be and adjust logic in the API to prioritize usage of browser mesh or client api.


justinbarry (approved):
LGTM

robert-e-davidson3:
`contracts.json` is empty now but needs to exist

robert-e-davidson3 (changes requested):
`contracts.json` is empty but needs content

robert-e-davidson3 (approved):

PR #4298, opened 2019-10-18
committing orbit-web build and copying it into augur-ui

robert-e-davidson3:


robert-e-davidson3:
the build involves some absolute paths so this approach does not work

PR #4299, opened 2019-10-18
augur-ui builds orbit-web then uses its output

robert-e-davidson3:
**Using this change will require everyone doing UI work to do a `yarn clean && yarn` to pick up the orbit-web changes.**

I went with this approach because:
1. User workflows are minimally affected. People have to do `yarn clean && yarn` but this is true of any major change . Importantly, there's no need to learn new git-submodule commands ;)
2. The orbit-web code is safely cordoned in an iframe.
3. The augur-ui webpack is minimally affected. No rule changes needed, just copying over yet another directory.
4. The augur-ui package.json doesn't need any new dependencies.

The major downsides:
1. The augur-ui `build` commands are more complex now.
2. Building augur-ui takes longer now because it first has to build orbit-web.
3. Changes to orbit-web aren't going to get picked up quite like people might expect. The augur-ui `dev` command merely builds the orbit-web code: it does not watch for changes to orbit-web. It will continue to watch for changes to augur-ui however so the typical UI dev flow will stay the same.


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🍕

 This link will always point to this PR's latest changes: https://peek.run/654hlpe

yarn.lock
```diff
@@ -2811,7 +2811,7 @@
   dependencies:
     defer-to-connect "^1.0.1"
 
-"@toruslabs/torus-embed@^0.2.3":
+"@toruslabs/torus-embed@0.2.3":
```
bconfortin, line 2814: Can this affect Andy? He is working on Torus stuff, I assume this is from the wallet as well.

robert-e-davidson3: @andrewdamelio 

This happened automatically when I called `yarn` but I can revert it if it's a problem.

bconfortin (approved):
Works for me. I believe once the changes in layout happen, we'll no longer modify Orbit anyway, so no point in adding it to the watcher

yarn.lock
andrewdamelio: no need to revert @robert-e-davidson3  👍

PR #4308, opened 2019-10-18
Chat is starting to look like the layout

bconfortin:
This is still a work in progress.

Note on sending a message: I can't turn the input[text] into a textarea because we don't have a "send" button and enter just breaks a line inside a textarea element.

I believe it's possible to override this behavior, but I'm not sure we want to (checking with Matt).


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🌈

 This link will always point to this PR's latest changes: https://peek.run/o9ph7nj

bthaile (changes requested):
I think this will work. Make sure mobile looks good. Should

bthaile (approved):
Lgtm, mashed the wrong button approved

PR #4309, opened 2019-10-19
fix tennis event dropdown

bthaile:
fix tennis event dropdown, had copy and paste error was showing horse racing events


robert-e-davidson3 (approved):
lgtm!

PR #4312, opened 2019-10-19
Gas optimizations

aaaaaaaaaaaaa5678:
Various changes focused on the trading paths to improve gas efficiency. Gas reductions as tested by tests/test_trade_gas_costs.py are approximately (depending on number of outcomes):

- ORDER CREATION BEST: 13-19%
- ORDER CREATION MAX: 12-16%
- CANCEL ORDER: 16-25%
- FILL ORDER TAKE SHARES: 27-41%
- FILL ORDER BOTH ETH: 23-27%
- FILL ORDER MAKER REVERSE: 23-29%
- FILL ORDER BOTH REVERSE: 15-22%

This change removes the "volume" parameter from the MarketVolumeChanged event as this can be calculated client-side by summing the individual volumes given in the "outcomeVolumes" parameter emitted by the same event.


packages/augur-core/source/contracts/Augur.sol
```diff
@@ -85,7 +85,7 @@ contract Augur is IAugur {
     event DisputeWindowCreated(address indexed universe, address disputeWindow, uint256 startTime, uint256 endTime, uint256 id, bool initial);
     event InitialReporterTransferred(address indexed universe, address indexed market, address from, address to);
     event MarketTransferred(address indexed universe, address indexed market, address from, address to);
-    event MarketVolumeChanged(address indexed universe, address indexed market, uint256 volume, uint256[] outcomeVolumes);
```
nuevoalex: We want to keep the summed volume to make syncing faster

packages/augur-core/source/contracts/trading/ShareToken.sol
```diff
@@ -30,30 +31,39 @@ contract ShareToken is ITyped, Initializable, VariableSupplyToken, IShareToken {
...
+        cancelOrder = _cancelOrder;
+
         completeSets = _augur.lookup("CompleteSets");
         claimTradingProceeds = _augur.lookup("ClaimTradingProceeds");
         profitLoss = IProfitLoss(_augur.lookup("ProfitLoss"));
         erc1820Registry = IERC1820Registry(_erc1820RegistryAddress);
         initialize1820InterfaceImplementations();
     }
 
+    function setUniverse(IUniverse _universe) external {
```
nuevoalex, line 62: This won't work for a fork. Also this won't work for the model of 1155 coming up soon anyway so probably should just remove. (We could keep a mapping of market to universe for the 1155 model perhaps but not sure its worth it there)

packages/augur-core/source/contracts/trading/Trade.sol
```diff
@@ -132,20 +132,27 @@ contract Trade is Initializable, ReentrancyGuard {
...
+            bytes32 _nextOrderId = _orders.getWorseOrderId(_orderId);
+            _lastTradePrice = _orderPrice;
             _bestAmount = fillOrder.fillOrder(_tradeData.sender, _orderId, _bestAmount, _tradeData.tradeGroupId, _tradeData.affiliateAddress);
             _orderId = _nextOrderId;
-            _orderPrice = orders.getPrice(_orderId);
+            _orderPrice = _orders.getPrice(_orderId);
             _tradeData.loopLimit -= 1;
         }
+        
+        if (_lastTradePrice != 0) {
```
nuevoalex, line 152: is this ever not true?

nuevoalex (commented):
Generally looks good, just had one thing we actually need that was removed and a few questions! There will definitely be a lot of conflicts with the ShareToken -> 1155 PR I have coming up but I can take care of those and send over to you for review.

packages/augur-core/source/contracts/Augur.sol
aaaaaaaaaaaaa5678: If it's necessary then I can sum the numbers in the contract. However can you describe how it significantly slows down syncing? It should be quite quick to sum up 8 numbers. Relative to the latency of a network transfer or disk write it should be trivial, so I have trouble understanding how it would add substantially to sync time.

packages/augur-core/source/contracts/trading/ShareToken.sol
aaaaaaaaaaaaa5678: What about it won't work? This gets called during migrateThroughOneFork to update it to the new universe, so it should have the correct value at any given time

packages/augur-core/source/contracts/trading/Trade.sol
aaaaaaaaaaaaa5678: This can happen if the above loop doesn't complete a single iteration, for example if it failed the test for isMatch or if loopLimit was set to 0.

packages/augur-core/source/contracts/Augur.sol
nuevoalex: Its not the biggest impact, but we generally err on the side of reducing sync time over small gas costs. If during a sync there are 10000 active markets this means we need to do a looping sum for every market to derive each market's volume during syncing where every ms is additional time before the UI is ready.

packages/augur-core/source/contracts/trading/ShareToken.sol
nuevoalex: Ah I see. Can stay for this PR then though we'll need to rethink with 1155 as noted

nuevoalex (approved):

PR #4369, opened 2019-10-22
Adjusted font sizes

bconfortin:
Adjusted font sizing and removed the orbit login screen. Still very raw, but at least the branding was removed.


bthaile (approved):
Lgtm

PR #4371, opened 2019-10-22
deployment writes to local-*.json files instead of the permanent records

robert-e-davidson3:
I also did a bit of cleanup but did NOT remove any networks from addresses.json and other files.

For future deployments to non-local networks (mainnet, kovan, etc), the deployer will need to manually overwrite the json files we commit to git (ex: upload-block-numbers.json).


bthaile:
I assume you tested cranking up docker:geth and deploying contracts and verifying that addresses.json doesn't get updated?

robert-e-davidson3:
> I assume you tested cranking up docker:geth and deploying contracts and verifying that addresses.json doesn't get updated?

I had tested it with ganache. I went ahead and tested docker:geth as well to the same good result.

bthaile:
Excellent

bthaile (approved):
Lgtm

nuevoalex (approved):
cool

PR #4378, opened 2019-10-23
 Dev Product Analytics [3431]

gitstart-augur:
To do list (Draft):

- [x] Setup IPFS worker 

- [x] Setup analytics.js plugin and interface with IPFS worker 

- [x]  Integrate analytics.js into a spot on Augur UI

- [x] First version of Analytics plugin: tracking all page transitions.

- [x] Start & connect IPFS service worker inside plugin via proxy


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). 🌈

**https://peek.run/w9jhq07**

-------

robert-e-davidson3:
I originally posted this in discord:

-----

It's definitely in the skeletal phase. The overall approach seems solid: set everything up then listen for events. They don't make the event listeners _do_ anything yet of course. When they do make them do stuff, we'll need corresponding code in the main thread to interpret the responses.

I'd prefer it if there was one event handler per event instead of bunching them together. They're iterating over lists of events for the two handlers they use but it'd be fine to just do a lot of `on()` calls. If we want to abstract it then use an interface or class for each of orbit and trackchannel. Those classes can use arrays if that's what makes the most sense for them though I doubt it will.

They need to handle messages sent to the worker. Right now it either sets up the listeners, ignores the message, or returns and error. There's no way to send Orbit chat messages, for instance, because there's no way to talk to the worker after you set it up. They should probably have a mapping from each `messageData.method` to a handler function.

The setup message should have more to it. There're a lot of defaults being used here that could just as easily be parameters. Partly that's for good design but also a lot of these defaults involve `process.env`, which isn't going to be available in the browser unless we use a webpack plugin or otherwise set it ourselves. And I prefer to avoid globals like `process.env` and magic values like `"/dns4/ws-star.discovery.libp2p.io/tcp/443/wss/p2p-websocket-star"`.

Finally, I'm a little worried that they have code specifically to suppress a typescript error but it looks like they're copying what we did in `Sync.worker.ts`, so... :/

gitstart-augur:
- By default, IPFS nodes already setup libp2p on starting (auto detect and use libp2p-node or libp2p browserify object), we only need something like custom libp2p when we want to control the connections.
- On these changes, I  refactor code and put more logs for showing current peers (please see console logs) and set an interval to broadcast a testing log every 30s to channel "augur-kovan-track-channel" and these messages are received on all orbit chat window (http://<<deploy_url>>/chat/index.html#/channel/augur-kovan-track-channel)

image

packages/augur-ui/package.json
```diff
@@ -63,11 +63,14 @@
     "highcharts": "7.2.0",
     "history": "4.10.1",
     "immutable-delete": "1.1.0",
+    "ipfs": "^0.39.0",
```
robert-e-davidson3: pin versions by removing the `^`

bconfortin:
@gitstart-augur Are those errors supposed to be happening?

errors

packages/augur-ui/package.json
```diff
@@ -63,11 +63,14 @@
...
     "history": "4.10.1",
     "immutable-delete": "1.1.0",
+    "ipfs": "0.39.0",
+    "analytics": "0.3.0",
     "lodash": "4.17.15",
     "memoizee": "0.4.14",
     "mnid": "0.1.3",
     "moment": "2.24.0",
     "moment-timezone": "0.5.26",
+    "orbit_": "^0.2.0-rc1",
```
bconfortin: Missed the removal of this `^` to pin the version

packages/augur-ui/src/services/ipfs/IPFS.worker.ts
```diff
@@ -0,0 +1,185 @@
...
+            }
+
+            currentChannelName = `${channel}-network-${ethereumNetwork}`;
+
+            currentChannel = await orbit.join(currentChannelName);
+
+            // subscribeOrbitChannelEvents.bind(this)(
+            //   currentChannel,
+            //   currentChannelName
+            // );
```
bconfortin: If this is going to be used in the future, I suggest adding it later instead of commenting it

packages/augur-ui/src/services/ipfs/IPFS.worker.ts
```diff
@@ -0,0 +1,185 @@
...
+
+          const orbitOptions = {
+            directory: `augur-orbit-chat-orbitdb-${ethereumNetwork}`,
+            id: senderAccount,
+          };
+
+          ipfs = await IPFS.create(ipfsOptions);
+          orbit = await Orbit.create(ipfs, orbitOptions);
+
+          // subscribeOrbitEvents.bind(this)();
```
bconfortin: If this is going to be used in the future, I suggest adding it later instead of commenting it

bconfortin (changes requested):
There are a few things that I think are worth checking out before merging, like the `subscribeOrbitEvents` binding currently commented out. Was that on purpose or did you forgot to uncomment it?

gitstart-augur:
> @gitstart-augur Are those errors supposed to be happening?
> 
> errors

Taking a look.

packages/augur-ui/src/services/ipfs/IPFS.worker.ts
gitstart-augur: Removed

gitstart-augur:
> There are a few things that I think are worth checking out before merging, like the `subscribeOrbitEvents` binding currently commented out. Was that on purpose or did you forgot to uncomment it?

This was basically for logging purpose. Removed it.

bconfortin (approved):

PR #4400, opened 2019-10-24
Depth chart adjustments

JohnDanz:
adjusting some edge case bugs with hover labels, a regression with zooming causing the container to grow, fixed some issues with the dotted lines rendering in the background of the chart.


bthaile (approved):
lgtm

PR #4426, opened 2019-10-25
Fix last price log order

### LINKED ISSUE #4419: Outcome last price doesn't update
trade on market and noticed that the outcome traded last price doesn't update. Trade history is correct. Even after clearing application data the price didn't update

justinbarry:
fixes <https://github.com/AugurProject/augur/issues/4419>

needed to sort filled logs for market and outcome


bthaile (approved):
lgtm

PR #4431, opened 2019-10-25
More explicit grid column configuration

bconfortin:


bthaile (commented):
seems like this would effect desktop. when mobile is really the only issue with text input being cut off.

bthaile (approved):
lgtm

PR #4440, opened 2019-10-26
move logic from UI to SDK to get user data when user synced is completed

bthaile:
Moved user aggregated login data to SDK. 

fixed `undefined` showing on portfolio page
image

### DIFF
--- a/packages/augur-sdk/src/state/getter/Users.ts
+++ b/packages/augur-sdk/src/state/getter/Users.ts
@@ -8,6 +8,7 @@ import {
   Doc,
   Timestamped,
   MarketData,
+  OrderState,
 } from '../logs/types';
 import {
   DisputeCrowdsourcerRedeemed,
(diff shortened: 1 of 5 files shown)


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). ⭐️

 This link will always point to this PR's latest changes: https://peek.run/m96h6pr

andrewdamelio (commented):
Look good!
noticed the below error when I'm signed in

image

bthaile:
~~I'm not able to repro the error above. If `getUserAccountData` didn't have account then there would be many more errors like that. I think it's coming from log-handlers. I'll address in a different PR when I can repro.~~

I see a use case where the error could be produced. If the user doesn't have any trades. fixed

andrewdamelio (approved):
fixed!

PR #4481, opened 2019-10-29
add user's created markets to login getter

bthaile:
I thought get staked rep getter returned user's created markets b/c of no-show REP bond, but doesn't seem to be the case. added explicit call to get user's created markets

### DIFF
--- a/packages/augur-sdk/src/state/getter/Users.ts
+++ b/packages/augur-sdk/src/state/getter/Users.ts
@@ -210,6 +210,11 @@ export class Users {
       orderState: OrderState.OPEN,
     });
 
+    const marketList = await Markets.getMarkets(augur, db, {
+      creator: params.account,
+      universe: params.universe,
+    });
+
     // user created markets are included, REP staked as no-show bond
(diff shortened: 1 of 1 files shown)


packages/augur-sdk/src/state/getter/Users.ts
```diff
@@ -269,7 +274,7 @@ export class Users {
       userStakedRep,
       userPositions,
       userPositionTotals,
-      marketsInfo
+      marketsInfo: [...marketList.markets, ...marketsInfo]
```
bconfortin, line 277: `marketsInfo: [...marketList.markets, ...marketsInfo]` I believe this would be better if you invert the values, ilke this: `marketsInfo: [...marketsInfo, ...marketList.markets]`, no?

Otherwise the values from marketList might be overwritten by marketsInfo

bthaile: overridden is fine. I didn't want to take the performance hit to pull out marketIds and re-get them.

bconfortin (approved):

PR #4486, opened 2019-10-29
adds script to merge addresses and block numbers after deploy

petong:
merges contract addresses and upload block numbers after deploy


robert-e-davidson3 (approved):
lgtm

Issue #3825, opened 2019-09-25
Trading: Gas price modal


pgebheim:
Redesign this modal to simplify, change copy too

design is inconsistent:
use `/ Trade`, we'll need to get the amount of gas needed to make a trade. for now. 

Calculate display value for `$AMOUNT / Trade`

gasPrice (user inputted or selected gas price in gwei)  
amount of gas per trade = `NEW_ORDER_GAS_ESTIMATE` (from constants.ts, we'll get a better value later) 

ETH_TRADE_COST = `formatGasCostToEther(NEW_ORDER_GAS_ESTIMATE, null, gasPrice)` from format-number to get the gasCost in ETH. 

We'll need to multiply that times the ETH/DAI exchange rate. I suggested a value of 0.00001 ETH to 1 DAI. something like that. 
EXCHANGE_RATE = 1000 (DAI / ETH)
we'll get the real value from the UNISWAP price feed (hasn't been created yet). 

AMOUNT = ETH_TRADE_COST * EXCHANGE_RATE

example: 
`fast` gasPrice might be 10 gwei 
Trade gas might be 700000

ETH_TRADE_COST = 7000000
AMOUNT = 7000000 * 0.00001 = $70.00
of course the value is way way too big, but we can tell that in the UI. 

example: 
`slow` gasPrice might be 1 gwei 
Trade gas might be 700000

ETH_TRADE_COST = 700000
AMOUNT = 700000 * 0.00001 = $7.00
of course the value is still too big, but we can tell that in the UI. 

updated after figured out I had the conversion from ETH to DAI backwards (fixed above)
```
formatEtherEstimate(
      formatGasCostToEther(
        NEW_ORDER_GAS_ESTIMATE,
        { decimalsRounded: 4 },
        createBigNumber(amount).times(GWEI_CONVERSION)
      )
    )
```

matt-bullock:
Desktop Prototype:
https://www.figma.com/proto/RMXc7jUWiC6amxz9Q2JK5L/Onboarding-and-Wallet-UX?node-id=1232%3A144049&viewport=16235%2C-12796%2C0.7076077461242676&scaling=min-zoom

Design File:
https://www.figma.com/file/RMXc7jUWiC6amxz9Q2JK5L/Onboarding-and-Wallet-UX?node-id=1232%3A144048

User edits transaction fee from their account dropdown
[...]

PR #4501, opened 2019-10-29
Custom gas price looking like layout for desktop

bconfortin:
https://github.com/augurproject/augur/issues/3825

### DIFF
--- a/packages/augur-ui/src/modules/modal/gas.tsx
+++ b/packages/augur-ui/src/modules/modal/gas.tsx
@@ -3,7 +3,6 @@ import React from 'react';
 import {
   Title,
   ButtonsRow,
-  SelectableTable,
   AlertMessage,
 } from 'modules/modal/common';
 
@@ -93,7 +92,7 @@ export class Gas extends React.Component<GasProps, GasState> {
             For more important transactions such as securing a sell order before anyone else takes it,
             we recommend a faster transaction fee.*
           </p>
-          <div className='gas-buttons'>
+          <div>
             {gasButtonsData.map(data => (
               <div onClick={data.action}
                    className={classNames({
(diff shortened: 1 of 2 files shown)


ghost (commented):

 This link will always point to this PR's latest changes: https://peek.run/m96h5gr

packages/augur-ui/src/modules/modal/gas.tsx
```diff
@@ -105,16 +104,30 @@ export class Gas extends React.Component<GasProps, GasState> {
...
+                onChange={(e) => {
+                  this.updateAmount(parseFloat(e.target.value));
+                }}
+              />
+            </div>
+            <div>
+              <div>
+                <span>&lt; $0.059</span><span> / Trade</span>
+              </div>
+              <span>0.012441 ETH</span>
```
bthaile, line 125: just hardcoding this value for now?

packages/augur-ui/src/modules/modal/gas.tsx
```diff
@@ -105,16 +104,30 @@ export class Gas extends React.Component<GasProps, GasState> {
...
+                type='number'
+                value={this.state.amount}
+                onChange={(e) => {
+                  this.updateAmount(parseFloat(e.target.value));
+                }}
+              />
+            </div>
+            <div>
+              <div>
+                <span>&lt; $0.059</span><span> / Trade</span>
```
bthaile, line 123: just hardcoding this value for now?

packages/augur-ui/src/modules/modal/modal.styles.less
```diff
@@ -1701,6 +1701,70 @@ div.ButtonsRow:last-of-type {
...
+          > span:first-of-type {
+            .mono-10-bold;
+
+            color: #E9E5F2;
+          }
+
+          > span:last-of-type {
+            .text-10-bold;
+
+            color: #848294;
```
bthaile: colors should not be hardcoded. should be using constants

bthaile (changes requested):
use color constants
don't hardcode values in the component, if only temporary values hardcode in container

packages/augur-ui/src/modules/modal/gas.tsx
bconfortin: Yeah. Just doing visuals

packages/augur-ui/src/modules/modal/modal.styles.less
bconfortin: Yeah, I was leaving those for the polish PR. As soon as I get the layout done I’ll check and change those, most likely on the next PR, if that’s ok.

bthaile: not going to allow hardcode colors to be merged up when we need to get the color constants worked out. There needs to be a really good reason to merge up hardcoded colors and I don't think this is a good reason.

bconfortin: Understood, won't happen again. Fixed on the latest commit (https://github.com/AugurProject/augur/pull/4501/commits/8ec12168e2ad43cd090f8ac47ba17dad136e3ca4).

bthaile: main reason is so much effort has gone into consolidating and coordinating the color pallet.

bthaile (approved):
sweet, lgtm

PR #4508, opened 2019-10-30
update gnosis safe and make signing standard

nuevoalex:
Update the Gnosis Safe and use eth_sign since that's supported in the newest version


packages/contract-dependencies-gnosis/src/ContractDependenciesGnosis.ts
```diff
@@ -161,12 +161,13 @@ export class ContractDependenciesGnosis extends ContractDependenciesEthers {
...
       nonce);
-    let sig = await this.signer.signDigest(ethers.utils.arrayify(txHashBytes));
+    let flatSig = await this.signer.signMessage(ethers.utils.arrayify(txHashBytes));
+    let sig = ethers.utils.splitSignature(flatSig);
 
     const signatures = [{
       s: new BigNumber(sig.s, 16).toFixed(),
       r: new BigNumber(sig.r, 16).toFixed(),
-      v: sig.v!
+      v: sig.v! + 4
```
robert-e-davidson3, line 170: This will be `NaN`, right? Is that intended?

packages/augur-core/source/contracts/GnosisSafe.sol
```diff
@@ -627,7 +698,32 @@ contract GnosisSafe is MasterCopy, BaseSafe, SignatureDecoder, SecuredTokenTrans
...
         external
-        returns (bool success)
+        returns (bool)
+    {
+        bytes32 txHash = checkTransaction(
+            to, value, data, operation, // Transaction info
+            safeTxGas, baseGas, gasPrice, gasToken, refundReceiver, // Payment info
+            signatures
+        );
+        //require(gasleft() >= safeTxGas, "Not enough gas to execute safe transaction");
```
robert-e-davidson3, line 708: Is this intended to be commented-out?

nuevoalex: Yea. There is a bug with ganache that will cause it to fail otherwise

packages/contract-dependencies-gnosis/src/ContractDependenciesGnosis.ts
nuevoalex: No it’ll be a number. The ! Is just to satisfy typescript

robert-e-davidson3 (approved):
lgtm!

PR #4509, opened 2019-10-30
fixed Users getter tests

robert-e-davidson3:
The `:getAccountTimeRangedStats` test really was timing out. It takes over 5 minutes to run on my macbook, in debug mode. Regular mode will be faster but CI and other dev machines might be slower so I set it to 400 seconds.

I also applied the ganache seed forking approach to the tests, isolating the tests' contract state from eachother, causing the P/L values to change.

There still might be issues with dangling promises or something else causing Jest to hang on a failure.


bthaile (approved):
lgtm

PR #4533, opened 2019-10-31
Added new gas layout for account dropdown

bconfortin:
https://github.com/augurproject/augur/issues/3825


bthaile (approved):
lgtm, we'll work out the proper numbers. I see what the design guys are going for.

PR #4534, opened 2019-10-31
allow user to loop through hours and minutes

bthaile:
fix time picker so user can loop through hour and minutes and AM/PM

### DIFF
--- a/packages/augur-ui/src/modules/common/form.tsx
+++ b/packages/augur-ui/src/modules/common/form.tsx
@@ -1555,7 +1555,9 @@ class IndividualTimeSelector extends React.Component<
   increment = () => {
     const value = this.state.value;
     if (!this.props.hasOptions) {
-      const newValue = parseFloat(value) + 1;
+      let newValue = parseFloat(value) + 1;
+      if (newValue > this.props.max) newValue = this.props.min;
+      if (newValue < this.props.min) newValue = this.props.max;
       this.onChange(newValue);
     } else {
       this.onChange(value === 'AM' ? 'PM' : 'AM');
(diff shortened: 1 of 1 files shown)


packages/augur-ui/src/modules/common/form.tsx
```diff
@@ -1555,7 +1555,9 @@ class IndividualTimeSelector extends React.Component<
   increment = () => {
     const value = this.state.value;
     if (!this.props.hasOptions) {
-      const newValue = parseFloat(value) + 1;
+      let newValue = parseFloat(value) + 1;
+      if (newValue > this.props.max) newValue = this.props.min;
+      if (newValue < this.props.min) newValue = this.props.max;
```
bconfortin, line 1560: Shouldn't this be like this? Because if min = 0 and max = 100, if you go from 100 to 101, you are being set back to 0. Shouldn't it just cap you at 100?

Same thing with min. If you are at 0 and try to to go to -1, you should return to 0 instead of going to 100, right?

```javascript
      if (newValue > this.props.max) newValue = this.props.max;
      if (newValue < this.props.min) newValue = this.props.min;
```

packages/augur-ui/src/modules/common/form.tsx
```diff
@@ -1565,7 +1567,9 @@ class IndividualTimeSelector extends React.Component<
   decrement = () => {
     const value = this.state.value;
     if (!this.props.hasOptions) {
-      const newValue = parseFloat(value) - 1;
+      let newValue = parseFloat(value) - 1;
+      if (newValue > this.props.max) newValue = this.props.min;
+      if (newValue < this.props.min) newValue = this.props.max;
```
bconfortin, line 1572: If the above comment makes sense, the values assigned here would also need to change.

packages/augur-ui/src/modules/common/form.tsx
bthaile: The issue was that it was capped. Its bad UX and a pain. Need to allow the user to loop through the values

packages/augur-ui/src/modules/common/form.tsx
bthaile: need to give the user ability to loop through the values and not get stuck.

packages/augur-ui/src/modules/common/form.tsx
bconfortin: Ohhhhh, now it makes sense. Didn't read the title. My bad.

bconfortin (approved):

PR #4544, opened 2019-11-01
Robert/zerox docker

robert-e-davidson3:


packages/augur-artifacts/src/index.ts
```diff
@@ -58,13 +58,23 @@ export interface ContractAddresses {
...
     GnosisSafe?: string;
     ProxyFactory?: string;
-    ZeroXTrade?: string;
-    ZeroXExchange?: string;
     BuyParticipationTokens?: string;
     RedeemStake?: string;
     CashFaucet?: string;
     GnosisSafeRegistry?: string;
     HotLoading?: string;
+    ZeroXTrade?: string; // TODO sup with this?
```
nuevoalex: Whats the comment here asking?

nuevoalex (approved):
LGTM. Confused on one comment

packages/augur-artifacts/src/index.ts
robert-e-davidson3: heheh I need to do a TODO sweep... At the time I wasn't sure if ZeroXTrade was still needed once we had the official 0x contracts.

robert-e-davidson3: That looks like only TODO.

PR #4546, opened 2019-11-01
Fixed for desktop. Mobile would require some work

bconfortin:
This fixes the chat for desktop, so the iframe is not recreated every time.

Since the mobile version is a modal, the modal altogether is recreated every time, so a different approach would have to take place in order for it to work, most likely using display toggle for none/block, position fixed and a big z-index.


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). ❤️

 This link will always point to this PR's latest changes: https://peek.run/1rehqow

bthaile (approved):
lgtm

PR #4605, opened 2019-11-06
uniswapv2 based price oracle

nuevoalex:


packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
```diff
@@ -0,0 +1,121 @@
...
+        IUniswapV2 exchange;
+        uint256 cashAmount;
+        uint256 repAmount;
+        uint256 blockNumber;
+        uint256 price;
+    }
+
+    mapping(address => ExchangeData) public exchangeData;
+
+    uint256 public gensisInitialRepPriceinAttoCash = 9 * 10**18;
```
robert-e-davidson3: s/gensis/genesis/

robert-e-davidson3:
```suggestion
    uint256 public genesisInitialRepPriceinAttoCash = 9 * 10**18;
```

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
```diff
@@ -0,0 +1,121 @@
...
+    function getRepPriceInAttoCash(IV2ReputationToken _reputationToken) external view returns (uint256) {
+        return calculateNewExchangeData(_reputationToken).price;
+    }
+
+    function calculateNewExchangeData(IV2ReputationToken _reputationToken) private view returns (ExchangeData memory) {
+        ExchangeData memory _exchangeData = exchangeData[address(_reputationToken)];
+        uint256 _blockNumber = block.number;
+        if (_blockNumber == _exchangeData.blockNumber) {
+            return _exchangeData;
+        }
```
robert-e-davidson3, line 54: This can only happen if someone runs this query twice in one block, right? Is this going to happen often enough to justify the gas cost of a conditional vs occasionally running through the whole method?

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
robert-e-davidson3: I'd commit this suggestion myself but it needs to be changed all over.

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
nuevoalex: That's a good point. I feel like it is worth having in to make the conditions more readable and since this won't be called very often the slight additional cost savings isn't worth removing it.

bthaile (approved):
lgtm

PR #4611, opened 2019-11-06
mistake using typeof to determin if array of object, needed to use Ar…

bthaile:
…ray.isArray


bconfortin (approved):

andrewdamelio (approved):
fixed my error locally

PR #4643, opened 2019-11-07
fix mobile errors

andrewdamelio:
Trading Tutorial doesn't support mobile layout. The tutorial would crash if you resized the window to a tablet or below while on the Trading Tutorial.

To fix this, we will redirect users to the markets page when they resize their tutorial window to small.

### DIFF
--- a/packages/augur-ui/src/modules/market/components/market-view/market-view.tsx
+++ b/packages/augur-ui/src/modules/market/components/market-view/market-view.tsx
@@ -47,6 +47,8 @@ import { formatShares, formatDai } from 'utils/format-number';
 import { convertUnixToFormattedDate } from 'utils/format-date';
 import { createBigNumber } from 'utils/create-big-number';
 import { TXEventName } from '@augurproject/sdk/src';
+import makePath from 'modules/routes/helpers/make-path';
+import { MARKETS } from 'modules/routes/constants/views';
 
 interface MarketViewProps {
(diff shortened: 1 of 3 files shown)


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🤙

 This link will always point to this PR's latest changes: https://peek.run/1rehqvv

packages/augur-ui/src/modules/market/components/market-view/market-view.tsx
```diff
@@ -597,7 +611,7 @@ export default class MarketView extends Component<
 
                       <TradingForm
                         market={market}
-                        initialLiquidity={preview && !initialLiquidity}
+                        initialLiquidity={preview && !tradingTutorial}
```
bthaile: what happen with `!initialLiquidity` boolean

phoebemirman: this should prob be initialLiquidity || (preview && !tradingTutorial)

phoebemirman: we don't need it here because the market creation doesn't use the market view

bthaile (approved):
lgtm, `initialLiquidity` isn't a good descriptive parameter for the function. can change down the line

PR #4666, opened 2019-11-08
Rep symbol and oi cash fees change

nuevoalex:


packages/augur-core/source/contracts/libraries/BytesToString.sol
```diff
@@ -0,0 +1,30 @@
+
+pragma solidity 0.5.10;
+
+
+/**
+ * @title BytesToString
+ * @dev Bytes operations to convert to string and remove whitespace
```
robert-e-davidson3: Removes nulls rather than whitespace.

robert-e-davidson3: I'm not sure why you're filtering out nulls but not other non-printable characters. Just to conform to solidity's (evm's?) string encoding? It'd make sense if it were null-terminated.

packages/augur-core/source/contracts/utility/RepSymbol.sol
```diff
@@ -0,0 +1,59 @@
...
+                }
+            } else if (_marketType == IMarket.MarketType.CATEGORICAL) {
+                uint256 _numOutcomes = _forkingMarket.getNumberOfOutcomes();
+                bytes32[] memory _outcomes = IAugurMarketDataGetter(_augurAddress).getMarketOutcomes(_forkingMarket);
+                for (uint256 _i = 1; _i < _numOutcomes; _i++) {
+                    if (_payoutNumerators[_i] != _numTicks) {
+                        _outcome = "MALFORMED";
+                    } else if (_payoutNumerators[_i] != 0) {
+                        _outcome = _outcomes[_i].bytes32ToString();
+                        _i = _numOutcomes;
```
robert-e-davidson3: Why this and not `break`?

packages/augur-core/source/contracts/utility/RepSymbol.sol
```diff
@@ -0,0 +1,59 @@
...
+                } else if (_payoutNumerators[1] != _numTicks) {
+                    _outcome = "MALFORMED";
+                }
+            } else if (_marketType == IMarket.MarketType.CATEGORICAL) {
+                uint256 _numOutcomes = _forkingMarket.getNumberOfOutcomes();
+                bytes32[] memory _outcomes = IAugurMarketDataGetter(_augurAddress).getMarketOutcomes(_forkingMarket);
+                for (uint256 _i = 1; _i < _numOutcomes; _i++) {
+                    if (_payoutNumerators[_i] != _numTicks) {
+                        _outcome = "MALFORMED";
+                    } else if (_payoutNumerators[_i] != 0) {
```
robert-e-davidson3: Is this conditional necessary, given that _numTicks is always positive anyway?

robert-e-davidson3 (approved):
LGTM

packages/augur-core/source/contracts/libraries/BytesToString.sol
nuevoalex: Because we store strings as bytes32s in this context the strings are padded to the right with a bunch of null characters

packages/augur-core/source/contracts/utility/RepSymbol.sol
nuevoalex: the payout for all non winning outcomes will be 0 so this is to skip them

packages/augur-core/source/contracts/utility/RepSymbol.sol
nuevoalex: No justifiable reason!

packages/augur-core/source/contracts/utility/RepSymbol.sol
robert-e-davidson3: They're all being chosen in the first case though since 0 != _numTicks.

packages/augur-core/source/contracts/libraries/BytesToString.sol
robert-e-davidson3: If they're always to the right then you could get the length by reading until you hit a null. That also simplifies the rest of this function.

nuevoalex: Someone could always add a null char to the actual string inbetween characters

packages/augur-core/source/contracts/utility/RepSymbol.sol
nuevoalex: Ah, right, actually this shouldn't work at all. Will refactor a bit

packages/augur-core/source/contracts/libraries/BytesToString.sol
robert-e-davidson3: We may need to restrict that. Someone could construct a market where two of the outcomes have an identical REP symbol by doing something like `so\0 happy` and `so\0 sad`.

nuevoalex: Those would end up being `sohappy` and `sosad` as is. For them to be identical theyd have to only differ by number of null chars

nuevoalex: Which is edge enough that I'm inclined to not care

packages/augur-core/source/contracts/utility/RepSymbol.sol
nuevoalex: In the process of refactoring found a different bug here actually! thanks for drawing attention to it

Issue #4173, opened 2019-10-13
React Refactor: convert to functional component `main-error-boundary.tsx`


bthaile:
convert UNSAFE_ methods and use useState and useEffect and useRef and convert to functional component

`augur/packages/augur-ui/src/modules/app/components/main-error-boundary.tsx`

```typescript
  UNSAFE_componentWillReceiveProps() {
    if (this.state.hasError) {
      this.setState({ hasError: false });
    }
  }

  componentDidCatch(error, info) {
    this.setState({ hasError: true });
  }

```

gitstart-augur:
Total Credit cost: 50

PR #4694, opened 2019-11-10
React Refactor: convert to functional component `main-error-boundary.tsx` 

gitstart-augur:
#4173


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). ❤️

 This link will always point to this PR's latest changes: https://peek.run/2n5h2wp

packages/augur-ui/src/modules/app/components/main-error-boundary.tsx
```diff
@@ -1,27 +1,31 @@
...
   };
 
-  UNSAFE_componentWillReceiveProps() {
-    if (this.state.hasError) {
-      this.setState({ hasError: false });
+
+  componentDidUpdate(prevProps: {}, prevState: MEBState) {
+    const { hasError } = this.state
+    if (prevState.hasError && hasError) {
+      setTimeout(() => {
```
bthaile: what is the advantage of using `setTimeout` here with 0?

pgebheim: @bthaile -- I'm assuming because setting state within componentDidUpdate is a no-no as it creates an update cycle. This looks like a bit of a hack to set the error state...

perhaps I'm wrong?

JohnDanz: @pgebheim you can setState in componentDidUpdate as long as it's in a conditional, like the if statement it's in.

pgebheim: Oh yes -- makes sense!

phunhakent: There is a rule for calling setState inside componentDidUpdate so I wrapped it with setTimeout (or setImmediate). Should I remove it?

gitstart-augur: There is a rule for calling setState inside componentDidUpdate so I wrapped it with setTimeout (or setImmediate). Should I remove it?

gitstart-augur: We can safely remove this:
https://reactjs.org/docs/react-component.html#componentdidupdate

bthaile (approved):
navigating around no issues.

bthaile (approved):
lgtm, fixed merge conflict

PR #4724, opened 2019-11-12
mousewheel-zoom-candles

JohnDanz:
updated zooming for candles to use mousewheel scrolling, updated Xaxis/crosshair labeling on price history


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://featurepeek.com). 🍦

 This link will always point to this PR's latest changes: https://peek.run/vv2h4lq

bthaile (approved):
lgtm, my magic pad doesn't have a scroll wheel and the chart doesn't rec my swiping. Chart appears to work other than my magic pad.

PR #4750, opened 2019-11-13
move template processing to log processing for market created

bthaile:
move processing to determine if market is derived from template.


robert-e-davidson3 (approved):
lgtm

PR #4805, opened 2019-11-15
add legacy rep faucet button to test v1 rep migration to v2

bthaile:
add legacy REP faucet button, to be used to test V1 REP to V2 migration.

image


phoebemirman (approved):

bthaile:
pulled down the PR and built it. it looks good.

PR #4837, opened 2019-11-19
Track gnosis safe nonce internally to avoid duplicate tx nonces

### LINKED ISSUE #4836: Gnosis - nonce issue
Getting nonce errors in the UI when using Gnosis safe.
{"exception":"SafeMultisigTxExists: Tx with nonce=15 for safe=0xD2849c8890C1d0418320CedED64672305fC5f6B7 already exists in DB"}

justinbarry:
closes #4836


andrewdamelio (changes requested):
This does help in sound regards but one big issue it does introduce is an inability to recover if a TX's fails -- since the nonces will be off by 1 until we refresh Augur.

andrewdamelio (approved):

PR #4842, opened 2019-11-20
Fix hard-coded universe value, make dynamic

mattxsmith:
The src/state/test-connector.ts example file in Augur SDK still uses an old universe value. The changes in this PR make the universe value align with whatever is in Addresses from the artifacts package. Also, I've made the Eth node endpoint be Alchemy instead of local.


bthaile (commented):
For a test connector using rinkeby is fine, kovan would be better. We use kovan as our primary test network.

mattxsmith:
Yeah, I used Rinkeby because that what it was in the original file. I've changed it to Kovan.

bthaile (approved):
lgtm

pgebheim:
Sorry, this needs to be reverted.

1) `augur-core` shouldn't be a runtime dep in the browser, and importing this here makes that not tree shake out. If we did want to use network configuration then that should be moved.

2) This is the test connector, which should be used in TESTS. That universe value is hard-coded on purpose and is not supposed to be the Kovan value.

mattxsmith:
@pgebheim That's fine, but the universe value should at least be the correct value, which it is not. Currently it gives this error when trying to get the markets:

`Error: Unknown universe: 0x02149d40d255fCeaC54A3ee3899807B0539bad60
`

(... which highlights the issue with hard-coding the value).

pgebheim:
When running a test? For tests we deploy into a Ganache container that has a snapshotted deploy, its not the Kovan/testnet deploy.

mattxsmith:
When following the instructions in the README, which is to run `yarn webpack-dev-server` and go to http://localhost:8080. Didn't realize the file was for actual tests, feel free to disregard.

pgebheim:
@mattxsmith -- Try this section of the main readme:

https://github.com/AugurProject/augur/#ui-development

PR #4947, opened 2019-11-28
update trading tutorial/torus update

andrewdamelio:
- torus will only use google as a login method
- trading tutorial - copy update
- trading tutorial - auto-scroll to the bottom of the page for STEP 9


bthaile (approved):
Lgtm

PR #4964, opened 2019-12-02
Removes unused member variable shareTokens.

MicahZoltu:
Someone should probably evaluate if this was _supposed_ to be used somewhere...


nuevoalex:
holdover from pre-1155 share tokens I believe

PR #5045, opened 2019-12-05
added flash script to validation template 

bthaile:
another flash utility to help validatie market creation extra info template value against template validation rules.


packages/augur-tools/src/flash/template-utils.ts
```diff
@@ -19,3 +20,19 @@ const findTemplate = (category: CategoryTemplate, hash: string): string => {
   template = category.templates.find(t => t.hash === hash);
   return template;
 };
+
+export const validateMarketTemplate = (title: string, templateInfo: string, outcomesString: string, longDescription: string): string => {
+  const extraInfoTemplate = JSON.parse(templateInfo) as ExtraInfoTemplate;
+  let outcomes = [];
+  if (outcomesString) {
+    outcomes = outcomesString.split(',').map(s => stringTo32ByteHex(s));
```
robert-e-davidson3:
```suggestion
    outcomes = outcomesString.split(',').map(stringTo32ByteHex);
```

robert-e-davidson3 (approved):
added one suggestion but it's just a nitpick

lgtm!

packages/augur-tools/src/flash/template-utils.ts
bthaile: I'm not a super fan of that syntax, too shortcutty, but I'll make the change

PR #5046, opened 2019-12-05
use MKRShutdownHandler for cash transfers

nuevoalex:


packages/augur-core/source/contracts/Augur.sol
```diff
@@ -188,9 +192,9 @@ contract Augur is IAugur, IAugurCreationDataGetter {
     // Transfer
     //
 
-    function trustedTransfer(IERC20 _token, address _from, address _to, uint256 _amount) public returns (bool) {
+    function trustedTransfer(address _from, address _to, uint256 _amount) public returns (bool) {
```
MicahZoltu: Consider naming this `trustedCashTransfer` to make it a bit more clear at the callsite what this is doing.

packages/augur-core/source/contracts/MKRShutdownHandler.sol
```diff
@@ -0,0 +1,51 @@
...
+import 'ROOT/libraries/math/SafeMathUint256.sol';
+
+
+contract MKRShutdownHandler {
+    using SafeMathUint256 for uint256;
+
+    ICash public cash;
+    IDaiVat public vat;
+
+    uint constant DAI_ONE = 10 ** 27;
```
MicahZoltu: VatDAI is `10**45`
DAI is `10**18`

Neither is `10**27`.

nuevoalex: The constant in their codebase is called `ONE`. As that doesn't really make sense in the context of Augur I wasn't sure on the best name. Perhaps `MKR_ONE`??

MicahZoltu: It doesn't make sense in the context of their codebase either.  😉

Ignoring gas costs, my recommendation would be to add a couple methods for `vatDaiToDai` and `daiToVatDai` and inside those just do `value.mul(10**27)` and `value.div(10**27)` directly (no constant).  The goal would be to isolate the Maker weirdness into well named functions and not try to figure out how to name that weird scalar multiplier.

packages/augur-core/source/contracts/MKRShutdownHandler.sol
```diff
@@ -0,0 +1,51 @@
...
+            if (_amount == 0) {
+                return;
+            }
+        }
+        require(cash.transferFrom(_from, _to, _amount));
+    }
+
+    function shutdownTransfer(address _from, address _to, uint256 _amount) internal returns (uint256) {
+        uint256 _cashBalance = cash.balanceOf(_from);
+        if (_cashBalance < _amount) {
```
MicahZoltu: This local variable doesn't appear to be adding any value vs inlining as it isn't used outside of the conditional check.

packages/augur-core/source/contracts/MKRShutdownHandler.sol
nuevoalex: We are already referencing and dealing with MKR code/contracts and making it a magic value seems wrong. that said the function idea does sound more readable so I'll rename the const and add those funcs

MicahZoltu: Perhaps call it `RAY` if you want to go with Maker naming scheme?

packages/augur-core/source/contracts/MKRShutdownHandler.sol
```diff
@@ -0,0 +1,58 @@
+pragma solidity 0.5.10;
+
+import 'ROOT/ICash.sol';
+import 'ROOT/external/IDaiVat.sol';
+import 'ROOT/libraries/math/SafeMathUint256.sol';
+
+
+contract MKRShutdownHandler {
```
MicahZoltu: Consider renaming this to something like `CashSender`.  Looking over the way this is used, it appears to basically be a base class for "things that need to be able to send cash".  The current naming is confusing because when I say to myself, "DisputeWindow is a MKRShutdownHandler" that doesn't really feel right, but "DisputeWindow is a CashSender" does.

MicahZoltu (commented):
I'm a little concerned about the maintainability of this solution.  If anyone does `cash.transfer` in the code, a generally reasonable line of code to write, things will break.  I would like to see some mechanism (I'm not sure what) that makes it so a future developer cannot shoot themselves in the foot by calling `cash.transfer` manually.  This is one of those things that is unlikely to get proper test coverage over time unless _every_ test is run both during and outside of emergency shutdown (which feels like a somewhat unreasonable test requirement).

nuevoalex:
I'll investigate adding some sort of magical check in our test Cash contract to make things fail if we do it the wrong way.

packages/augur-core/source/contracts/MKRShutdownHandler.sol
epheph: "ONE" is the new name in their contracts, but i agree it is terrible. The name and the value differ by `10^27-1`.

packages/augur-core/tests/solidity_test_helpers/MockReputationToken.sol
```diff
@@ -51,7 +51,7 @@ contract MockReputationToken is ITyped, Initializable, MockVariableSupplyToken,
         setMigrateInValue = _setMigrateInValue;
     }
 
-    function setTrustedTransfer(bool _setTrustedTransferValue) public {
+    function settrustedCashTransfer(bool _setTrustedTransferValue) public {
```
epheph: set[T]rustedCashTransfer ? Don't matter tho

packages/augur-core/source/contracts/CashSender.sol
```diff
@@ -0,0 +1,58 @@
...
+
+    function cashTransfer(address _to, uint256 _amount) internal {
+        address _from = address(this);
+        if (vat.live() == 0) {
+            _amount = shutdownTransfer(_from, _to, _amount);
+            if (_amount == 0) {
+                return;
+            }
+        }
+        require(cash.transfer(_to, _amount));
```
epheph, line 29: In other places, there is protection in place to make sure things function even if cash balances are less. I haven't fully worked it out, but the fact that Augur overpays on withdraw:

```
if (_sDaiAmount.mul(_chi) < _amount.mul(RAY)) {
            _sDaiAmount += 1;	            _sDaiAmount += 1;
        }	        
}
```

and during a shutdown, DSR = ONE (so not interest accruing), we could end up every-so-slightly under?

nuevoalex: I don't think in any of the places this is used there is an expectation or an allowance that the balance is less actually.

The referenced bit for DSR handling should only come into effect when interest has accrued and would generate a rounding error. We do not allow an overdraw in that code from the pot so if for some reason the rounding error case is triggered but there is insufficient money in the pot (not sure how it would happen) we'd just fail there.

packages/augur-core/source/contracts/trading/CreateOrder.sol
```diff
@@ -107,4 +112,79 @@ contract CreateOrder is Initializable, ReentrancyGuard {
...
+            _shortOutcomes[_i] = _indexOutcome;
+            _indexOutcome++;
+        }
+
+        // Figure out how many almost-complete-sets (just missing `outcome` share) the creator has
+        uint256 _attosharesHeld = _orderData.shareToken.lowestBalanceOfMarketOutcomes(_orderData.market, _shortOutcomes, _orderData.creator);
+
+        // Take shares into escrow if they have any almost-complete-sets
+        if (_attosharesHeld > 0) {
+            _orderData.sharesEscrowed = SafeMathUint256.min(_attosharesHeld, _attosharesToCover);
```
MicahZoltu, line 149: Style: Elsewhere in the codebase `_attosharesHeld.min(_attosharesToCover)` is used.  I personally prefer the style you have here over what is standard in the codebase, but I think consistency is valuable so I recommend picking one and sticking to it.

packages/augur-core/source/contracts/trading/CreateOrder.sol
```diff
@@ -107,4 +112,79 @@ contract CreateOrder is Initializable, ReentrancyGuard {
...
+            _attosharesToCover -= _orderData.sharesEscrowed;
+            uint256[] memory _values = new uint256[](_numberOfShortOutcomes);
+            for (uint256 _i = 0; _i < _numberOfShortOutcomes; _i++) {
+                _values[_i] = _orderData.sharesEscrowed;
+            }
+            _orderData.shareToken.unsafeBatchTransferFrom(_orderData.creator, address(_orderData.augurTrading), _orderData.shareToken.getTokenIds(_orderData.market, _shortOutcomes), _values);
+        }
+
+        // If not able to cover entire order with shares alone, then cover remaining with tokens
+        if (_attosharesToCover > 0) {
```
MicahZoltu, line 159: Reading this code again after being away for a while makes me wonder how much simpler trading would be if we didn't allow mixed orders (part shares, part tokens).  I further wonder how frequently users _actually_ want to do position flip + raise like this.  It feels like we could just assert that orders are _either_ shares _or_ tokens and probably not negatively impact anyone, and likely also simplify a lot of code at the same time.

MicahZoltu (approved):
My biggest concern out of all my comments is that it would be _really_ easy to call `cash.transfer` somewhere in the codebase rather than `cashTransfer`, and if that happens somewhere there is a high probability of stuck funds.  Some way of preventing this would be valuable, but I have been unable to come up with an elegant solution to protecting the codebase from such things.  In fact, even just identifying whether you found all of the `cash.transfer` calls would be a manual "find in files" solution!

nuevoalex:
I added a python tests that scans contracts for instances actually in my last push.

packages/augur-core/source/contracts/trading/CreateOrder.sol
nuevoalex: Don't think its worth simplifying and removing functionality at this point

packages/augur-core/source/contracts/CashSender.sol
epheph: Yes i think you are correct. the only time it can round down is if it has already accrued interest since sDaiAmount * chi will not necessarily be divisible by the deposited vDai amount

epheph:
It would be nice if cash could be a token like DAI-HRD or chai that functions like a standard token, handling shutdown at the token-level, without bothering every integrating system with complexities of vatDai shutdown behavior . Not saying it should change by any means (since now there's a new complex component to review and extra burden on the user), just musing aloud.

PR #5113, opened 2019-12-07
0x

robert-e-davidson3:
This PR provides 0x in the gnosis docker-compose. Also a ton of 0x integration work.


andrewdamelio:
@robert-e-davidson3 in the relayer logs I see the below
```
 Cannot seed wallet with funds. Please faucet 0xbd355A7e5a7ADb23b51F54027E624BfE0e238DF6
```

Is it possible to have the fauceting automatic so devs don't need to do anything to prime the safe maker?

Also, I did seed the above address but it didn't seem to help to get the gnosis safe created.

andrewdamelio:
@robert-e-davidson3 When I load the UI and I look at the SDK object zeroX is undefined.
```AugurSDK.zeroX is undefined```

Do you see the same if you boot up the ui?

File: packages/augur-sdk/src/state/getter/ZeroXOrdersGetters.ts [hunk 1]
```diff
@@ -62,11 +62,16 @@ export class ZeroXOrdersGetters {
...
     const orderType = params.orderType ? `0x0${params.orderType}` : undefined;
 
-    let currentOrdersResponse = await db.ZeroXOrders.where('[market+outcome+orderType]').equals([
-      params.marketId,
-      outcome,
-      orderType
-    ]).toArray();
+    let currentOrdersResponse;
+    if (typeof outcome === 'undefined' && typeof orderType === 'undefined') {
+      currentOrdersResponse = await db.ZeroXOrders.toArray();
```
nuevoalex: shouldn't this still be querying based on market? As is this would return all existing orders on Augur.

nuevoalex: Also shouldn't this be an OR instead of an AND? I think there probably isnt a case where only one will be undefined but technically if there were this would raise an error

nuevoalex (changes requested):
Just a small SDK change needed then LGTM!

packages/augur-sdk/src/state/getter/ZeroXOrdersGetters.ts
robert-e-davidson3: Ah yeah. I did this fast to get it working but it definitely needs a fixup. There's a TODO at the beginning of this getter that makes me think a full refactor is needed.

I'll implement these suggestions real quick.

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). 😍

**https://peek.run/0elh2jk**

-------

packages/augur-sdk/src/state/getter/ZeroXOrdersGetters.ts
robert-e-davidson3: Like this?
```
currentOrdersResponse = await db.ZeroXOrders.where('[market+outcome+orderType]').between(
        [params.marketId, Dexie.minKey, Dexie.minKey],
        [params.marketId, Dexie.maxKey, Dexie.maxKey],
      ).toArray();
```

robert-e-davidson3 (commented):
> @robert-e-davidson3 in the relayer logs I see the below
> 
> ```
>  Cannot seed wallet with funds. Please faucet 0xbd355A7e5a7ADb23b51F54027E624BfE0e238DF6
> ```
> 
> Is it possible to have the fauceting automatic so devs don't need to do anything to prime the safe maker?
> 
> Also, I did seed the above address but it didn't seem to help to get the gnosis safe created.

I'm pretty sure it's possible. @justinbarry is probably the best person for this

robert-e-davidson3:
> @robert-e-davidson3 When I load the UI and I look at the SDK object zeroX is undefined.
> `AugurSDK.zeroX is undefined`
> 
> Do you see the same if you boot up the ui?

Fixed!

nuevoalex (approved):
👍

justinbarry:
@andrewdamelio @robert-e-davidson3 Can we faucet on deploy someplace? It probably shouldn't happen on the relay side.

Issue #5107, opened 2019-12-06
Unnecessary msg.sender allows


epheph:
There are a few places that there is more privilege to a function than necessary:

does not need to allow `this`:
https://github.com/AugurProject/augur/blob/13b5cefdc88df60195f4795cf7b5ececc1fec4e1/packages/augur-core/source/contracts/trading/ProfitLoss.sol#L63
https://github.com/AugurProject/augur/blob/67c0fec95e65fca8d27751f28a6c06e5247978c6/packages/augur-core/source/contracts/trading/FillOrder.sol#L487
https://github.com/AugurProject/augur/blob/1ad689ec463c4d7bfc11addb6cd8b3d86080852e/packages/augur-core/source/contracts/trading/Orders.sol#L316
https://github.com/AugurProject/augur/blob/1ad689ec463c4d7bfc11addb6cd8b3d86080852e/packages/augur-core/source/contracts/trading/Orders.sol#L286
https://github.com/AugurProject/augur/blob/1ad689ec463c4d7bfc11addb6cd8b3d86080852e/packages/augur-core/source/contracts/trading/Orders.sol#L307

Issue #5097, opened 2019-12-06
Validate construction ordering.


MicahZoltu:
https://github.com/AugurProject/augur/blob/c175149d6991b14936e92f02ef5602f29e8007bf/packages/augur-core/source/contracts/reporting/ReputationToken.sol#L36-L37

Some constructors such as the one above have a dependency on other contracts being deployed and registered first.  Rather than relying on the deployment process to hopefully get deployment order correct, instead validate that the value returned by `augur.lookup` is not `0` with something like `require(warpSync != address(0))`.

I have not yet found any exploits revolving around dependencies being `0`, but the system will certainly break in various ways if this happens.

Issue #5075, opened 2019-12-05
Misnamed method `ReputationToken.assertReputationTokenIsLegitSibling`


MicahZoltu:
`ReputationToken.assertReputationTokenIsLegitSibling` actually asserts that the given reputation token is a child reputation token, not a sibling.  Sibling would be "my universe's parent is the same as its universe's parent".

Issue #5070, opened 2019-12-05
Misnamed function: InitialReporter.designatedReporterWasCorrect()


epheph:
https://github.com/AugurProject/augur/blob/7271dc4a8a06fdc7b6ae6f3ec91ee733b4cc7ae2/packages/augur-core/source/contracts/reporting/InitialReporter.sol#L125

The reporter responsible for the `payoutDistributionHash` might not have been the designatedReporter. Perhaps it should be called 'isCorrect' or `initialReportWasCorrect` ?

epheph:
Also this will return `true` while the market is not finalized (hasn't set a payoutDistributionHash) and the initialReport has not been reported (0x00000000 = 0x0000000).

this is probably the larger issue. want new issue or just add to this one?

PR #5114, opened 2019-12-07
Low audit issues

nuevoalex:
Includes:

#4957 
#4958 
#4967 
#5067 
#5097 
#5070 
#5075 
#5107


epheph (approved):
These changes look great. There are 3 more `msg.sender == address(this)` in Orders.sol (did i identify those correctly?) for #5107  .

nuevoalex:
@epheph yea those are actually required for the TestOrders.sol contract to function which we use to do automated tests for the logic in storing orders

packages/augur-core/source/contracts/reporting/ReputationToken.sol
```diff
@@ -35,6 +35,8 @@ contract ReputationToken is VariableSupplyToken, IV2ReputationToken {
         parentUniverse = _parentUniverse;
         warpSync = _augur.lookup("WarpSync");
         legacyRepToken = IERC20(augur.lookup("LegacyReputationToken"));
+        require(warpSync != address(0));
+        require(legacyRepToken != IERC20(0));
         erc1820Registry = IERC1820Registry(_erc1820RegistryAddress);
```
MicahZoltu, line 40: Need one for `erc1820Registry` as well, plus the same thing in several other contracts.

packages/augur-core/source/contracts/reporting/Universe.sol
```diff
@@ -93,11 +93,24 @@ contract Universe is IUniverse {
...
+        require(marketFactory != IMarketFactory(0));
+        require(disputeWindowFactory != IDisputeWindowFactory(0));
+        require(shareToken != IShareToken(0));
+        require(repPriceOracle != IRepPriceOracle(0));
+        require(formulas != IFormulas(0));
+        require(cash != ICash(0));
+        require(daiVat != IDaiVat(0));
+        require(daiPot != IDaiPot(0));
+        require(daiJoin != IDaiJoin(0));
+    }
```
MicahZoltu, line 112: Out of curiosity, why did you break this out into a separate function?  It feels like splitting it like this is going to increase the chances of missing/forgetting one vs something like:
```solidity
marketFactory = IMarketFactory(augur.lookup("MarketFactory"))
require(marketFactory != IMarketFactory(0);
```
Also, if you wanted to save a little gas by using a local variable:
```solidity
address _temp;
_temp = augur.lookup("MarketFactory");
require(_temp != address(0));
marketFactory = IMarketFactory(_temp);
_temp = augur.lookup("DisputeWindowFactory");
require(_temp != address(0));
disputeWindowFactory = IDisputeWindowFactory(_temp);
...
```

MicahZoltu (approved):

packages/augur-core/source/contracts/reporting/ReputationToken.sol
epheph: It enters erc1820Registry on the next line, there is an implicit check (technically, on extcodesize != 0).

MicahZoltu: Good point @epheph.  However, there is still the same issue in multiple other contracts.  Not sure if @nuevoalex would prefer a bunch of small PRs, or a list in GitHub, or to just hunt them down yourself?

packages/augur-core/source/contracts/reporting/Universe.sol
nuevoalex: Thought this was more readable.

packages/augur-core/source/contracts/reporting/ReputationToken.sol
nuevoalex: I'd prefer them being explicitly listed so they can be evaluated and we all see that.

Issue #4993, opened 2019-12-03
Free extra tokens when _longOutcome >= numOutcomes in  buyCompleteSetsForTrade


epheph:
https://github.com/AugurProject/augur/blob/67c0fec95e65fca8d27751f28a6c06e5247978c6/packages/augur-core/source/contracts/reporting/ShareToken.sol#L152

When _longOutcome exceeds numOutcomes, the shortRecipient receives a full CompleteSet, and the longRecipient receives an _amount of tokens for an invalid outcomeId .

However, because _longOutcome is a uint256 that gets truncated to a uint8 here:
```
    function getTokenId(IMarket _market, uint256 _outcome) public pure returns (uint256 _tokenId) {
        bytes memory _tokenIdBytes = abi.encodePacked(_market, uint8(_outcome));
        assembly {
            _tokenId := mload(add(_tokenIdBytes, add(0x20, 0)))
        }
    }
```

If you picked a special value for _longOutcome , like 257, you would end up with an _amount of extra tokens!
the tokenId would be exactly the same as _outcome=1, so outcome 1 would mint twice as many tokens as it was supposed to (which are not backed by CASH). One set goes to short recipient (along with all other outcomes), one set goes to long recipient.

nuevoalex:
Nice find!

Issue #4980, opened 2019-12-03
Sharetoken reading numTicks from an untrusted market


epheph:
I couldn't quite get this to actual exploit, but there's danger on this line:

https://github.com/AugurProject/augur/blob/67c0fec95e65fca8d27751f28a6c06e5247978c6/packages/augur-core/source/contracts/reporting/ShareToken.sol#L247

A user can directly call this function providing their own malicious _market, and could lie about numTicks, and it is not specifically verified to be a valid market. Later on, it is checked against its internal registry of markets[].(numTicks|numOutcomes) which needs to come from a real universe, so there might not be a specific vulnerability?

Seems like it could also be a re-entrancy point since it is user-provided with full gas forwarded?

In either case, i don't see a reason why it wouldn't just use the verified `market` values injected via `initializeMarket`

epheph:
As mentioned in #5120 , this is not an actual untrusted market due to burn just above. The oddity here is that it is calling into the market to get a value which is already available in local storage, due to solidity stack handling issues.

Issue #4977, opened 2019-12-02
RepPriceOracle can be re-initialized


epheph:
initialize does not call endInitialization, anyone can swap out augur, cash, or uniswap factory

https://github.com/AugurProject/augur/blob/8640bc720b02d9b616a6d52a9f468fc7f4e8c6fe/packages/augur-core/source/contracts/reporting/RepPriceOracle.sol#L29

Issue #4970, opened 2019-12-02
Can take anyone's "Cash" for validity bond


epheph:
**Describe the bug**
In V1, the validity bond was provided by forwarding along msg.value. In V2, it uses Augur.trustedTransfer to transfer cash from the sender to the market contract itself.

There is no restriction on
MarketFactory.createMarket to prevent someone from calling this with someone else's address as _sender:
```
    function createMarket(IAugur _augur, IUniverse _universe, uint256 _endTime, uint256 _feePerCashInAttoCash, IAffiliateValidator _affiliateValidator, uint256 _affiliateFeeDivisor, address _designatedReporterAddress, address _sender, uint256 _numOutcomes, uint256 _numTicks) public returns (IMarket _market);

```

So long as you provide a valid _augur and _universe (with you providing the REP) you can take someone else's Cash and put it into this new market

Furthermore, if you make "_sender" a different, already-deployed market contract, you would steal the validity bond from one market to another.

https://github.com/AugurProject/augur/blob/67c0fec95e65fca8d27751f28a6c06e5247978c6/packages/augur-core/source/contracts/factories/MarketFactory.sol#L27

This would also prevent a market from distributing its fees until someone else brought more Cash to the market.

PR #5120, opened 2019-12-08
Small high priorities and nit rollup

nuevoalex:
Fixes for issues:

#4980
#4970
#4977
#4993

Also includes already made PRs:

#5117
#5112
#5108
#5101
#5098
#4964


MicahZoltu (approved):

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
```diff
@@ -30,6 +30,7 @@ contract RepPriceOracle is IRepPriceOracle, Initializable {
         augur = _augur;
         cash = _augur.lookup("Cash");
         uniswapFactory = IUniswapV2Factory(_augur.lookup("UniswapV2Factory"));
+        endInitialization();
```
epheph, line 33: This seems to be done at the top usually, probably to avoid possibility of re-enter ?

packages/augur-core/source/contracts/reporting/ShareToken.sol
```diff
@@ -242,6 +241,7 @@ contract ShareToken is ITyped, Initializable, ERC1155, IShareToken, ReentrancyGu
         require(isApprovedForAll(_longParticipant, msg.sender) == true, "ERC1155: need operator approval to burn long account shares");
 
         _internalTransferFrom(_shortParticipant, _longParticipant, getTokenId(_market, _outcome), _amount, bytes(""), false);
+        // NOTE: burnCompleteSets will validate the market provided is legitimate
         (uint256 _payout, uint256 _creatorFee, uint256 _reportingFee) = burnCompleteSets(_market, _longParticipant, _amount, _sourceAccount, _fingerprint);
 
         uint256 _longPayout = _payout.mul(_price) / _market.getNumTicks();
```
epheph, line 247: what do you think of using 

`markets[address(_market)].numTicks` instead of `_market.getNumTicks()`

epheph (approved):
left some suggestions, but these changes address all the issues in the list

packages/augur-core/source/contracts/reporting/ShareToken.sol
nuevoalex: Causes stack depth issues sadly

epheph: That's too bad! it's definitely an unusual way to get numTicks when you are guaranteed to have it locally/cheaper/safer! Worth a comment?

And ya, burnSets will def protect from the untrusted market.

PR #5129, opened 2019-12-09
only init Gnosis on AVAILABLE event

### LINKED ISSUE #5096: error when login with portis with gnosis safe
web console errors, sometimes get 2 sometimes get one when logging in

image

andrewdamelio:
closes #5096

-Before we were initializing the gnosis safe on the events CREATED/AVAILABLE. This should only happen on AVAILABLE.

## Testing/QA
You should be able to create a new safe without console errors. Your safe should be usable also without having to refresh the page OR logout/login.


bthaile (approved):
it works, I was able to create a new safe with portis and make a trade.

PR #5133, opened 2019-12-09
Update Events.ts

pgebheim:
Throw detailed error if log's contract isn't known in the local ABI.


bthaile (approved):
lgtm

justinbarry:
It looks like this code was copied over to the ZeroX branch by hand.

Issue #5119, opened 2019-12-08
Can fire extraneous events.


MicahZoltu:
While in `ZeroXTrade.trade`, `transferFromAllowed` is set to true.  The user is given an opportunity to execute their own code during `_exchange.fillOrder(...)`, and during this time they can leverage that boolean value being set to true to call `safeTransferFrom` and `safeBatchTransferFrom` a number of times bounded only by gas.

These methods only emit an event, but extraneous emission of these events could cause DB corruption or other UI side problems.

Issue #5033, opened 2019-12-04
User-provided IExchange isn't required to validate 0x signature


epheph:
https://github.com/AugurProject/augur/blob/67c0fec95e65fca8d27751f28a6c06e5247978c6/packages/augur-core/source/contracts/trading/ZeroXTrade.sol#L215

The _exchange being called is provided by calldata, and while the intention is that they will provide a 0x exchange, the user can provide anything they want, including their own contract which never performs the signature check.

Once they pass that line, they have specified an order for a creator that is not themselves that they are can take at whatever price they want.

nuevoalex:
This is basically intentional. The Exchange is being provided in the signed order so that the system can handle upgrades to the exchange contract if they occur in the future. The expectation is that order validation off chain ensures the specified exchange is valid.

nuevoalex:
Talked about this more. The issue is that a malicious user can do this directly to force a trade from another account which did not sign anything.

epheph:
Additionally, there is no market validity check in the 0x trade code path, it is checked on every other entry into the order book

Issue #5104, opened 2019-12-06
Anyone can take eth from ZeroXTrade (possible re-entrancy to take from user)


epheph:
https://github.com/AugurProject/augur/blob/67c0fec95e65fca8d27751f28a6c06e5247978c6/packages/augur-core/source/contracts/trading/ZeroXTrade.sol#L232

If anyone calls ZeroXTrade.trade:
```
    function trade(
        uint256 _requestedFillAmount,
        bytes32 _fingerprint,
        bytes32 _tradeGroupId,
        IExchange.Order[] memory _orders,
        bytes[] memory _signatures
    )
``` 

with `_orders = []`, the contract will skip all processing and just send all eth it is sitting on. Normally, this isn't a huge deal (but perhaps still worth addressing?), since this contract should not be sitting on a balance, unless it is in the middle of a trade (to pay the _protocolFee per trade).

Depending on how https://github.com/AugurProject/augur/issues/5033 gets solved, there might be an opportunity for someone who has a malicious `getExchangeFromAssetData(_order.makerAssetData)` injected into the taker's order data (taker would need to get that bad data off-chain and bring it to the chain themselves) to enter into trade and take all the ether the user was going to use for _protocolFees.

epheph:
Also, while using a re-entrancy for a user's provided _exchange (even if signature matches like we discussed), they could go to town on `safeTransferFrom` and `safeBatchTransferFrom` emitting whatever log events they wanted. These don't move assets, so it would only affect logging, but it could be strange to see these confusing/misleading to have these events emitted

PR #5146, opened 2019-12-09
Zero x fixes

nuevoalex:
Fixes:

#5033 
#5119 
#5104


MicahZoltu (approved):
I believe this addresses the immediate issue.  I think it would be valuable to re-audit this contract though as when I was looking at it there were several attack avenues I didn't fully explore because this bug was certainly worse than the other avenues.

epheph:
I believe this does fully resolve #5119 .

It _DOES_ patch the hole for re-entering via user-provided _exchange ~~, but the assetData does not need to refer to the ShareToken , so it can enter any contract that fulfills the ERC-1155 interface, and while doing so, can emit any message it wants.~~ <- this is not true, i had a bad model of what was encoded in AssetData . everything is 👍

epheph:
It does patch #5033 .
For #5104 , anyone can still take ETH from ZeroXTrade, they just have to process an order of any size to do so. This might be "wontfix"

epheph:
As discussed in discord, this line:

https://github.com/AugurProject/augur/blob/master/packages/augur-sdk/src/state/db/ZeroXOrders.ts#L15

and the offset lines near the bottom will require changes of `-=64`

epheph (approved):

PR #5168, opened 2019-12-10
Fixes reducer method of empty array

bconfortin:
For some reason I removed the initial value of ZERO from the reducer on the code cleanup. Adding it back up to prevent this error: `Uncaught TypeError: Reduce of empty array with no initial value`.


bthaile (approved):
lgtm

PR #5179, opened 2019-12-10
Throw error when Gnosis relay returns incorrect safe address

### LINKED ISSUE #5116: Malicious relayer or MitM can give user a bad safe, or no safe and just pocket the fee.
https://github.com/AugurProject/augur/blob/7b9c2675588ac9a7f0b3a9dd740a37366953272f/packages/augur-sdk/src/api/Gnosis.ts#L299-L314

When creating a safe via a relayer, the UI will send a request to the relayer to create a safe and the relayer will respond with an address.  The UI currently doesn't validate the address before putting it into the "safesToCheck" queue.

A malicious relayer or man in the middle could return any address they want from that request, and the user would then send assets there.  Worse, the relayer could create an actual safe at that address but where the initializer is their own, rather than the expected initializer.  This would make it so the user sees an otherwise functioning safe that even matches the expected code, but the relayer could later steal any funds the user deposits during this session.

The fix for this is easy (once #4975 is resolved): Call `this.calculateGnosisSafeAddress` and compare it against `response.safe`.

justinbarry:
closes #5116


packages/augur-sdk/src/api/Gnosis.ts
```diff
@@ -305,32 +301,24 @@ export class Gnosis {
...
-      });
-
-      this.augur.setGnosisStatus(GnosisSafeState.WAITING_FOR_FUNDS);
-
-      return response;
-    } catch(error) {
-      const restoreAddress = await this.calculateGnosisSafeAddress({
-        payment: '1',
+      const calculatedSafeAddress = await this.calculateGnosisSafeAddress({
+        ...response,
```
MicahZoltu: It feels a bit dangerous to blindly use the response like this without validating that it contains reasonable data.  The `owner: params.owner` override I _think_ will protect against the obvious malicious relayer attack, but I'm mildly concerned that the relayer has the power to put basically whatever they want into this parameter.  "Probably fine", just a bit of a red flag.

MicahZoltu: Digging a bit more into this, it appears that `calculateGnosisSafeAddress` has a `param` whose type is unnecessarily restrictive.  I think it should really be `params: { owner: string, payment: string}`.  With that change, you could then change the callsite to `calculateGnosisSafeAddress({ owner: params.owner, response.payment })`.  Though, it is unclear to me what `payment` is or why you need to get it from the response.

packages/augur-sdk/src/api/Gnosis.ts
```diff
@@ -379,7 +366,7 @@ export class Gnosis {
         shareTokenAddress,
         affiliates,
         fingerprint,
-        referralAddress,
+        NULL_ADDRESS,
```
MicahZoltu: Style: I think having this named added value, since this is an otherwise an untyped tuple.

MicahZoltu (approved):
A couple comments, but no deal breakers that  I see.

packages/augur-sdk/src/api/Gnosis.ts
```diff
@@ -305,32 +301,24 @@ export class Gnosis {
...
-        paymentToken: null,
       });
+      if(ethUtil.toChecksumAddress(calculatedSafeAddress) !== ethUtil.toChecksumAddress(response.safe)) {
+        this.augur.setGnosisStatus(GnosisSafeState.ERROR);
+        throw new Error(`Potential malicious relay. Returned ${response.safe}. Expected ${calculatedSafeAddress}`);
+      }
 
       this.safesToCheck.push({
-        safe: ethUtil.toChecksumAddress(restoreAddress),
+        safe: ethUtil.toChecksumAddress(response.safe),
```
MicahZoltu, line 294: To avoid someone accidentally introducing a bug later, I would recommend changing this to `calculatedSafeAddress`.  We can assume in the current code that this is the same, but should the above conditional ever change that assertion would no longer hold.  To avoid accidentally introducing a security vulnerability it feels prudent to just use the version that we are confident is good.

packages/augur-sdk/src/api/Gnosis.ts
```diff
@@ -305,32 +301,24 @@ export class Gnosis {
...
         owner: params.owner,
         status: GnosisSafeState.WAITING_FOR_FUNDS,
       });
 
       this.augur.setGnosisStatus(GnosisSafeState.WAITING_FOR_FUNDS);
-      throw ethUtil.toChecksumAddress(restoreAddress);
-    }
+
+      return response;
```
MicahZoltu, line 303: I have a mild general concern about returning the `response` here rather than a specifically constructed object where we know exactly its contents.  Following this object around the entire codebase to figure out if there are any attack vectors by controlling its contents is much harder than just checking that the original function does proper validation.

MicahZoltu:
I left a couple more comments, the returning of `response` without destructuring it and validating its contents is of notable concern to me.  Going to leave the PR as approved, but I strongly encourage changing the pattern here to something like:
```typescript
const { apple, banana } = remoteCall()
validateOrThrowApple(apple)
validateOrThrowBanana(banana)
return { apple, banana }
```
This makes sure that any extra properties the `remoteCall` adds to the object won't get passed along, and it ensures that the properties it _does_ pass along have been validated.

packages/augur-sdk/src/api/Gnosis.ts
justinbarry: @MicahZoltu The payment is amount we need to send to the safe address for the relay to create the safe. It is basically a "bid" on the part of the relay for their safe creation services.

MicahZoltu: Ah, makes sense.  You may want to validate that the value you get back in there is a number (or whatever the expected type is) and potentially convert it to a `bigint` (or whatever you are using for big number storage in memory) rather tan leaving it as a string as you pass it around.

andrewdamelio (changes requested):
Should remove the below since your change above [L318-L332] removed the restored address getting thrown, in place of an error.

https://github.com/AugurProject/augur/pull/5179/files#diff-02ab91134a24c215795a88d9d2d6176bL207-L213

andrewdamelio (approved):

PR #5224, opened 2019-12-13
hide icons on left for mobile

phoebemirman:
fixes scott's issue in ui channel


bthaile (approved):
nice, that'll reduce confusion.

PR #5241, opened 2019-12-13
add date start validation

### LINKED ISSUE #5222: Add a check in the market creation form and a validation rule for templates
For Crypto templates (or any other that use this format)

We use the day of the exchange....users can still settle on that day but before the opening time that the exchange uses
so like chosing  dec 13 btc/usd on coinbse and having resolution time as dec 13th 1 am est

Need to add a check in market creation form and validation rule for templates that use this format

Create new input type DATE_START. this will represent the start of a day, and will need to be validated as such in market creation form and template validations.

1. Change crypto templates to use DATE_START instead of DATEYEAR.
1. need to run generate-templates flash script to rebuilt templates.ts (see flash README.md in augur-tools)
2. add new market creation validation so user can't create a market where the DATE_START is after market endTime (event expiration)
3. make sure in submit-new-market that value is timestamp, should be same as DATEYEAR 
4. need to create new template validation in templates-template to verify that DATE_START is before market endTime. To test use flash script validate-template (see flash README.md for details) update flash documentation for the new error should be similar as `estimated schedule date time is after market event expiration endTime`.

phoebemirman:
closes https://github.com/AugurProject/augur/issues/5222


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). ⛳️

**https://peek.run/93phqww**

-------

packages/augur-tools/src/templates-template.ts
```diff
@@ -314,6 +315,12 @@ function estimatedDateTimeAfterMarketEndTime(inputs: ExtraInfoTemplateInput[], e
   return Number(input.timestamp) > Number(endTime);
 }
 
+function dateStartAfterMarketEndTime(inputs: ExtraInfoTemplateInput[], endTime: number) {
+  const input = inputs.find(i => i.type === TemplateInputType.DATESTART);
+  if (!input) return false;
+  return Number(input.timestamp) > Number(endTime);
```
bthaile: add is equal to as well. start time needs to be before market endTime

packages/augur-tools/src/templates-template.ts
```diff
@@ -354,6 +361,12 @@ export const isTemplateMarket = (title, template: ExtraInfoTemplate, outcomes: s
       return false;
     }
 
+    // check DATESTART isn't after market event expiration
+    if (dateStartAfterMarketEndTime(template.inputs, new BigNumber(endTime).toNumber())) {
```
bthaile: dont need BigNumber here, timestamps are well within javascript number max value

bthaile (approved):
lgtm

Issue #5244, opened 2019-12-13
ZeroXTrade does not enforce ERC1155_PROXY_ID in AssetData


epheph:
While the off-chain calls to ZeroXTrade (createZeroXOrderFor) will encode the ERC1155_PROXY_ID (`0xa7cb5fb7`) into the first four bytes of the order that is submitted to 0x.fillOrder, there is no check inside the actual trade function to ensure that is what is being passed along to the 0x Exchange. Alternate PROXY_IDs could behave differently and might even look in different places for actual asset being transferred, which could lead to unexpected fill amounts being returned, and potentially even unexpected trades occurring (although the tokenAddress being immediately after proxy_id is likely to limit severity). New proxies could be added in the future with unexpected functionality.

Issue #5240, opened 2019-12-13
Gas savings, use library instead of calling into ShareToken for pure function getTokenId(s)


epheph:
getTokenId takes the market and outcome and smashes them together into a single uint256. In many spots, this is done by calling into the ShareToken contract to perform the calculation, incurring an extra cost of ~1000 gas. If each contract could instead use a library to do this, there would be decent gas savings (with small deployment cost increase)

https://github.com/AugurProject/augur/blob/67e3972432a68b4ba23ca1c3b9579b923857a77c/packages/augur-core/source/contracts/trading/CreateOrder.sol#L155

https://github.com/AugurProject/augur/blob/67e3972432a68b4ba23ca1c3b9579b923857a77c/packages/augur-core/source/contracts/trading/CreateOrder.sol#L179

https://github.com/AugurProject/augur/blob/67e3972432a68b4ba23ca1c3b9579b923857a77c/packages/augur-core/source/contracts/trading/CancelOrder.sol#L116

https://github.com/AugurProject/augur/blob/67e3972432a68b4ba23ca1c3b9579b923857a77c/packages/augur-core/source/contracts/trading/CancelOrder.sol#L119

https://github.com/AugurProject/augur/blob/dbdb6351ec6046ac1e1e509911171aec0f2cb5de/packages/augur-core/source/contracts/trading/FillOrder.sol#L205

https://github.com/AugurProject/augur/blob/dbdb6351ec6046ac1e1e509911171aec0f2cb5de/packages/augur-core/source/contracts/trading/FillOrder.sol#L211

https://github.com/AugurProject/augur/blob/dbdb6351ec6046ac1e1e509911171aec0f2cb5de/packages/augur-core/source/contracts/trading/FillOrder.sol#L231

https://github.com/AugurProject/augur/blob/dbdb6351ec6046ac1e1e509911171aec0f2cb5de/packages/augur-core/source/contracts/trading/FillOrder.sol#L237

Issue #5228, opened 2019-12-13
Market.InitialReporterWasCorrect returns true before initial report


epheph:
I mentioned this in #5070 as a comment, but i think it got missed

While the market is not finalized (hasn't set a payoutDistributionHash) and the initialReport has not been reported (`bytes32(0) == bytes32(0)`).

This does not impact code that depends on that value, but it is misleading and not true for anyone viewing the contract values or externally integrating with Augur. I could also imagine in the future someone assuming that `true` here means an initial report was completed without also checking that to be explicitly true.

Issue #4963, opened 2019-12-02
Make all private variables public, or ensure there are getters for each.


MicahZoltu:
https://github.com/AugurProject/augur/blob/c1973d49be54288cce7d1b5d0c021ce22a8092ed/packages/augur-core/source/contracts/Augur.sol#L72-L92

When auditing a deployed contract, it is important to be able to know what the current internal state of the contract is.  This is incredibly difficult when there are private members with no getters available.  You can sometimes derive answers to questions of "is this private variable set" by calling functions that will switch on that, but this is not ideal.

Easy solution is to just make all of those variables public (assuming no naming conflicts).  Alternative solution would be to add getters for any that are missing one (e.g., `shareTokens` doesn't appear to have a getter).

epheph:
Are you suggesting to change the actual production contracts that get deployed to open up all private variables?

MicahZoltu:
Yeah.  Having tried to audit production contracts with private variables, I can say that it is a huge PITA.  This is especially true when the private variables contain potentially security sensitive information like "who has admin access".  Private variables forces the auditor to try to derive current state by tracking all transactions sent to the contract over time, or using things like `getStorageAt` (which can be quite difficult to use).

MicahZoltu:
For example, if you have an `owner` variable and it is private, auditing to ensure that the owner has correctly been cleared may prove difficult, especially if there are multiple potential owner variables that can set each other.

PR #5247, opened 2019-12-14
Publicize gas and proxy check

nuevoalex:
Addresses:

#5244
#5228
#4963
#5240


packages/augur-core/source/contracts/libraries/TokenId.sol
```diff
@@ -0,0 +1,28 @@
+pragma solidity 0.5.10;
+
+import 'ROOT/reporting/IMarket.sol';
+
+
+library TokenId {
+
+    function getTokenId(IMarket _market, uint256 _outcome) internal pure returns (uint256 _tokenId) {
```
MicahZoltu, line 8: If this is going to pack an 8-bit value it should take an 8-bit value in.  Otherwise, you may run into a bug/exploit where someone passes in a large value and it gets truncated and the tokenID ends up matching another tokenID.

packages/augur-core/source/contracts/reporting/ShareToken.sol
```diff
@@ -8,13 +8,15 @@ import 'ROOT/libraries/Initializable.sol';
...
 import 'ROOT/CashSender.sol';
+import 'ROOT/libraries/TokenId.sol';
 
 
 /**
  * @title Share Token
  * @notice ERC1155 contract to hold all Augur share token balances
  */
 contract ShareToken is ITyped, Initializable, ERC1155, IShareToken, ReentrancyGuard, CashSender {
+    using TokenId for IShareToken;
```
MicahZoltu: I believe this line can be deleted.  Not only are you not using it anywhere, I don't believe it is actually valid since `TokenId` is not a library that operates on `IShareToken`.

packages/augur-core/source/contracts/trading/FillOrder.sol
```diff
@@ -13,10 +13,12 @@ import 'ROOT/trading/IProfitLoss.sol';
 import 'ROOT/trading/Order.sol';
 import 'ROOT/libraries/Initializable.sol';
 import 'ROOT/trading/IAugurTrading.sol';
+import 'ROOT/libraries/TokenId.sol';
 
 
 library Trade {
     using SafeMathUint256 for uint256;
+    using TokenId for IShareToken;
```
MicahZoltu: Same issue as mentioned above, this `using` statement can/should be removed.

packages/augur-core/source/contracts/trading/CreateOrder.sol
```diff
@@ -22,6 +23,7 @@ import 'ROOT/CashSender.sol';
 contract CreateOrder is Initializable, ReentrancyGuard, CashSender {
     using SafeMathUint256 for uint256;
     using Order for Order.Data;
+    using TokenId for IShareToken;
```
MicahZoltu: Remove using statement.

packages/augur-core/source/contracts/trading/CancelOrder.sol
```diff
@@ -16,13 +16,15 @@ import 'ROOT/IAugur.sol';
...
 import 'ROOT/CashSender.sol';
+import 'ROOT/libraries/TokenId.sol';
 
 
 /**
  * @title Cancel Order
  * @notice This allows you to cancel orders on the book.
  */
 contract CancelOrder is Initializable, ReentrancyGuard, ICancelOrder, CashSender {
+    using TokenId for IShareToken;
```
MicahZoltu: Remove using statement.

packages/augur-core/source/contracts/trading/ZeroXTrade.sol
```diff
@@ -351,6 +351,16 @@ contract ZeroXTrade is Initializable, IZeroXTrade, IERC1155 {
             address _kycToken
         )
     {
+         // Read the bytes4 from array memory
+        assembly {
+            _assetProxyId := mload(add(_assetData, 32))
+            // Solidity does not require us to clean the trailing bytes.
+            // We do it anyway
```
MicahZoltu: Why do you not have to do the logical and here?  Seems necessary to me...

MicahZoltu (approved):
No roadblocks, though I would like an answer to the question about the comment.

packages/augur-core/source/contracts/trading/ZeroXTrade.sol
nuevoalex: My understanding is the bytes4 value will not overflow but will just cut off the extra

MicahZoltu: Since you do the check anyway, I'll accept that statement at face value.  🙂

epheph: That seems magical you don't have to do it. Assembly doesn't know about the data type, so does non-assembly always mask the data it pulls off the stack?

packages/augur-core/source/contracts/libraries/TokenId.sol
epheph: Example: https://github.com/AugurProject/augur/issues/4993

epheph (approved):
This feels way better, having a library for these pure functions

packages/augur-core/source/contracts/trading/ZeroXTrade.sol
nuevoalex: Seems so.

```
contract Test {

    function decodeAssetData(bytes memory _assetData)
        public
        pure
        returns (
            bytes4 _assetProxyId
        )
    {
        assembly {
            _assetProxyId := mload(add(_assetData, 32))
            _assetProxyId := and(_assetProxyId, 0xFFFFFFFF00000000000000000000000000000000000000000000000000000000)
        }
        
        require(_assetProxyId == "0x22222222");

        return _assetProxyId;
    }
}
```

for example when passed "0x22222222" will succeed and return the expected result.

MicahZoltu:
```
contract Test {
    function decodeAssetData(bytes memory _assetData) public pure returns (bytes1 _assetProxyId)
    {
        assembly {
            _assetProxyId := mload(add(_assetData, 32))
        }
        require(_assetProxyId == 0xff);
        return _assetProxyId;
    }
}
```
I think this is the code that is surprising to @epheph and I.  If you pass in `0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff` I didn't expect the `require` check to pass, though now that I see it in action I suspect I know _why_ it passes (assignment to `bytes1` effectively does the truncation for you).

nuevoalex: oh I left the masking in heh. Yea it does the magic.

PR #5248, opened 2019-12-14
Adds require to ZeroXTrade.initialize

MicahZoltu:


packages/augur-core/source/contracts/trading/ZeroXTrade.sol
```diff
@@ -424,4 +425,4 @@ contract ZeroXTrade is Initializable, IZeroXTrade, IERC1155 {
     }
 
     function () external payable {}
-}
\ No newline at end of file
+}
```
MicahZoltu, line 428: I used the GitHub editor, not sure why it added the newline here.  If you care (e.g., repo style is no trailing newline) I can remove it, but if not I like the trailing newlines so will leave it.  🙂

PR #5264, opened 2019-12-16
Updates ZeroXExchange mainnet address to v3.

MicahZoltu:
Just in case someone forgets to update this before deployment, better to have latest than an old version that we know is incompatible.

Pulled address from https://github.com/0xProject/0x-monorepo/blob/development/packages/contract-addresses/addresses.json#L4


MicahZoltu:
Note: With MCD being released, I recommend filling in those addresses here too so your mainnet deploy doesn't accidentally end up with testnet MCD.

Issue #5265, opened 2019-12-16
Have deployment script check that if production is set, then all dependencies are external.


MicahZoltu:
Currently, the deployment script checks `externalAddresses` for various dependencies.  However, it doesn't appear there is a check that makes sure that if `networkConfiguration.isProduction` then all external dependencies _must_ be supplied.

Recommendation for handling this is to have each of the checks like
```typescript
if (externalAddresses.RepPriceOracle) {
    this.augur.registerContract('RepPriceOracle', externalAddresses.RepPriceOracle)
} else {
    this.uploadRepPriceOracle()
}
```
instead look something like this:
```typescript
if (this.configuration.isProduction || externalAddresses.RepPriceOracle) {
    assert(externalAddresses.RepPriceOracle)
    this.augur.registerContract('RepPriceOracle', externalAddresses.RepPriceOracle)
} else {
    this.uploadRepPriceOracle()
}
```
This way you can't accidentally deploy a testnet contract to production as the script will fail instead of happily deploying test code to production.

Issue #5260, opened 2019-12-15
0x protocol fee.


MicahZoltu:
https://github.com/AugurProject/augur/blob/b380634f04fcf5fa18e3ae3f7739c70f1dc41026/packages/augur-core/source/contracts/trading/ZeroXTrade.sol#L215

Currently the protocol fee is hard coded in the contracts.  This means that if the fee rate changes, then Augur will stop working with 0x.

However, if a change is made to blindly fetch the fee at runtime and forward the whole thing, then the 0x Exchange owner has the opportunity to exit scam by jacking the fee way up and seeing how many users confirm the transaction without checking the fee amount (my guess is most users will fall to this scam until word gets out).

I'm not sure how to address this in a way that both protects Augur users from malicious fee changes while also allowing non-malicious fee changes through.  Best option I can think of is to ensure that the UI makes it _incredibly_ prominent how much of a fee is being charged by 0x, and to protect Augur's reputation make it clear that this portion of the fee is a 0x fee, not an Augur fee.

MicahZoltu:
Another option would be to have the UI accept any fee changes within a range, but require a UI update to go beyond that.

epheph:
And pass the protocol fee in as an argument in the calldata?

MicahZoltu:
Yeah, I believe it is already part of the order.

PR #5272, opened 2019-12-17
Protocol fee and various nits

nuevoalex:
Addresses:

#5260

#5265

Includes:

#5262

#5263

#5248

#5264

#5275


packages/augur-core/source/contracts/trading/ZeroXTrade.sol
```diff
@@ -212,7 +213,7 @@ contract ZeroXTrade is Initializable, IZeroXTrade, IERC1155 {
 
         transferFromAllowed = true;
 
-        uint256 _protocolFee = 150000 * tx.gasprice;
+        uint256 _protocolFee = exchange.protocolFeeMultiplier().mul(tx.gasprice);
```
MicahZoltu, line 216: This will allow the 0x multisig to set the fee to whatever they want, and the fee goes to wherever the multisig wants.  What protections will be in place against a compromised/malicious 0x multisig account?  Will the UI provide some protections against this (e.g., hard code the fee in the UI so a change requires a UI update)?

nuevoalex: Yes the UI will. The UI will always inform the user what fees are being charged.

packages/augur-core/source/libraries/ContractDeployer.ts
```diff
@@ -82,23 +82,25 @@ Deploying to: ${networkConfiguration.networkName}
...
+            if (!externalAddresses.RepPriceOracle) throw new Error('Must provide RepPriceOracle');
             console.log(`Registering Rep Price Oracle Contract at ${externalAddresses.RepPriceOracle}`);
             await this.augur!.registerContract(stringTo32ByteHex('RepPriceOracle'), externalAddresses.RepPriceOracle);
         } else {
             await this.uploadRepPriceOracle();
         }
 
         // Cash
-        if (externalAddresses.Cash) {
+        if (this.configuration.isProduction || externalAddresses.Cash) {
```
MicahZoltu: It feels like this should be:
```typescript
if (this.configuration.isProduction
    || externalAddresses.Cash
    || externalAddresses.DaiVat
    || externalAddresses.DaiPot
    || externalAddresses.DaiJoin) {
    if (!(externalAddresses.Cash && externalAddresses.DaiVat && externalAddresses.DaiPot && externalAddresses.DaiJoin)) {
        ...
    }
}
```
As it currently stands, if `externalAddresses.DaiVat` is set but `externalAddresses.Cash` is not the code won't make it into this block.  That being said, with the `isProduction` check this is really just a testing thing so not a big deal.

MicahZoltu: Same comment for the other blocks that have multiple requirements.

packages/augur-core/source/libraries/ContractDeployer.ts
```diff
@@ -143,7 +145,8 @@ Deploying to: ${networkConfiguration.networkName}
         }
 
         // Proxy Factory & Gnosis Safe
-        if (externalAddresses.ProxyFactory) {
+        if (this.configuration.isProduction || externalAddresses.GnosisSafe) {
+            if (!externalAddresses.GnosisSafe) throw new Error('Must provide GnosisSafe');
```
MicahZoltu: I'm curious why you didn't put this check inside the check below.  I personally like the individual checks like this rather than multi-checks, but it doesn't follow the same style as other sections.

MicahZoltu (approved):

packages/augur-core/source/libraries/ContractDeployer.ts
nuevoalex: I'm just putting it after the first condition everywhere

packages/augur-core/source/libraries/ContractDeployer.ts
MicahZoltu: Actually, due to the addition of `isProduction` the check below **MUST** test that `externalAddresses.Cash` is provided.

MicahZoltu (changes requested):
Changed my mind on the approval.  Needs a test of `externalAddresses.Cash` in case `this.configuration.isProduction` is set but `externalAddresses.Cash` is not.

MicahZoltu (approved):

PR #5276, opened 2019-12-17
Renames variables to be more accurate.

MicahZoltu:
I spent a lot of time trying to figure out what this code did because the variable name `_amountRemainingFillerWants` implies something different from what the variable actually contains.  In reality, this value is the min of "number of shares the order filler wants to trade" and "number of shares the order creator wants to trade".  The variable name previously implied that it was the amount that the filler wants only, which isn't true.  It is worth noting that the code `uint256 _amountFilled = _amountFillerWants.sub(_tradeSize);` is now more clearly incorrect, really that variable holds `_amountFillerWantsRemaining`, so it shouldn't be passed to `updateProfitLoss`!  I have fixed this as well.

### DIFF
--- a/packages/augur-core/source/contracts/trading/FillOrder.sol
+++ b/packages/augur-core/source/contracts/trading/FillOrder.sol
@@ -515,8 +515,8 @@ contract FillOrder is Initializable, ReentrancyGuard, IFillOrder {
 
         sellCompleteSets(_tradeData);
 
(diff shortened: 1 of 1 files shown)


MicahZoltu:
Cleared up in chat.  `tradeData` is mutated as functions are called on it such that `filler.sharesToSell` and `filler.sharesToBuy` are decremented as the trade occurs.  By the time this code is reached, all that is remaining are shares to buy/sell that were not fulfilled by the order processing code above, thus the variable names are correct.

PR #5277, opened 2019-12-17
Minor: Improves naming of local variable.

MicahZoltu:
Just trying to make this function a bit more clear and easy to read.  Very much a non-critical change.

PR #5284, opened 2019-12-18
0x trade interval requirement

nuevoalex:
Labeled work in progress since 0x is in flux atm and I want it to work before adding a thing that intentionally makes it fail


packages/augur-core/source/contracts/WarpSync.sol
```diff
@@ -21,7 +21,7 @@ contract WarpSync is IWarpSync, Initializable {
 
     uint256 private constant MARKET_LENGTH = 1 days;
     uint256 private constant MAX_NUM_TICKS = 2 ** 256 - 2;
-    int256[] private PRICES = [0, 1 ether];
+    int256[] private PRICES = [int256(uint256(1) << 255), int256(~(uint256(1) << 255))];
```
MicahZoltu: Wat?  At the least, this magic horribleness is deserving of a comment...

packages/augur-core/source/contracts/Augur.sol
```diff
@@ -93,6 +93,8 @@ contract Augur is IAugur, IAugurCreationDataGetter, CashSender {
 
     int256 private constant DEFAULT_MIN_PRICE = 0;
     int256 private constant DEFAULT_MAX_PRICE = 1 ether;
+    uint256 private constant RECOMMENDED_TRADE_INTERVAL = 10;
+    uint256 private constant DEFAULT_RECOMMENDED_TRADE_INTERVAL = 10 ** 17;
```
MicahZoltu: These two variables are named very similarly, implying that one is just a default for the other (which is odd when both are constants), but they appear to have different types (one is maybe a fixed point number, while the other is just a normal number).  Recommend renaming these or doing something to improve the clarity as to their purpose.

MicahZoltu: Remove the "Recommended" bit from this.  From what I can tell these are hard coded values that cannot be changed, so there is nothing about them that makes them just a "recommendation".

nuevoalex: I labeled it this way since this is in the core contracts here where this isn’t enforced in any way, but is a recommended value for use by external trading contracts

nuevoalex: Yea I’ll think more on the naming

MicahZoltu: Oh, I just noticed that this is in the core contracts.  Can we make it so that this value is a market input parameter instead of hard coded?  The UI can then fix it, and optionally the UI could even hide any market that doesn't set it to a reasonable value, but that way we _can_ change it later.

From the point of view of the core contracts, this value doesn't prevent the market from being valid and reportable, so I don't think it should be hard coded in the core contracts.

nuevoalex: We don't want it to be configurable really and because of scalars it will be variable. The purpose of doing it at market creation like this is to save gas since there isnt really a registration of markets in the trading contracts so we'd have to at a minimum waste gas checking if it has been precalculated.

nuevoalex:
Now also covers:

#5277

#5286

#5285

packages/augur-core/source/contracts/trading/ZeroXTrade.sol
```diff
@@ -308,9 +317,14 @@ contract ZeroXTrade is Initializable, IZeroXTrade, IERC1155 {
...
             _tokenValues,
-            _callbackData,
-            _kycToken
+            _callbackData
         );
+
+        uint256 _assetDataLength = _assetData.length;
+
+        // We must pad the kycToken as the 0x exchnage requires the assetData be a multiple of 32 not counting the selector
+        _assetData = abi.encodePacked(_assetData, bytes32(uint256(address(_kycToken))));
```
MicahZoltu, line 326: Won't `abi.encode(_assetData, _kycToken)` achieve this?  Or will it do other things to the `_assetData` that are undesirable?

nuevoalex: The comment above the line explains. A simple `abi.encode` will not have the correct length and will get rejected by 0x.

packages/augur-core/source/contracts/trading/ZeroXTrade.sol
```diff
@@ -20,6 +21,8 @@ contract ZeroXTrade is Initializable, IZeroXTrade, IERC1155 {
 
     bool transferFromAllowed = false;
 
+    uint256 constant public TRADE_INTERVAL_VALUE = 10 ** 19; // Trade value of 10 DAI
```
MicahZoltu, line 24: Something to keep in mind, if Maker does a global shutdown DAI will become pegged to ETH, not USD and "10 DAI" has potential to be worth a lot or a little.

packages/augur-core/source/contracts/trading/ZeroXTrade.sol
MicahZoltu: Ah, it is because `abi.encode` will offset the dynamic `_assetData`:
> Non-standard Packed Mode [...] dynamic types are encoded in-place and without the length.

packages/augur-core/source/contracts/Augur.sol
MicahZoltu: Looks like these got left in from an earlier iteration.  Can be removed.

packages/augur-core/source/contracts/WarpSync.sol
MicahZoltu: Consider: `[int256(2**255), int256(2**255-1)]`.  Still not great, but maybe better?

MicahZoltu: Another option for ultimate clarity would be:
```
int256 private INT256_MIN = int256(2**255);
int256 private INT256_MAX = int256(2**255 - 1);
int256[] private PRICES = [INT256_MIN, INT256_MAX];
```

packages/augur-core/source/contracts/trading/ZeroXTrade.sol
```diff
@@ -69,13 +72,15 @@ contract ZeroXTrade is Initializable, IZeroXTrade, IERC1155 {
...
 
+    IAugur augur;
     IFillOrder public fillOrder;
     ICash public cash;
     IShareToken public shareToken;
     IExchange public exchange;
 
     function initialize(IAugur _augur, IAugurTrading _augurTrading) public beforeInitialized {
         endInitialization();
+        augur = _augur;
```
MicahZoltu, line 83: You don't appear to actually use this, did you mean to add it?

nuevoalex: Its used to check allowances in the balanceOf calls

packages/augur-core/source/contracts/trading/ZeroXTrade.sol
```diff
@@ -69,13 +72,15 @@ contract ZeroXTrade is Initializable, IZeroXTrade, IERC1155 {
     // solhint-disable-next-line var-name-mixedcase
     bytes32 public EIP712_DOMAIN_HASH;
 
+    IAugur augur;
```
MicahZoltu, line 75: See comment below about deletion, but if you keep it this should have an explicit visibility modifier.

packages/augur-core/source/contracts/trading/ZeroXTrade.sol
nuevoalex: oh thats a different PR actually... Well it _will_ be doing that heh

MicahZoltu: Nitpick: Consider moving this change to that other PR then.  😄  If not, see comment above about modifier.

packages/augur-core/source/contracts/trading/ZeroXTrade.sol
```diff
@@ -245,12 +250,16 @@ contract ZeroXTrade is Initializable, IZeroXTrade, IERC1155 {
...
-        (IERC1155 _zeroXTradeToken, uint256 _tokenId) = getZeroXTradeTokenData(_order.makerAssetData);
+    function validateOrder(IExchange.Order memory _order, uint256 _fillAmountRemaining) internal view {
+        (IERC1155 _zeroXTradeTokenMaker, uint256 _tokenIdMaker) = getZeroXTradeTokenData(_order.makerAssetData);
         (IERC1155 _zeroXTradeTokenTaker, uint256 _tokenIdTaker) = getZeroXTradeTokenData(_order.takerAssetData);
-        require(_zeroXTradeToken == _zeroXTradeTokenTaker);
-        require(_tokenId == _tokenIdTaker);
-        require(_zeroXTradeToken == this);
+        (address _market, uint256 _price, uint8 _outcome, uint8 _type) = unpackTokenId(_tokenIdMaker);
+        uint256 _numTicks = IMarket(_market).getNumTicks();
+        uint256 _tradeInterval = TRADE_INTERVAL_VALUE / _numTicks;
```
MicahZoltu, line 258: Consider some renaming for clarity:
```solidity
uint256 _tradeIntervalInShares = TRADE_INTERVAL_IN_DAI / _numTicks;
```

packages/augur-core/source/contracts/trading/ZeroXTrade.sol
MicahZoltu, line 24: I generally say that storage values should be public, but in the case of constants I think it just adds noise and doesn't add value.

MicahZoltu (approved):
6am Micah approves this PR.  Left some comments to help improve readability.

PR #5301, opened 2019-12-19
update templates and add some flash scripts

### LINKED ISSUE #5290: Template fixes after Review
Go to Politics, world, multiple choice:
Who will be [Position] of [Location] by [Specific Datetime]?

1) Position drop down ... Change Prince to Crown Prince

2) Also need to add an additional Rule to this market under  Added Resolution Rules saying:

If the Country in this market has multiple people with the same title, the market should resolve as invalid.

Crypto Market (need to add a sentence into all crypto markets using trading view in its templates)

Add:  To do this you would go to 00:00 on the date specified in the market question and depending on where you are located, you count backwards or forwards the number of candlesticks your time zone is.  For example if you are located in UTC -5 and you wanted to find the Open Price for December 16th on Coinbase, you would go to the hourly chart and search for 00:00 on December 16th and count 5 candlesticks backwards to December 15th 1900 EST.

*****Validate timezones to exchanges for Event resolution times

bthaile:
fixes https://github.com/AugurProject/augur/issues/5290

also added a flash script for creating zerox orders and convenience script for creating markets.


packages/augur-tools/src/templates-lists.ts
```diff
@@ -235,8 +235,8 @@ export const LIST_VALUES = {
       label: 'King',
     },
     {
-      value: 'Prince',
-      label: 'Prince',
+      value: 'Crowned Prince',
+      label: 'Crowne Prince',
```
phoebemirman: typo here

phoebemirman (changes requested):
typo - missing a d

packages/augur-tools/src/templates-lists.ts
bthaile: fixed typo

bthaile:
I looked it up `Crown` price seems to be the correct term.

bthaile:
<https://en.wikipedia.org/wiki/Crown_prince> it's a title `crown prince` not the action of crowning or been crowned.

phoebemirman (approved):

PR #5303, opened 2019-12-20
Trivial gas savings and minor clarity improvement.

MicahZoltu:
The primary purpose of this change is to make it a bit more clear to the reader that the price of the parent universe token will be used if a parent universe exists, otherwise the genesis universe price will be used.  As a side effect, I believe we'll save a trivial amount of gas in one of the code paths (but that is inconsequential).


justinbarry:
@nuevoalex will cherry-pick these commits to another branch he is currently working on.

PR #5305, opened 2019-12-20
Correct early return for first price calculation.

MicahZoltu:
The exchange may be set (via `getOrCreateUniswapExchange`) but the price not yet set.  In such a situation we want to return early here with the initial price rather than proceeding to the calculations (which would end up as `0`).

I also removed an unnecessary local variable for `_exchange`.

### DIFF
--- a/packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
+++ b/packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
@@ -56,13 +56,12 @@ contract RepPriceOracle is IRepPriceOracle, Initializable {
             return _exchangeData;
         }
 
-        IUniswapV2 _exchange = _exchangeData.exchange;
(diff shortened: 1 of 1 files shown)


epheph:
This would now enter the conditional when the internally tracked price starts at initialPrice, but then declines all the way back to absolutely 0? Is getting to absolute 0 even possible, since it early-exists when either side is 0? It might be.

MicahZoltu:
Everything starts with `0` values (default for Solidity).  The exchange address is set at some point, but the price is set in that same transaction.  Prior to this PR if you entered this code while in that state (exchange set but price is not) you would skip the early return and use `0` as the historic weighted average price.  You would then need to slowly work off that `0` historic value over time, resulting in the feed read being too low for some time.

This PR changes it so until the price is set, the early return will be hit.  I believe the price cannot be set until _after_ the exchange is set, so we'll bypass the early return in any case where the exchange or the price are not yet set.

Issue #5030, opened 2019-12-04
Design QA: App bar + nav bar: Nav bar styling


matt-bullock:
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/79238a9e-3dd6-40e6-b17a-278fa5cd0273)

https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=425%3A93

- [x] 48px height

- [x] Update text size and style

- [x] Check background colour is correct

- [x] Active item underline sits on the bottom of bar

matt-bullock:
Background colour should be #161518

matt-bullock:
Font should be Mono Bold 12px

PR #5310, opened 2019-12-23
fix-module-header-background

JohnDanz:
fixed the background color for module-headers so that it uses table background instead of dark grey

https://github.com/AugurProject/augur/issues/5030


bthaile (approved):
lgtm

Issue #5314, opened 2019-12-24
kycToken.balanceOf provides re-entrancy point for cancel+fill


epheph:
There is a possibility to steal the escrow of other users via a malicious re-entry using the kycToken to cancel an order (receiving escrowed funds back) after the FillOrder has loaded the order data but before it writes that data back out. This only takes a single attacker to execute and does not require a specific interaction from a victim. The victim here is the AugurTrading contract itself. The steps are:

1.) Upload a "kycToken" which you control, which has a balanceOf function which cancels an augur on-chain order. This contract will also be the one that owns Cash and interacts with the Augur order book.
2.) Add an order to the Augur order book via the above kycToken contract. For this example, place an order for 2 "YES" shares offering 1 ether (pretend this is a reasonable price). Use this kycToken contract itself as the kycToken. 1 ether is escrowed into AugurTrading
3.) With a random EOA account,  take this order via `FillOrder.publicFillOrder`. For simplicity, this account owns 2 "YES" shares outright (but this could be done via any of the `tradeMakerXForFillerY`)
4.) fillOrder loads up the order data
5.) fillOrderInternal then makes a call to `_kycToken.balanceOf(filler)`. This function will cancel the original order, refunding the original 1 ether to the kycToken contract itself. kycToken contract side is now back to even. balanceOf(filler) returns 1 so the transaction proceeds.
6.) The rest of the order fill takes place with the data loaded into memory from before the cancel. EOA account will send 2 "YES" and still receives 1 ETH from AugurTrading. kycToken contract will receive these 2 "YES".

In the end, AugurTrading has lost 1 ETH, having given both kycToken and EOA 1 ETH. 2 yes shares simply changed hands from EOA to kycToken (both which the attacker controls)

Loads up the data:
https://github.com/AugurProject/augur/blob/6b856caae657a83fa3cdfe9e8532dc2ecf7a5ecf/packages/augur-core/source/contracts/trading/FillOrder.sol#L484-L490

then calls kycToken.balanceOf:
https://github.com/AugurProject/augur/blob/6b856caae657a83fa3cdfe9e8532dc2ecf7a5ecf/packages/augur-core/source/contracts/trading/FillOrder.sol#L507

Issue #5312, opened 2019-12-23
publicCreateOrders does not ensure dynamic arrays are identical length


epheph:
https://github.com/AugurProject/augur/blob/6b856caae657a83fa3cdfe9e8532dc2ecf7a5ecf/packages/augur-core/source/contracts/trading/CreateOrder.sol#L95

The impact here is partially limited by Order.create checks for `_prices` and `_attoshareAmounts` being > 0, but if the _outcomes or _types array length is less than _types, the orders will accidentally be placed for invalid or `BID`.

Issue #5309, opened 2019-12-21
Rename RepPriceFeed internal variables.


MicahZoltu:
https://github.com/AugurProject/augur/blob/602a716c5cda08b28c8f66596f6658c384d88265/packages/augur-core/source/contracts/reporting/RepPriceOracle.sol#L70-L71

I spent quite a bit of time trying to figure out how this code worked because these variable names imply that they represent the current Uniswap reserves, not the accumulators.  Recommend renaming these variables to more clearly indicate that they are accumulators, not reserves/amounts.

PR #5319, opened 2019-12-24
Rename and order safety

nuevoalex:
Includes:

#5309

#5312

#5314


packages/augur-core/source/contracts/trading/FillOrder.sol
```diff
@@ -491,20 +493,22 @@ contract FillOrder is Initializable, ReentrancyGuard, IFillOrder {
...
+        IERC20 _kycToken = storedContracts.orders.getKYCToken(_orderId);
+        require(_kycToken == IERC20(0) || _kycToken.balanceOf(_filler) > 0, "FillOrder.fillOrder: KYC token failure");
         Trade.Data memory _tradeData = Trade.create(storedContracts, _orderId, _filler, _amountFillerWants, _fingerprint);
         return fillOrderInternal(_filler, _tradeData, _amountFillerWants, _tradeGroupId);
     }
 
     function fillZeroXOrder(IMarket _market, uint256 _outcome, IERC20 _kycToken, uint256 _price, Order.Types _orderType, uint256 _amount, address _creator, bytes32 _tradeGroupId, bytes32 _fingerprint, address _filler) external returns (uint256) {
         require(msg.sender == zeroXTrade);
         require(augur.isKnownMarket(_market));
+        require(_kycToken == IERC20(0) || _kycToken.balanceOf(_filler) > 0, "FillOrder.fillOrder: KYC token failure");
```
epheph, line 505: Re-entering is technically still possible here, but looks to be without effect, since this call happens after the ZeroXTrade code, after the actual 0x fill (for maker amount remaining) and before a few checks. The code being skipped is here:
```
        if (!creatorHasFundsForTrade(_order, _amount)) {
            return 0;
        }
        // If the maker is also the taker we also just skip the trade
        if (_order.makerAddress == _taker) {
            return 0;
        }
```

Which are just optimizations, so this looks good to me. Since these trades are two direct transferFrom's, without pooling other resources together to withdraw extra from, the potential for exploit is more limited than a DEX with pooled maker assets.

epheph (approved):

Issue #5271, opened 2019-12-16
Remove commas to separate thousands in Quantities only in Order book and Trade History


matt-bullock:
Only remove from these two places

PR #5323, opened 2019-12-26
remove-commas-orderbook-tradehistory

JohnDanz:
removed commas for the trade history and order book. did this by adding a new option to format-number called `removeComma` which defaults to false.
https://github.com/AugurProject/augur/issues/5271


bthaile:
.value has no commas. I dont think we need to parse already formatted value.

bthaile (approved):
Lgtm

PR #5327, opened 2019-12-27
Gnosis safe proxy callback

nuevoalex:
Addresses the vulnerability issue when doing the automatic registration of gnosis safe by using the new callback function of the proxy factory


packages/augur-core/source/contracts/GnosisSafeRegistry.sol
```diff
@@ -40,46 +56,83 @@ contract GnosisSafeRegistry is Initializable {
...
+        assembly {
+            // Skip selector and length:
+            let _initializerData := add(_initializer, 36)
+            // Load the to and data params
+            _to := mload(add(_initializerData, 64))
+            _data := add(_initializerData, mload(add(_initializerData, 96)))
+        }
+        require(_to == address(this));
+        // Requires the expected and actual data arguments are equal
+        require(_data.length == _expectedSetupForAugurData.length && keccak256(_data) == keccak256(_expectedSetupForAugurData));
```
epheph, line 110: keccak256 equality isn't enough? Not complaining, just curious

nuevoalex: it _might_ be, but figured this would ensure there were no possible ways to fake equality

packages/augur-core/source/contracts/GnosisSafeRegistry.sol
```diff
@@ -40,46 +56,83 @@ contract GnosisSafeRegistry is Initializable {
...
+        // Now validate the address of the safe is what should have been produced via create2
+        uint256 _saltNonceWithCallback = uint256(keccak256(abi.encodePacked(_saltNonce, address(this))));
+        bytes32 _salt = keccak256(abi.encodePacked(keccak256(_initializer), _saltNonceWithCallback));
+        address _expectedAddress = generateCreate2(address(proxyFactory), _salt, deploymentData);
+        require(_proxy == _expectedAddress);
```
epheph, line 115: It doesn't help with security, but it might be a good spot to emit some of these values so the UI or others could discover what safes are valid without trying various different values?

epheph:
The contract logic looks good. You can enter an alternate callback/GnosisSafeRegistry, but i don't see any vulnerabilities around it

packages/augur-core/source/contracts/GnosisSafeRegistry.sol
nuevoalex: What do you mean exactly? Any safe in the registry would be valid no?

epheph (approved):
🎸

packages/augur-core/source/contracts/GnosisSafeRegistry.sol
epheph: I just mean emitting "this user has a valid safe, and it used this nonce and initializer hash" so the UI could hook into it, or if the initializer changes, you could say "ohh,they used THIS initializer which we also consider valid"

epheph: But if you never need to scan and always just need to look up a specific user, prolly not worth adding a log

Issue #4880, opened 2019-11-22
Design QA: Market List Page- Cards - Condensed View - Desktop


matt-bullock:
Design: https://www.figma.com/file/Do4qnkCNM8P7lwBmCdYZCS4X/Market-Cards?node-id=186%3A262

 - [x] Update Label Style: Total Volume and event expiration. All should be the same label style

- [ ] Update number and date style: number for total vol and date should all be the same style

- [ ] Check share icon button is in with correct share menu - See this ticket: https://app.zenhub.com/workspaces/augur-5bfed3424b5806bc2bf8d173/issues/augurproject/augur/4547

- [x] Move up Total Volume so Total Vol label is in line with categories. Should also be in line with event expiration

- [x] Reduce text size of Market title on condensed cards - see design

- [x] Add dollar sign to start of Total Volume

- [ ] Move icons indicating if user has a position in dai, a stake in rep, MC or DR to the bottom left: 8px from bottom, 12px from left - see designs

#### Interactions ####

- [x] Make interactions the same as classic view

matt-bullock:
Update volume number to 12px (same as the date)

matt-bullock:
Can you update the icons to sit 8px from the bottom so there is a space between icon and volume amount when the cards have a single line market title. this is what's happening in build:

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/b9a9839b-6ce6-4306-b924-982accb78f1c)

phoebemirman:
ok im only seeing this for the condensed view

matt-bullock:
yeah that's ok, the classic view has enough space

Issue #4878, opened 2019-11-21
Design QA: Market List Page- Cards - Classic View - Desktop


matt-bullock:
Design: https://www.figma.com/file/Do4qnkCNM8P7lwBmCdYZCS4X/Market-Cards?node-id=186%3A5427

 - [x] Update Label Style: Total Volume, Open Interest and event expiration. All should be the same label style

- [x] Update number and date style: number for total vol, OI and date should all be the same style

- [ ] Check share icon button is in with correct share menu - See this ticket: https://app.zenhub.com/workspaces/augur-5bfed3424b5806bc2bf8d173/issues/augurproject/augur/4547

- [x] Move up Total Volume and OI so Total Vol label is in line with categories. Should also be in line with event expiration

- [x] Change copy on show all outcomes button- see design for latest

- [x] Show all/less button should have 12px spacing top and bottom

- [x] Add dollar sign to start of Total Volume and Open interest

- [ ] Update default state of favourite market star icon

- [x] Move icons indicating if user has a position in dai, a stake in rep, MC or DR to the bottom left: 8px from bottom, 12px from left - see designs

#### Outcomes ####

- [ ] Update height of outcome box - see design
- [x] Round corners on percentage line
- [x] Percentage line: 2px height
- [ ] Percentage line: 8px spacing left and right
- [ ] Percentage line: 8px spacing from bottom
- [ ] Percentage  line: 8px spacing from outcome name
- [ ] Outcome name: 8px spacing from left
- [ ] Last price percentage number: 8px spacing from right
- [ ] Last price percentage number: Update styling and colour
- [ ] Last price percentage number: Change to 0.00% rather than -%
- [x] When at 0.00% make text colour secondary text colour

#### Interactions ####
- [x] If user click anywhere on card that isn't already an interaction it should take them to the market page.
- [ ] Hover state to show the above: Add grey outline to card. Design: https://www.figma.com/file/Do4qnkCNM8P7lwBmCdYZCS4X/Market-Cards?node-id=3033%3A0
- [ ] Hover state when hovering on an outcome: becomes lighter grey fill and stroke. Design: https://www.figma.com/file/Do4qnkCNM8P7lwBmCdYZCS4X/Market-Cards?node-id=3033%3A1

- [ ] Clicking an outcome should take you to the market page with that outcome already selected on the market page

- [ ] Hovering shield icon brings up tooltip. Copy: Template markets have pre-defined terms and a greater chance of validly resolving than custom markets
[...]

PR #5348, opened 2019-12-31
restyle condensed and normal market cards to mock

phoebemirman:
https://github.com/AugurProject/augur/issues/4880
https://github.com/AugurProject/augur/issues/4878


bthaile (approved):
Code looks good

PR #5355, opened 2020-01-01
Exchange closing validations

### LINKED ISSUE #5300: Finance- Event resolution time Can Not be before Market Closing & Closing Time List
Need to get the Market exchange closing times for exchanges in the dropdown (also the closing times for indexes)

Once and exchange (or index is selected in the market question). this should trigger not allowing Event expiration to be created until 2 hrs after the closing time.

We must acct for daylight savings and will use a 2 hr buffer

For Example if Nasdaq was selected (closing bell is 5 pm est)  Earliest Event expiration could start would be 7 pm on same day as the market

Need to Not allow for user to select a weekend as well in creation of Market Question.

Also need to add an Added Resolution Rule about Weekends and or Holidays when market is closed

We need to add in and Additional Rules: If it’s a stock, currency or cryptocurrency and its ticker is not used in the market question, the market should be "Invalid".

***Remove from Index dropdown list:
Dow Jones Global Titans
FTSE All-World Index
MSCI World Index
S&P Global 100 Index

https://docs.google.com/spreadsheets/d/1Zf98WJZLvXlR94YTqpeaGjuuS4lF1oZLcMAq5Wt0hKA/edit?usp=sharing

bthaile:
closes https://github.com/AugurProject/augur/issues/5300

add exchange closing times to templates and market creation. Don't allow user to have market endtime before exchange close time on same day as indicated in the market question. Added new resolution rule. 

image

template markets using new exchange closing gets recreated correctly
image


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). 🏅

**https://peek.run/1rehw35**

-------

packages/augur-tools/src/templates-lists.ts
```diff
@@ -8113,6 +8113,95 @@ export const LIST_VALUES = {
...
+  },
+  'Toronto Stock Exchange (traded in Canadian Dollar)': {
+    offset: -4,
+    hour: 16,
+    minutes: 0,
+  },
+  'Euronext NV (traded in Euro)': { offset: 2, hour: 17, minutes: 30 },
+};
+
+export const FIN_INDEXES_CLOSING_OFFSETS = {
```
andrewdamelio, line 8161: do we need a ```FIN_INDEXES_HOILDAYS_OFFSETS```

bthaile: I could combine them but they are different. Exchanges and Indexes even though indexes are traded on exchanges. I would like to keep them separate

andrewdamelio (approved):
LGTM

PR #5365, opened 2020-01-03
Uniswap full integration

nuevoalex:
Does not include the actual Uniswap and UniswapFactory contracts yet as they aren't public


packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
```diff
@@ -1,25 +1,35 @@
...
 import 'ROOT/reporting/IRepPriceOracle.sol';
+import 'ROOT/ens/IENSRegistry.sol';
+import 'ROOT/ens/IENSResolver.sol';
 
 
 contract RepPriceOracle is IRepPriceOracle, Initializable {
 
+    uint256 constant Q112 = 2**112;
+
+    bytes32 public constant UNISWAP_REGISTRY_ENS_NAME = 0xf259e1e59b6e1e9de21a18289a53332ebc255d2e4ece8a91e0e90c628a9c7f87; // "uniswapv2.eth"; // TODO: set this to the actual namehash of the owned ENS name
```
MicahZoltu: Use a contract address, not an ENS name.  ENS ownership is (usually) centralized/controlled so the ENS name owner for `uniswapv2.eth` could change what it points to at any time.

packages/augur-core/source/contracts/uniswap/interfaces/IUniswapV2Factory.sol
```diff
@@ -1,7 +1,15 @@
...
 
 
 interface IUniswapV2Factory {
-    function getExchange(address tokenA, address tokenB) external view returns (address);
+    event ExchangeCreated(address indexed token0, address indexed token1, address exchange, uint256);
+
+    function exchangeBytecode() external view returns (bytes memory);
+
+    function sortTokens(address tokenA, address tokenB) external pure returns (address token0, address token1);
+    function getExchange(address tokenA, address tokenB) external view returns (address exchange);
```
MicahZoltu: This should return an `IUniswapExchange`, not an `address`.

packages/augur-core/source/contracts/uniswap/interfaces/IUniswapV2Factory.sol
```diff
@@ -1,7 +1,15 @@
...
+pragma solidity 0.5.15;
 
 
 interface IUniswapV2Factory {
-    function getExchange(address tokenA, address tokenB) external view returns (address);
+    event ExchangeCreated(address indexed token0, address indexed token1, address exchange, uint256);
+
+    function exchangeBytecode() external view returns (bytes memory);
+
+    function sortTokens(address tokenA, address tokenB) external pure returns (address token0, address token1);
```
MicahZoltu: This should return two `IErc20`s, not two addresses.

MicahZoltu: Should also take two `IErc20`s.

packages/augur-core/source/contracts/uniswap/interfaces/IUniswapV2Factory.sol
MicahZoltu: Should also take `IErc20` parameters, not `address` parameters.

packages/augur-core/source/contracts/uniswap/interfaces/IUniswapV2Factory.sol
```diff
@@ -1,7 +1,15 @@
...
+    function getExchange(address tokenA, address tokenB) external view returns (address exchange);
+    function exchanges(uint) external view returns (address exchange);
```
MicahZoltu: Should return an `IUniswapExchange`, not an `address`.

packages/augur-core/source/contracts/uniswap/interfaces/IUniswapV2.sol
```diff
@@ -1,6 +1,26 @@
-pragma solidity 0.5.10;
+pragma solidity 0.5.15;
 
 
 interface IUniswapV2 {
```
MicahZoltu: Since Uniswap consists of two contracts, consider renaming this interface to `IUniswapExchange` for clarity.

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
```diff
@@ -45,73 +67,82 @@ contract RepPriceOracle is IRepPriceOracle, Initializable {
...
         if (_parentUniverse != IUniverse(0)) {
             IV2ReputationToken _parentReputationToken = _parentUniverse.getReputationToken();
-            _initialPrice = exchangeData[address(_parentReputationToken)].price;
+            return exchangeData[address(_parentReputationToken)].price;
         }
-        return _initialPrice;
+        return genesisInitialRepPriceinAttoCash;
     }
 
     function getOrCreateUniswapExchange(IV2ReputationToken _reputationToken) public returns (IUniswapV2) {
```
MicahZoltu: Consider restructuring this function for clarity:
```solidity
address _exchangeAddress = uniswapFactory.getExchange(cash, _reputationToken);
return (_exchangeAddress == 0)
    ? uniswapFactory.createExchange(cash, _reputationToken)
    : _exchangeAddress;
```

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
```diff
@@ -45,73 +67,82 @@ contract RepPriceOracle is IRepPriceOracle, Initializable {
...
         IUniverse _parentUniverse = _reputationToken.getUniverse().getParentUniverse();
-        uint256 _initialPrice = genesisInitialRepPriceinAttoCash;
         if (_parentUniverse != IUniverse(0)) {
             IV2ReputationToken _parentReputationToken = _parentUniverse.getReputationToken();
-            _initialPrice = exchangeData[address(_parentReputationToken)].price;
+            return exchangeData[address(_parentReputationToken)].price;
         }
-        return _initialPrice;
+        return genesisInitialRepPriceinAttoCash;
     }
```
MicahZoltu: Consider inverting the conditional into a guard clause:
```solidity
IUniverse _parentUniverse = _reputationToken.getUniverse().getParentUniverse();
if (_parentUniverse == IUniverse(0)) return genesisInitialRepPriceinAttoCash;
IV2ReputationToken _parentReputationToken = _parentUniverse.getReputationToken();
_initialPrice = exchangeData[address(_parentReputationToken)].price;
return exchangeData[address(_parentReputationToken)].price;

Comment (MicahZoltu) [reply packages/augur-core/source/contracts/reporting/RepPriceOracle.sol]: Reasoning: The "interesting" part of a function shouldn't live inside of a conditional.  Conditional contents should (ideally) contain the uninteresting or simplest path while the main path is the most interesting one.

File: packages/augur-core/source/contracts/reporting/RepPriceOracle.sol [hunk 8]
```diff
@@ -45,73 +67,82 @@ contract RepPriceOracle is IRepPriceOracle, Initializable {
...
     function initializeUniverse(IV2ReputationToken _reputationToken) private {
-        exchangeData[address(_reputationToken)].exchange = getOrCreateUniswapExchange(_reputationToken);
+        uint256 _blockNumber = block.number;
+        IUniswapV2 _exchange = getOrCreateUniswapExchange(_reputationToken);
+        exchangeData[address(_reputationToken)].exchange = _exchange;
+        if (_blockNumber != _exchange.blockNumberLast()) {
+            _exchange.sync();
+        }
         uint256 _initialPrice = getInitialPrice(_reputationToken);
         exchangeData[address(_reputationToken)].price = _initialPrice;
```
MicahZoltu: Ignoring the massively terrible idea that this upgradable exchange presents for the moment... calling `initializeUniverse` will reset the price back to its default state which I don't think is appropriate behavior during an exchange upgrade.  In general, I don't think we should be re-initializing at all in that scenario but instead should be just changing the exchange address we have on file and assuming that the previous exchange was reasonably priced.

MicahZoltu: I suppose we do need to reset the internal `repPriceAccumulated` value to the new exchange accumulator as well.

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
```diff
@@ -33,6 +43,18 @@ contract RepPriceOracle is IRepPriceOracle, Initializable {
...
+        require(!uniswapUpgraded, "Already upgraded");
+        IENSResolver _resolver = IENSResolver(ensRegistry.resolver(UNISWAP_REGISTRY_ENS_NAME));
+        require(_resolver != IENSResolver(0), "Resolver for ENS name not found");
+        address _newUniswapFactory = _resolver.addr(UNISWAP_REGISTRY_ENS_NAME);
+        require(_newUniswapFactory != address(0), "Could not get address from resolver");
+        uniswapFactory = IUniswapV2Factory(_newUniswapFactory);
+        uniswapUpgraded = true;
     }
 
     // TODO: Consider when this should be called other than when the price is requested as part of new fee setting
```
MicahZoltu: I _think_ that is the only time we need to call it if doing an exponential moving average?  For doing other types of moving averages it may need to be called on a regular interval (bot rewarded).

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
MicahZoltu: Hmm, I just noticed that you actually reset the price after calling this in the upgrade scenario... so maybe the rest of this is fine to reset...

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
```diff
@@ -45,73 +67,82 @@ contract RepPriceOracle is IRepPriceOracle, Initializable {
...
             return _exchangeData;
         }
 
+        if (uniswapUpgraded && !_exchangeData.upgradeHandled) {
+            uint256 _price = _exchangeData.price;
+            initializeUniverse(_reputationToken);
+            exchangeData[address(_reputationToken)].price = _price;
+            exchangeData[address(_reputationToken)].upgradeHandled = true;
+            return exchangeData[address(_reputationToken)];
+        }
```
MicahZoltu: Once again trying _really hard_ to ignore the upgradable exchange while reviewing...

It feels like this code should be extracted out into `pokeRepPriceInAttoCash` if possible.  That would allow us to mark this function as `view` I believe, which is valuable for auditing purposes (isolate view code from non-view code).

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
```diff
@@ -1,25 +1,35 @@
...
+    IENSRegistry public ensRegistry;
+    bool public uniswapUpgraded;
 
     struct ExchangeData {
         IUniswapV2 exchange;
-        uint256 cashAmountAccumulated;
-        uint256 repAmountAccumulated;
+        uint256 repPriceAccumulated;
         uint256 blockNumber;
+        uint256 blockTimestamp;
```
MicahZoltu: What is the purpose of storing the timestamp?  Just for convenience for integrating applications?  It doesn't appear to be read anywhere in this contract.

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
MicahZoltu: Actually, you probably _should_ call this once every `period`.

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
```diff
@@ -1,25 +1,35 @@
...
 contract RepPriceOracle is IRepPriceOracle, Initializable {
 
+    uint256 constant Q112 = 2**112;
+
+    bytes32 public constant UNISWAP_REGISTRY_ENS_NAME = 0xf259e1e59b6e1e9de21a18289a53332ebc255d2e4ece8a91e0e90c628a9c7f87; // "uniswapv2.eth"; // TODO: set this to the actual namehash of the owned ENS name
+
     IAugur public augur;
     address public cash;
-    uint256 public tau = 1 days / 15; // The time period (in blocks) of the twa, assuming a block is mined every 15 seconds. TODO: Review this value
+    uint256 public period = 3 days; // TODO: revisit if this is an appropriate period
```
MicahZoltu: 3 days seems too long for a simple moving average.  The longer it is, the more delayed the price feed will be and thus price changes won't be reflected quickly.  The shorter it is, the more at risk of manipulation attacks it is, though this is a function of market liquidity as well.

24 hours _feels_ good to me, but I don't have strong evidence to back that up.

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
```diff
@@ -45,73 +67,82 @@ contract RepPriceOracle is IRepPriceOracle, Initializable {
...
-    }
+        uint256 _secondsElapsed = _blockTimestamp - _exchangeData.blockTimestamp;
+        uint256 _priceAverage = _price;
 
-    function getWeight(uint256 _blockDelta) public view returns (uint256) {
-        uint256 _weight = _blockDelta * 1 ether / tau;
-        if (_weight > 1 ether) {
-            return 1 ether;
+        if (_secondsElapsed < period) {
+            _priceAverage = (_exchangeData.price * (period - _secondsElapsed) + _price * _secondsElapsed) / period;
```
MicahZoltu: Probably worth a comment indicating why safe math isn't necessary here (or anywhere else in this file).  Alternatively, use safe math.

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
```diff
@@ -45,73 +67,82 @@ contract RepPriceOracle is IRepPriceOracle, Initializable {
...
+        IUniswapV2 _exchange = getOrCreateUniswapExchange(_reputationToken);
+        exchangeData[address(_reputationToken)].exchange = _exchange;
+        if (_blockNumber != _exchange.blockNumberLast()) {
+            _exchange.sync();
+        }
         uint256 _initialPrice = getInitialPrice(_reputationToken);
         exchangeData[address(_reputationToken)].price = _initialPrice;
-        exchangeData[address(_reputationToken)].blockNumber = block.number;
+        exchangeData[address(_reputationToken)].blockNumber = _blockNumber;
+        exchangeData[address(_reputationToken)].blockTimestamp = block.timestamp; // solium-disable-line security/no-block-members
```
MicahZoltu: I'm curious why you don't just turn off that solium check?  I don't think it really adds value in this project since we use `block.timestamp` in a number of places, as well as `block.number`.

MicahZoltu (dismissed):
Overall the code looks good and I believe the math is correct (should definitely get more eyes on it though).

However, I believe I understand what is going on with the ENS stuff and it is basically introducing upgradability into this contract, which means upgradability into Augur.  Whoever controls the Uniswap ENS address has the ability to completely halt Augur which is (IMO) unacceptable and out of alignment with the Augur ethos.

The current attack vector for a malicious ENS owner would be to point the address at a contract that just reverts anytime it is called then trigger the permanent migration of Augur to that contract, and wait for the next reporting window where fee recalculations will fail, thus bringing down the whole system.

Even if you put in measures to address this (which would add a lot of code complexity) there are other attack vectors sitting behind that down to the one that cannot be fixed which is to provide a legitimate looking price feed that lies about the price, then setting REP target market cap to infinity or epsilon.

Recommended remediation:
Remove the Uniswap contract upgradability mechanism.  Either wait for Uniswap v2 to launch, or launch your own Uniswap look-alike contract that just has a REP:DAI pair (would not be hard to build, you can copy/paste Uniswap's codebase more or less and then delete a bunch of code).

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
nuevoalex: The point is is to allow it to change once after our initial deployment based on the uniswap team setting the ENS name

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
nuevoalex: the exchange.sync makes this not `view`-able

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
nuevoalex: its used to weigh the accumulated delta if the amount of time passed since last update is less than the period

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
nuevoalex: Unless we have something poking it when it isnt strictly needed we actually would need it to be 7 days. One idea is that Market creation _could_ poke this every time though that does add a fair bit of gas to it.

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
nuevoalex: We really only use block.timestamp in 3 places, but yea should probably just turn off that check.

MicahZoltu:
A note worth adding to that, this scenario could happen if the Uniswap contract you build against differs in the right way from the Uniswap contract that ultimately is pointed at by ENS.  If they, for example, make some last-minute change it could result in Augur getting bricked without anyone being a bad actor.

epheph:
I agree, ENS seems like too much control. Could there just be a one-time swap to a new address, followed by an ownership burn?

joeykrug:
> I agree, ENS seems like too much control. Could there just be a one-time swap to a new address, followed by an ownership burn?


I imagined it would just lookup the ens, if available, then swap to it, and not keep using ens on an ongoing basis? @nuevoalex

nuevoalex:
@joeykrug that’s exactly how it works. I believe the complaint from @epheph and @MicahZoltu is that an outside party will have control over the upgrade

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
```diff
@@ -1,25 +1,37 @@
...
+    using SafeMathUint256 for uint256;
+
+    uint256 constant Q112 = 2**112;
+
+    bytes32 public constant UNISWAP_REGISTRY_ENS_NAME = 0xf259e1e59b6e1e9de21a18289a53332ebc255d2e4ece8a91e0e90c628a9c7f87; // "uniswapv2.eth"; // TODO: set this to the actual namehash of the owned ENS name
 
     IAugur public augur;
     address public cash;
-    uint256 public tau = 1 days / 15; // The time period (in blocks) of the twa, assuming a block is mined every 15 seconds. TODO: Review this value
+    uint256 public period = 3 days; // TODO: revisit if this is an appropriate period
```
epheph: constant to avoid sload?

packages/augur-core/source/contracts/reporting/RepPriceOracle.sol
```diff
@@ -33,92 +45,113 @@ contract RepPriceOracle is IRepPriceOracle, Initializable {
...
     }
 
     // TODO: Consider when this should be called other than when the price is requested as part of new fee setting
     function pokeRepPriceInAttoCash(IV2ReputationToken _reputationToken) external returns (uint256) {
-        if (exchangeData[address(_reputationToken)].exchange == IUniswapV2(0)) {
+        if (exchangeData[address(_reputationToken)].exchange == IUniswapV2Exchange(0)) {
             initializeUniverse(_reputationToken);
         }
         ExchangeData memory _newExchangeData = calculateNewExchangeData(_reputationToken);
         exchangeData[address(_reputationToken)] = _newExchangeData;
```
epheph: I believe this might be extremely gas-inefficient. Because you are taking memory struct and assigning to storage, all 7 variables in the struct must be rewritten to storage, even if none of them require updates. Even in the case where some updates are required, it will rewrite these attributes each time: exchange, repIsToken0, upgradeHandled, even though they never or rarely change. If the calculate could be the one that actually performs the update, it could be more efficient.

Maybe not important since the cadence of calling this function is low?

packages/augur-core/source/contracts/BaseSimpleDex.sol
```diff
@@ -0,0 +1,182 @@
...
+
+    function publicBurn(address _to) external nonReentrant returns (uint256 _tokenAmount, uint256 _cashAmount) {
+        uint256 _liquidity = balances[address(this)];
+
+        _tokenAmount = _liquidity.mul(tokenReserve) / totalSupply;
+        _cashAmount = _liquidity.mul(cashReserve) / totalSupply;
+        require(_tokenAmount > 0 && _cashAmount > 0, "Insufficient liquidity");
+        transferToken(_to, _tokenAmount);
+        transferCash(_to, _cashAmount);
+        burn(address(this), _liquidity);
```
epheph: Burn before calling external contracts?

packages/augur-core/source/contracts/BaseSimpleDex.sol
```diff
@@ -0,0 +1,182 @@
...
+    function transferToken(address _to, uint256 _value) private;
+
+    function getTokenBalance() public returns (uint256);
+
+    function transferCash(address _to, uint _value) private {
+        cashTransfer(_to, _value);
+    }
+
+    function getCashBalance() public view returns (uint256) {
+        return cash.balanceOf(address(this));
```
MicahZoltu: `cashBalanceOf`?  Why wasn't this caught by the new tests that check for calling `cash` directly?

packages/augur-core/source/contracts/BaseSimpleDex.sol
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    function initializeInternal(address _augurAddress, address _token) internal {
```
MicahZoltu: Style Nitpick: Put initializer above other functions.

packages/augur-core/source/contracts/BaseSimpleDex.sol
```diff
@@ -0,0 +1,182 @@
...
+        uint256 _tokenReserve = tokenReserve;
+        if (_tokenReserve != 0 && _cashReserve != 0 && _blocksElapsed > 0) {
+            // cannot reasonably overflow unless the supply of Cash, REP, or ETH became several OOM larger
+            tokenPriceCumulativeLast += _cashReserve.mul(10**18).mul(_blocksElapsed).div(_tokenReserve);
+        }
+        tokenReserve = _tokenBalance;
+        cashReserve = _cashBalance;
+        blockNumberLast = _blockNumber;
+        emit Sync(_tokenBalance, _cashBalance);
+    }
```
MicahZoltu: Since we aren't trying to build a general purpose price feed, I think we can simplify things a lot and remove the RepPriceOracle contract by just inlining the moving average calculation to here.  Uniswap did it this way so they didn't have to be opinionated on the specifics of the moving average that third parties would use, but in our case we want a single specific moving average calculation.  I think this would be a positive change as it would simplify some code, and I think remove some storage variables (not confident on the latter).

packages/augur-core/source/contracts/BaseSimpleDex.sol
```diff
@@ -0,0 +1,182 @@
...
+        uint256 _tokenBalance = getTokenBalance();
+        uint256 _cashBalance = getCashBalance();
+        uint256 _tokenReserve = tokenReserve;
+        uint256 _cashReserve = cashReserve;
+        uint256 _tokenAmount = _tokenBalance.sub(_tokenReserve);
+        uint256 _cashAmount = _cashBalance.sub(_cashReserve);
+
+        _liquidity = totalSupply == 0 ?
+            _tokenAmount.mul(_cashAmount).sqrt() :
+            (_tokenAmount.mul(totalSupply) / _tokenReserve).min(_cashAmount.mul(totalSupply) / _cashReserve);
```
MicahZoltu: Make `totalSupply` a local variable since it is referenced 3 times in the common code path.

packages/augur-core/source/contracts/BaseSimpleDex.sol
```diff
@@ -0,0 +1,182 @@
...
+    function buyToken(address _recipient) public nonReentrant returns (uint256 _tokenAmount) {
+        uint256 _cashBalance = getCashBalance();
+        uint256 _cashAmount = _cashBalance.sub(cashReserve);
+
+        _tokenAmount = getCashSaleProceeds(_cashAmount);
+        require(_tokenAmount > 0, "tokenAmount must be > 0");
+        transferToken(_recipient, _tokenAmount);
+
+        update(getTokenBalance(), _cashBalance);
+        emit Swap(msg.sender, address(cash), _cashAmount, _tokenAmount, _recipient);
```
MicahZoltu, line 172: During Maker global shutdown, the swap may be between vDAI and the token, rather than between DAI (cash) and the token.  In such a scenario, this event will produce incorrect results.

packages/augur-core/source/contracts/BaseSimpleDex.sol
```diff
@@ -0,0 +1,182 @@
...
+    IERC20 public cash;
+
+    uint256 public tokenReserve;
+    uint256 public cashReserve;
+    uint256 public blockNumberLast;
+    uint256 public tokenPriceCumulativeLast;
+
+    event Mint(address indexed sender, uint256 tokenAmount, uint256 cashAmount);
+    event Burn(address indexed sender, uint256 tokenAmount, uint256 cashAmount, address indexed to);
+    event Swap(address indexed sender, address indexed tokenIn, uint256 amountIn, uint256 amountOut, address indexed to);
```
MicahZoltu, line 29: Since we only need to support REP:DAI and ETH:DAI, consider changing this event signature to have a short-string for `tokenIn` rather than an address.  This will make it a bit easier to read in various block explorers, and also address the DAI/vDAI issue.

packages/augur-core/source/contracts/BaseSimpleDex.sol
```diff
@@ -0,0 +1,182 @@
...
+
+contract BaseSimpleDex is Initializable, ReentrancyGuard, VariableSupplyToken, CashSender {
+    using SafeMathUint256  for uint256;
+
+    string constant public name = "Simple Decentralized Exchange";
+    string constant public symbol = "SDE";
+
+    IAugur public augur;
+    address public token;
+    IERC20 public cash;
```
MicahZoltu: Consider removing this variable.  We should never be referencing it directly (always through `CashSender`) and as seen in other comments on this PR its presence tempts one to actually use it.

packages/augur-core/source/contracts/BaseSimpleDex.sol
nuevoalex: because it just got merged to master ;p

packages/augur-core/source/contracts/BaseSimpleDex.sol
nuevoalex: because of what was noted above concerning not using `cashBalance` correct?

MicahZoltu: In short, yes.  This function may transfer DAI or it may transfer vDAI, depending on the state of Maker.

packages/augur-core/source/contracts/BaseSimpleDex.sol
MicahZoltu: I think the math may be easier to understand here if you reverse the order from
```
tokenAmount * totalSupply / tokenReserve
```
to
```
totalSupply * tokenAmount / tokenReserve
```
It helps make it more clear that you are multiplying `totalSupply` by the percentage of tokens you are adding relative to the tokens that existed previously.

packages/augur-core/source/contracts/BaseSimpleDex.sol
```diff
@@ -0,0 +1,182 @@
...
+
+        update(_tokenBalance, _cashBalance);
+        emit Mint(msg.sender, _tokenAmount, _cashAmount);
+    }
+
+    function publicBurn(address _to) external nonReentrant returns (uint256 _tokenAmount, uint256 _cashAmount) {
+        uint256 _liquidity = balances[address(this)];
+
+        _tokenAmount = _liquidity.mul(tokenReserve) / totalSupply;
+        _cashAmount = _liquidity.mul(cashReserve) / totalSupply;
```
MicahZoltu: Same comment as above, consider swapping the order of the variables to make it more clear that you are multiplying the reserve by the user's share of total supply.

MicahZoltu: Consider extracting `totalSupply` into a local variable since you reference it twice.

packages/augur-core/source/contracts/BaseSimpleDex.sol
```diff
@@ -0,0 +1,182 @@
...
+            // cannot reasonably overflow unless the supply of Cash, REP, or ETH became several OOM larger
+            tokenPriceCumulativeLast += _cashReserve.mul(10**18).mul(_blocksElapsed).div(_tokenReserve);
+        }
+        tokenReserve = _tokenBalance;
+        cashReserve = _cashBalance;
+        blockNumberLast = _blockNumber;
+        emit Sync(_tokenBalance, _cashBalance);
+    }
+
+    function publicMint(address _to) external nonReentrant returns (uint256 _liquidity) {
```
MicahZoltu: The way Uniswap v2 is implemented, there are two contracts.  One is the exchange contract which offers only low-level access and is basically designed for contract interaction, not end-user interaction.  The other is a helper contract that makes the API actually usable by end users.  It doesn't appear that `publicMint` or `publicBurn` are called anywhere else in Augur at the moment, and without that wrapper contract it makes the usability of this contract really bad.

I recommend either writing your own wrapper or cribbing the Uniswap wrapper.

MicahZoltu: As an example of one of the many usability issues, you can only burn _all_ of your tokens with `publicBurn`.  If you use a wrapper, then presumably you would transfer your tokens to the wrapper first, and then the wrapper would burn everything it had.

MicahZoltu: Same general problem with `sellToken` and `buyToken`.

nuevoalex: sellToken and buyToken have convenience functions `autoSellToken` and `autoBuyToken` if thats what you mean

packages/augur-core/source/contracts/Augur.sol
```diff
@@ -113,7 +112,7 @@ contract Augur is IAugur, IAugurCreationDataGetter, CashSender {
         require(registry[_key] == address(0), "Augur.registerContract: key has already been used in registry");
         require(_address.exists());
         registry[_key] = _address;
-        if (_key == "ShareToken" || _key == "MarketFactory") {
+        if (_key == "ShareToken" || _key == "MarketFactory" || _key == "EthExchange") {
```
epheph, line 115: I understand why this line is here, but this is also the first place where Orders code/trust (ZeroX orders in this case) bleeds into the rest of Augur. Before this, it was really kind of bolted onto the side

packages/augur-core/source/contracts/BaseSimpleDex.sol
epheph: right, autoBuy|Sell does the right thing, but there needs to be an auto-version of providing and removing liquidity?

packages/augur-core/source/contracts/Augur.sol
nuevoalex: Yea its unfortunate. We _could_ avoid this by having yet another approval for the EthExchange though I'd rather not if possible. The reason I think this is alright is that while only the trading code _looks_ at the ETH exchange its not strictly a part of trading. Really it's its own thing.

packages/augur-core/source/contracts/BaseSimpleDex.sol
nuevoalex: I'll consider adding those in another PR

nuevoalex: After talking it out with @epheph will go ahead and add these now

packages/augur-core/source/contracts/BaseSimpleDex.sol
```diff
@@ -0,0 +1,192 @@
...
+        uint256 _cashBalance = getCashBalance();
+        uint256 _tokenReserve = tokenReserve;
+        uint256 _cashReserve = cashReserve;
+        uint256 _totalSupply = totalSupply;
+        uint256 _tokenAmount = _tokenBalance.sub(_tokenReserve);
+        uint256 _cashAmount = _cashBalance.sub(_cashReserve);
+
+        _liquidity = _totalSupply == 0 ?
+            _tokenAmount.mul(_cashAmount).sqrt() :
+            (_tokenAmount.mul(_totalSupply) / _tokenReserve).min(_cashAmount.mul(_totalSupply) / _cashReserve);
```
MicahZoltu, line 84: Uniswap has to support tokens with arbitrary scale, which is why they set liquidity to this rather complicated value (note the square root).  Since we only have to deal with tokens that are of the scale 10^18, and you can multiply 10^18 by itself about 4 times before you are in danger of running out of space in a uint256, we can instead have liquidity just be `k` (meaning `token_a * token_b`).  I believe such a change will simplify some calculations in a few places, including this one.

epheph: From a usability standpoint, having the initial liquidity value be so large is likely to be awkward for things that treat it as an ERC20. Depositing 1 ETH and 160 DAI will yield 160e18 tokens. I think the sqrt is just a usability thing?

packages/augur-core/source/contracts/trading/ZeroXTrade.sol
```diff
@@ -258,6 +262,10 @@ contract ZeroXTrade is Initializable, IZeroXTrade, IERC1155, CashSender {
             uint256 _amountTraded = doTrade(_order, totalFillResults.takerAssetFilledAmount, _fingerprint, _tradeGroupId, msg.sender);
 
             _fillAmountRemaining = _fillAmountRemaining.sub(_amountTraded);
+            _maxTrades -= 1;
```
epheph, line 265: underflow here, but they can only hurt themselves.

epheph (approved):
lgtm, I have concerns about publicMintAuto that I will file in an issue. I'm hoping it will be a wontfix due to this implementation being replaced by uni v2

Issue #4564, opened 2019-11-03
Consolidate bulk sync calls to reduce strain on eth node


justinbarry:

joeykrug:
prob good to do this

justinbarry:
This work has been done on the warp sync branch.

PR #5381, opened 2020-01-04
Warp Sync

justinbarry:
# Proposal

In order to power fast UI starts with histortical data, and to combat the slow removal of logs from the Ethereum full nodes, we need a method of generating verifiable datasets that can be used to bootstrap the UI databases. This feature is called warp sync.

This work comes into two halves (1) Contract work which is already completed, (2) SDK work needed to generate the dataset and the necessary hashes.

## Description of Contracts

The contract portion of the work assumes that from histortical augur data a consistent dataset can be generated by syncing the logs up to a certain block. This data set can then have some sort of hashing function applied to it which outputs a real number, (0, 2^256-1).

Each day a market is created automatically by the contracts, which allows people to report on the hash as of a certain block. Users can report the value of the above hash in order to claim the initial reporter pot.

This market can then be referenced directly via contract calls to return the last valid hash, allowing the UI to do one contract call to ask for the valid hash, and then fetch the warp sync data via that hash and bootstrap its database.

## Goals for User Experience

In order to power a good user experience we want to be able to provide two variants of warp sync files without changing the above contract constraints. We would like users to be able to quickly download the entire warpsync dataset OR the data from a specific checkpoint (range of blocks). This will allow clients that have synced recently (for example 36 hours ago) to fetch only the data they need isntead of re-downloading a large file.

Syncing also needs to be fast and not rely solely upon centralized parties -- although centralization may be introduced as a speed or reliability improvement, it is technically not required.

In order to have the best user experience, warp sync data (from IPFS) will need to be combined with syncing from Eth nodes in such a way that the user is given the most relevant information first. The structures described below are intended to support the various use-cases where the UI may want to load data quickly for a certain view. it is To Be Discovered whether or not these actually work well in those cases.

### Overview

The solution to the above set of problems is to utilize IPFS to create a content-addressed payload, where the HASH part of the IPFS Content ID (CID) is used to report on the daily recurring market.

For each day there is a range of blocks that will produce a valid warp sync file. Currently the idea is to take each 5240 blocks, fetch all the logs, and then to serialize them into a text file in a consistent format. This file is referred to as a checkpoint, and can be itself put into IPFS.

In order to organize all the checkpoints, as well as provide a single warp sync file that encompases all the data, we will need to take advantage of two IPFS features: Directories, and the Composability of files by treating them as normal DAG nodes.

#### Terms
- Warppoint  # The top level hash of all warp sync data, agreed upon by the Augur oracle
- Checkpoint # A block number specifying when a snapshot was taken, represents the data between the Checkpoint block number and the previous checkpoint block number.

#### Definition
```
- ${warppoint}/                             # ROOT hash for each warp sync there will be a unique hash
  - version                                 # Format Version, clients can use this to determine whether or not they're compatiable with the warp sync format.
  - range                                   # The range of blocks this checkpoint represents
  - index                                   # A serialized form of all content suitable for a new client getting all information
  - checkpoints/                            # A directory holding historical checkpoints, this can be capped at N checkpoints.
    - ${checkpoint_blocknumber}             # A directory holding data for a checkpoint as of a blocknumber
      - index                               # All checkpoint data in a serialized form consumable for clients that just need that checkpoint
  - tables                                  # A directory containing serialized forms of each table
    - CompleteSetsPurchased
    - CompleteSetsSold
    - DisputeCrowdsourcerCompleted
    - ... MORE
    - TransferBatch
  - market/                               # A directory containing market rollups
    - volume                              # All market volumes
    - oi                                  # All market OIs
    - ${market_id}/                       
      - orders                            # Orders filtered by a market
  - token/                                # A directory containing token rollups
    - balance                             # Balances for all owner,token
  - share_token/                          # A directory containing share token rollups
    - balance                             # Balances for all [account, marketid, outcome]
  - account/${account_id}                 # Indexed lookups for an account
    - orders
    - pnl
    - market/${market_id}                 # indexed lookups for a market within an account
      - orders
      - pnl
```

### Serialization Strategy
 
To power the above we'll need two different serialzation strategies for the data. It is important to note that in the above format, a lot of the flexibility comes from the ability of IPFS to store a chunk of data, and then store pointers to that data (represented by the CID of the data being stored in a Link) in other DAG nodes. This design allows us to store one chunk of data, but refer to it in multiple contexts.

For example, say we were injesting Order events, and wanted to be able access any individual days worth of Orders as well as being able to fetch orders by their month.

{{ Insert Example Here }}

##### Per-Item Serialization

In order to be able to construct interesting sub-indices of the data, the first pass for data storage will be to store each serialized log / database row as its own IPFS file. This has the advantage of allowing these various pieces of data to be composed in new and interesting ways, for example the same piece of data can be used to represent "All orders made today" as "All orders made by user X this week." However, it does add some IPFS metadata overhead since each object is hashed and stored separately, and any files that need to include that object must include a Link to that particular CID.

*Begin with per-item serialization everywhere and then we'll move to bulk serialization as an optimization*

##### Bulk Serialization

Bulk serialization mode is optimal for serializing the log data that won't be indexed in novel ways. For instance, all Market Created and Market Finalized logs just need to be retrieved, but we don't point to a specific single Market Creation log from any of our other IPFS directories. In this case, it is more efficient to store more logs per item serialized to IPFS as this reduces the amount of metadata that needs to be stored per log by a small amount.

So for instance, when bulk serializing logs, we may take 50 serialized logs and concatenate them into one chunk, and add them to IPFS as a group. This means that when re-constructing the file, the IPFS clients only need to resolve on IPFS CID for this whole batch of 50 logs, instead of one CID per log.

### Loading data from market page
```
- if viewing market then
  - first load on contract data using the market hot load helper
  - at the same time / next -- do http calls to the latest warp point, to specific user endpoints for orders and pnl, also filtered by market. Currently I am thinking this tree.
  /Ipfs/warppoint/rollup/user/uid/market/mid/(pnl|orders)
  - next, create the syncabledbs for orders and pnl with the additional filter for the user, for the range since the last warp point to now (on average 1 day). follow this same process for all this user’s data after this is done for the market the user is viewing. Then do it for markets data for discovery
  - data will be loaded like a normal sync and queryable via getters
  - during full sync (next) data will be reloaded and overwritten without causing harm
 - then begin the full sync of everything else (warp + info since last checkpoint)
^ one possible variant
May make sense to do markets data for discovery then all user data for that user
Do we have markets sorted by active or not in the db?
```

### Checkpoint Generation
- [x] Serializes all needed tables
- [x] Filter out tables not needed for warp sync
- [x] Serializes indices into ipfs structure (from spec)
- [x] Generates a checkpoint on the last block of the day
- [x] Always generates and pins the last days checkpoint

### Ipfs replicating
- [x] Works when running as server
- [x] Works for client running browser

### Expansion
- [x] Transition between multiple sync "strategies" on load
- [x] Individual Market loading (Along with market specific use data)
- [x] Load user data from warp sync index (if necessary, e.g. its a new sync for this user)

- [x] Load "diff" from most recently loaded checkpoint to current hash
- [x] For both warp and present day eth node data. Note that for warp we should only load data since users last use of the UI.
- [x] 1) Load warp data + data from eth node for the specific market a user is looking at first
- [x] 2) Then load warp data + data from eth node that user’s specific data in general
- [x] 3) then load warp market data + data from eth node
- [x] 4) then load any other warp data + data from eth node
Note only the methods to do this are done. The client needs to be modified to provide the relevant account/market.

### Left to do
- [x] Trigger downloads for 1 and 2 at the same time, then 3 and 4 sequentially
- [x] Pass Account/Market Address to sdk on app startup.
- [ ] Offer to report hash to market.
- [ ] Identify current checkpoint hash
    - [ ] Via market
    - [ ] Via UI?

Proposal copied from https://gist.github.com/pgebheim/3f2faae37f29f8e22d4bf4d5a0c6dbab


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). 🏅

**https://peek.run/l74h5qm**

-------

justinbarry:
closes #4564

packages/augur-sdk/src/state/Controller.ts
```diff
@@ -4,41 +4,29 @@ import * as fp from 'lodash/fp';
...
-      );
-      this.blockAndLogStreamerListener.startBlockStreamListener();
-    } catch (err) {
-      console.log(err);
-    }
+    this.logFilterAggregator.listenForAllEvents(this.updateMarketsData);
+    this.logFilterAggregator.notifyNewBlockAfterLogsProcess(this.notifyNewBlockEvent.bind(this));
+    // this.logFilterAggregator.listenForBlockRemoved(
+    //   db.rollback.bind(db)
+    // );
```
robert-e-davidson3: TODO

packages/augur-sdk/src/state/Server.ts
```diff
@@ -1,13 +1,16 @@
...
 import { EndpointSettings } from "./getter/types";
 import { EventEmitter } from "events";
 
+configureDexieForNode(false);
+
 export async function run() {
   const settings = require("@augurproject/sdk/src/state/settings.json");
 
-  const api = await Sync.start(settings.ethNodeURLs[4], settings.testAccounts[0], false);
+  const api = await Sync.start(settings.ethNodeURLs[42], settings.testAccounts[0], false);
```
robert-e-davidson3: Should be pulled from config somewhere.

justinbarry:
Thoughts on finding final block of the day not covered by checkpoint
1. Traverse back to last know day boundary. If none found start at contract upload block number as startingBlock
2. Use binary search tree to determine the boundary of the following day
3. Start at block from step one and add 5760. Grab the block header and check if still current day.
4. If same day, make note of current block as "lastPreviousDay", advance 20 blocks and continue to 6.
5. If following day, make note of current block as "firstNextDay", retreat 20 blocks and continue to 6.
6. On every step calculate distance between block number of lastPreviousDay and firstNextDay.
7. If difference is exactly 1 return tuple [lastPreviousDay, firstNextDay].
8. If greater, repeat with the new values.

pgebheim (approved):

PR #5382, opened 2020-01-04
Removes tab character from a bunch of templates.

MicahZoltu:


bthaile:
Also need to regen templates. List values are included in the validations. I can do that on this branch just to make sure

MicahZoltu:
I don't know how to do that, so feel free to update this branch with the necessary stuff or create a new PR.  I just figured it would be easier to submit a PR with the problem rather than trying to describe it in an issue.  :)

bthaile (approved):
lgtm, we'll need to re-generate templates in another PR

PR #5383, opened 2020-01-04
zeroX open orders frozen funds

bthaile:
collect potential frozen funds in dai for open orders, use user's position to determine if open orders will use shares or dai.

when open orders change based on events consumed in log-handlers then UI will call new userOpenOrders getter to get logged in user's open orders with frozen funds calculation. 

open orders frozen funds is added to position frozen funds and displayed in Account Summary overview and in top bar. 

Screen Shot 2020-01-04 at 9 42 40 PM

### DIFF
--- a/packages/augur-sdk/src/state/getter/Users.ts
+++ b/packages/augur-sdk/src/state/getter/Users.ts
@@ -30,13 +30,15 @@ import { formatBytes32String } from 'ethers/utils';
 import * as _ from 'lodash';
 import * as t from 'io-ts';
 import { QUINTILLION } from '../../utils';
-import { OnChainTrading, MarketTradingHistory, Orders } from './OnChainTrading';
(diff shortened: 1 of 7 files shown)


packages/augur-sdk/src/state/getter/Users.ts
```diff
@@ -1175,6 +1095,52 @@ function getLastDocBeforeTimestamp<TDoc extends Timestamped>(
...
+      new BigNumber(order.amount),
+      minOutcomeShareAmount
+    );
+    cost = new BigNumber(order.amount)
+      .minus(useShares)
+      .times(displayMaxPrice.minus(new BigNumber(order.price)));
+    Object.keys(userSharesBalances).forEach(o => {
+        if (outcome !== o) {
+          userSharesBalances[o] = new BigNumber(userSharesBalances[o])
+                  .minus(minOutcomeShareAmount)
```
joeykrug: this should be useShares I believe

joeykrug (approved):

PR #5387, opened 2020-01-05
Minor tweaks to the movie template copy.

MicahZoltu:
Why are these for US only?

### DIFF
--- a/packages/augur-tools/src/templates-source.ts
+++ b/packages/augur-tools/src/templates-source.ts
@@ -4954,8 +4954,8 @@ export const TEMPLATES = {
       },
       {
         marketType: YES_NO,
-        question: `Will [0] gross $[1] USD or more, in it's opening weekend in the US?`,
-        example: `Will Avangers: Endgame gross $350 million USD or more in it's opening weekend in the US?`,
+        question: `Will [0] gross $[1] USD or more over its opening weekend in the US?`,
+        example: `Will Avangers: Endgame gross $350 million USD or more over its opening weekend in the US?`,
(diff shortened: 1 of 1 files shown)


packages/augur-tools/src/templates-source.ts
```diff
@@ -4971,15 +4971,20 @@ export const TEMPLATES = {
         resolutionRules: {
           [REQUIRED]: [
             {
-              text: `Gross total should include the first Friday through Sunday of the movie release and does not include extra days due to holidays`,
+              text:
+                'Gross total should include sales from the first Friday (local showing time) through the first Sunday (local showing time) of the movie release',
```
bthaile, line 4975: not sure about `(local showing time)` usually the websites that collect the gross movie sales data don't have local showing times. It would be confusing to add.

MicahZoltu: We should figure out how existing sites calculate the metric and include that then.  I assumed it was local showing time, since I believe a Thursday midnight showing on the East Coast counts as a "weekend" showing for the sake of box office numbers, while a showing at _the exact same time_ on the West Coast would not (since that would be 21:00 PST).

I believe this is even more extreme for worldwide numbers.

These templates need to have clear resolution rules, whatever they are, so that the community can decide on what is correct when there is disagreement across multiple reporting sites.

bthaile:
can't really use UTC since movie sales occur across multiple timezones. like in the united states. it all depends on how gross sales is calculated by the data aggregator website/organization. I think it's find being vague.

PR #5388, opened 2020-01-05
Typo fix in templates.

MicahZoltu:


bthaile (approved):
lgtm, we'll need to re-generate the templates which will occur with other PR.

Issue #5386, opened 2020-01-05
Provide consistency in financial exchange list.


MicahZoltu:
https://github.com/AugurProject/augur/blob/d66819804c4db26fe0f6f16014131aeb41a6e28c/packages/augur-tools/src/templates-lists.ts#L7917-L7956

The units for these is inconsistent.  "traded in USD" vs "traded in Canadian Dollar" vs "traded in HKD Hong Kong Dollar" vs "traded in Japanese Yen" vs "traded in Euro".  Recommend just using currency ticker for each, as I _believe_ those are unambiguous, well defined, and uniform.  The Shanghai Stock Exchange ones are a bit bizarre as I'm not familiar with A shares vs B shares, but we should consistently use currency ticker for each.

It is worth noting that most of these would be invalid according to the guidelines in the v2 UI:
> If it’s a stock, currency or cryptocurrency and its ticker is not used in the market question.

Recommended values:
```javascript
    {
      label: 'New York Stock Exchange (traded in USD)',
      value: 'New York Stock Exchange (traded in USD)',
    },
    {
      label: 'Nasdaq (traded in USD)',
      value: 'Nasdaq (traded in USD)',
    },
    {
      label: 'American Stock Exchange (traded in USD)',
      value: 'American Stock Exchange (traded in USD)',
    },
    {
      label: 'Tokyo Stock Exchange (traded in JPY)',
      value: 'Tokyo Stock Exchange (traded in JPY)',
    },
    {
      label: 'London Stock Exchange (traded in GBP)',
      value: 'London Stock Exchange (traded in GBP)',
    },
    {
      label: 'Shanghai Stock Exchange (A shares traded in CNY)',
      value: 'Shanghai Stock Exchange (A shares traded in CNY)',
    },
    {
      label: 'Shanghai Stock Exchange (B shares traded in USD)',
      value: 'Shanghai Stock Exchange (B shares traded in USD)',
    },
    {
      label: 'Hong Kong Stock Exchange (traded in HKD)',
      value: 'Hong Kong Stock Exchange (traded in HKD)',
    },
    {
      label: 'Toronto Stock Exchange (traded in CAD)',
      value: 'Toronto Stock Exchange (traded in CAD)',
    },
    {
      label: 'Euronext NV (traded in EUR)',
      value: 'Euronext NV (traded in EUR)',
    },
```

PR #5397, opened 2020-01-06
Use currency ticker in market templates.

jfarid27:
## Description

Removes extra description in specified currencies for market templates.

## Issue
#5386


bthaile (approved):
currency names look good. I'll re-generate templates in different PR

PR #5403, opened 2020-01-06
Updated market rules for invalid markets

### LINKED ISSUE #5267: Update Market Rules
Here are the updated rules to be used in Market creation, for both template and non template markets, as well is Reporting and disputing pgs as well

Market rules

A market is invalid if: 

•	The market question, resolution details or its outcomes are ambiguous, subjective or unknown.
•	The result of the event was known at market creation time.
•	The outcome was not known at event expiration time.
•	It can resolve without at least one of the outcomes listed being the winner, unless it is explicitly stated how the market will otherwise resolve in the resolution details. 
•	The title, details and outcomes are in direct conflict with each other.
•	The market can resolve with more than one winning outcome
•	Any of the outcomes don’t answer the market question ONLY. (outcomes cannot introduce a secondary question)
•	If using a resolution source (a source is a noun that reports on or decides the result of a market), the source's URL or full name is NOT in the Market Question, regardless of it being in the resolution details.
•	If using a resolution source, it is not referenced consistently between the Market Question and Resolution Details e.g. as either a URL or its full name
•	For any sports markets that list a player or team not in the correct league, division or conference, at the time the market was created, the market should resolve as invalid.
•	A market only covers events that occur after market creation time and on or before reporting start time. If the event occurs outside of these bounds it has a high probability as resolving as invalid.

bconfortin:
Closes #5267


bthaile (approved):
lgtm

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). ⚡️

**https://peek.run/kq3h4qr**

-------

Issue #5234, opened 2019-12-13
Migrate REP button and tooltip 


cillianhunter0:
Desktop: 
https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=12885%3A588
- [ ] Change the button type (blue button is too big for the new nav bar) **HIGH**
- [ ] Switch out the tooltip icon (new icon being used everywhere) **LOW**

![image.png](https://images.zenhubusercontent.com/5c7591d7eec15745ffae99b1/0bdab0f1-8fa0-43ae-9533-4078b8a88566)

Mobile: 
https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=12888%3A934
- [ ] Change the button type (blue button is too big for the new nav bar) **HIGH**
- [ ] Add the tooltip **HIGH**

![image.png](https://images.zenhubusercontent.com/5c7591d7eec15745ffae99b1/fce5cf7c-0949-4660-9e38-13780e918d1f)

jfarid27:
So I pushed up #5411, but I would recommend punting on adding the new styling for the tooltips icons. It's definitely not a priority.

joeykrug:
tooltip in build looks better anyway imo

cillianhunter0:
Closing as not priority.

PR #5411, opened 2020-01-07
Migrate REP button to new styling in side-nav and top-nav

jfarid27:
## Description

Adds new button styling to the Migrate REP button.

## Issue

AugurProject/augur#5234

## Testing

- Make sure if you're trying this out to change the `showMigrateRepButton` to truthy in the files. I'm not sure how else to set up the state where you need to migrate in a local dev mode.

### Top-Nav
Note here the condition where you need to add REP and migrate it shouldn't ever be reached. I just tried to get a screengrab of it working so you could technically ignore the "Add Funds" button and spacing at the moment.
Screenshot from 2020-01-06 20-26-47

### Side-Nav
Screenshot from 2020-01-06 21-30-53

### DIFF
--- a/packages/augur-ui/src/modules/app/components/side-nav/side-nav.styles.less
+++ b/packages/augur-ui/src/modules/app/components/side-nav/side-nav.styles.less
@@ -49,6 +49,34 @@
   }
 }
 
+.SideNavMigrateRep {
+  button {
+    .mono-10-bold;
+    text-transform: uppercase;
+  }
+}
+
+.SideNavMigrateTooltipHint {
+  display: inline;
+  position: relative;
+  top: 0.2rem;
+  left: 0.3rem;
+
+  > svg {
+    max-height: @size-16;
+    max-width: @size-16;
+
+    > circle {
+      fill: @color-secondary-action;
+      stroke: @color-secondary-action-outline;
+    }
+
+    > path {
+      fill: @color-primary-text;
+    }
+  }
+}
+
 .SideNav__container {
   background: @color-module-background;
   position: absolute;
--- a/packages/augur-ui/src/modules/app/components/side-nav/side-nav.tsx
+++ b/packages/augur-ui/src/modules/app/components/side-nav/side-nav.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import ReactTooltip from 'react-tooltip';
 import { Link } from 'react-router-dom';
 import classNames from 'classnames';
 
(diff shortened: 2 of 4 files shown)


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). 🎖

**https://peek.run/kq3h4nr**

-------

packages/augur-ui/src/modules/app/components/side-nav/side-nav.styles.less
```diff
@@ -49,6 +49,34 @@
...
+    > circle {
+      fill: @color-secondary-action;
+      stroke: @color-secondary-action-outline;
+    }
+
+    > path {
+      fill: @color-primary-text;
+    }
+  }
+}
```
jfarid27, line 78: I'm also open to discussing this change. These are copies of the TooltipHint in the common folder, but I couldn't figure out how to add an "&" nested parent selector since the webpack strategy garbles the CSS output. I think it's ok to just keep it and optimize for code re-use later to get this out the door, but if anyone has a better way to do this I'm all ears.

packages/augur-ui/src/modules/app/components/side-nav/side-nav.tsx
```diff
@@ -78,10 +80,39 @@ const SideNav = ({
...
+                    type="light"
+                  >
+                    <p>
+                      {
+                        'You have V1 REP in your wallet. Migrate it to V2 REP to use it in Augur V2'
+                      }
+                    </p>
+                  </ReactTooltip>
+                </span>
+              )}
```
JohnDanz, line 110: I think putting both the button and label in the same `<span>` since they have the same `showMigrateRepButton` requirement might be cleaner. This would have an added benefit of being able to use flex to align these things instead of relative positioning the icon/label

JohnDanz (changes requested):
Minor change request, thanks for the PR @jfarid27

packages/augur-ui/src/modules/app/components/side-nav/side-nav.tsx
jfarid27: Fixed!

jfarid27: Sorry about the close then reopen. Extra spans have been removed.

Screenshot from 2020-01-10 11-47-22

JohnDanz (approved):

PR #5412, opened 2020-01-07
fixed some notifications / order event stuff

joeykrug:


andrewdamelio (approved):
nice!

packages/augur-sdk/src/api/ZeroX.ts
```diff
@@ -329,10 +329,15 @@ export class ZeroX {
...
-      return this.meshClient.addOrdersAsync([order]);
+    try {
+      if (this.browserMesh) {
+        return this.browserMesh.addOrdersAsync([order]);
+      } else {
+        return this.meshClient.addOrdersAsync([order]);
+      }
+    }
+    catch(error) {
+      return setTimeout(this.addOrder(order), 5000);
```
andrewdamelio, line 340: 👍

andrewdamelio (approved):
lgtm

packages/augur-sdk/src/api/ZeroX.ts
pgebheim: This needs to actually detect the error and probably do some sort of back-off with a limit or we can get into a terrible busy loop here.

PR #5416, opened 2020-01-07
Adds .editorconfig for augur-core.

MicahZoltu:
This will supercede the repo-wide .editorconfig that sets `indent_size` for `.ts` files to `2`.

The alternative solution would be to reformat all TS files in core, which seems like the wrong solution for me to undertake.


nuevoalex:
Done in https://github.com/AugurProject/augur/pull/5502

PR #5424, opened 2020-01-07
Fixed the user rep box overlapping container

bconfortin:
This fixes the user rep box when the user is not logged in. Currently it's overlapping the container.

Current
Captura de Tela 2020-01-07 às 12 00 00

Fixed
Captura de Tela 2020-01-07 às 11 59 54


bthaile (approved):
Lgtm

PR #5432, opened 2020-01-07
0x multi asset

nuevoalex:


packages/augur-sdk/src/state/db/ZeroXOrders.ts
```diff
@@ -200,6 +210,20 @@ export class ZeroXOrders extends AbstractTable {
       this.augur.events.emit('OrderEvent', {eventType: OrderEventType.Cancel, ...storedOrder});
       return false;
     };
+
+    const multiAssetData = defaultAbiCoder.decode(multiAssetDataAbi, `0x${storedOrder.signedOrder.makerAssetData.slice(10)}`);
+    const amounts = multiAssetData[0] as BigNumber[];
+    if (!amounts[0].eq(1)) return false;
+    if (!amounts[1].eq(0)) return false;
+    if (!amounts[2].eq(0)) return false;
```
epheph, line 223: Check amounts length to avoid calling `.eq` on `undefined`?

epheph: Actually, check amounts length to avoid extra multi transfers in there, like another multi-transfer that targets dai but with actual value?

packages/augur-core/source/contracts/trading/ZeroXTrade.sol
```diff
@@ -367,16 +450,66 @@ contract ZeroXTrade is Initializable, IZeroXTrade, IERC1155 {
...
+        // Slice the selector off the asset data
+        bytes memory _noSelectorAssetData = _assetData.slice(4, _assetData.length);
+
+        (_amounts, _nestedAssetData) = abi.decode(_noSelectorAssetData, (uint256[], bytes[]));
+        
+        // Validate storage refs against the decoded values. Need to convert to memory to compare
+        {
+            require(_amounts[0] == 1);
+            require(_amounts[1] == 0);
+            require(_amounts[2] == 0);
```
epheph, line 500: Check length == 3 here too?

packages/augur-core/source/contracts/trading/ZeroXTrade.sol
```diff
@@ -90,6 +103,10 @@ contract ZeroXTrade is Initializable, IZeroXTrade, IERC1155 {
         fillOrder = IFillOrder(_augurTrading.lookup("FillOrder"));
         require(fillOrder != IFillOrder(0));
 
+        multiAssetValues = [1, 0, 0];
```
epheph: Can this be done without storage ?

packages/augur-core/source/contracts/trading/ZeroXTrade.sol
```diff
@@ -367,16 +450,66 @@ contract ZeroXTrade is Initializable, IZeroXTrade, IERC1155 {
...
+            address _tokenAddress,
+            uint256[] memory _tokenIds,
+            uint256[] memory _tokenValues,
+            bytes memory _callbackData,
+            address _kycToken
+        )
+    {
+         // Read the bytes4 from array memory
+        assembly {
+            _assetProxyId := mload(add(_assetData, 32))
```
epheph, line 480: It's not a problem, but this is a little funny. We are using the memory allocated for the return of the ERC1155 _PROXY_ID to check the `MULTI_ASSET_PROXY_ID`, which is then ignored due to the explicit return.

Probably the most code-efficient way to do it, but a little odd

packages/augur-sdk/src/state/db/ZeroXOrders.ts
nuevoalex: I think this is implicit in the overall length check isn't it?

packages/augur-core/source/contracts/trading/ZeroXTrade.sol
nuevoalex: Yes. Will do

epheph:
This code has really turned into quite a lot of code and hacks to support this! I have some concerns about abi decoding that I'd like to look into more here and in a few other areas. Doesn't block merging but i'm going to keep poking on this aspect.

I'm also worried about just how much calldata this will be when brought to chain. over 4KB, just for the 2x(AssetData)

packages/augur-sdk/src/state/db/ZeroXOrders.ts
nuevoalex: explained to me how it isn't! Will fix.

epheph: Explained offline, but for others reading. If someone takes the tokenValues and maps it to the same offset as tokenIds, you just saved yourself 4 * 32bytes (length + 3 * uint256's), giving you plenty of space to do whatever you want to the abi decoded structure

packages/augur-core/source/contracts/libraries/token/ERC1155.sol
```diff
@@ -210,6 +210,9 @@ contract ERC1155 is ERC165, IERC1155 {
         internal
     {
         require(ids.length == values.length, "ERC1155: IDs and values must have same lengths");
+        if (ids.length == 0) {
```
epheph, line 213: This prevents a potential re-entrancy on on the acceptance check of the fake ShareToken transfer.

epheph (approved):
I wish there was a way to keep the calldata lower. Can taker assetdata just be empty? Or just 0 cash?

The last suggestion I had was about using encodeCashAssetData at runtime instead of reading the pre-computed encode from storage. Even more efficient would be storing the keccek256 to be compared later, but that might be going overboard.

packages/augur-core/source/contracts/trading/ZeroXTrade.sol
```diff
@@ -367,16 +451,67 @@ contract ZeroXTrade is Initializable, IZeroXTrade, IERC1155 {
...
+        
+        // Validate storage refs against the decoded values.
+        {
+            require(_amounts.length == 3);
+            require(_amounts[0] == 1);
+            require(_amounts[1] == 0);
+            require(_amounts[2] == 0);
+            require(_nestedAssetData[1].equals(cashAssetData));
+            require(_nestedAssetData[2].equals(shareAssetData));
+        }
```
epheph, line 503: For posterity: we don't need to check the length of nestedAssetData == 3, because `_amounts` is what drives the actual multi-asset process.

epheph: Also protected from mismatch inside multi-asset proxy

```
                // Revert if number of elements in `amounts` differs from number of elements in `nestedAssetData`
                if sub(amountsLen, nestedAssetDataLen) {
                    // Revert with `Error("LENGTH_MISMATCH")`
```

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). ✌️

**https://peek.run/93phoj2**

-------

packages/augur-core/source/contracts/trading/AugurTrading.sol
```diff
@@ -47,7 +47,7 @@ contract AugurTrading is IAugurTrading {
     //  9:	tokensEscrowed
     event OrderEvent(address indexed universe, address indexed market, OrderEventType indexed eventType, uint8 orderType, bytes32 orderId, bytes32 tradeGroupId, address[] addressData, uint256[] uint256Data);
     event ProfitLossChanged(address indexed universe, address indexed market, address indexed account, uint256 outcome, int256 netPosition, uint256 avgPrice, int256 realizedProfit, int256 frozenFunds, int256 realizedCost, uint256 timestamp);
-    event MarketVolumeChanged(address indexed universe, address indexed market, uint256 volume, uint256[] outcomeVolumes);
+    event MarketVolumeChanged(address indexed universe, address indexed market, uint256 volume, uint256[] outcomeVolumes, uint256 timestamp);
```
epheph, line 50: This is for dev-only?

nuevoalex: No it’s needed for real

epheph: when not in fake-time, wouldn't the log event time be identical?

nuevoalex: what is log event time? There is no time associated with logs AFAIK, just a blockNumber. This can be used to look up a time but requires a request per block

epheph: got it, so it's on-chain event emit to avoid off-chain block time lookups. i figured you already had all timestamp info for blocks for other non-event-based processing, but i haven't looked into the ingest system as much as i have the contracts.

packages/augur-core/source/libraries/ContractCompiler.ts
```diff
@@ -235,7 +237,13 @@ export class ContractCompiler {
                 remappings: [ `ROOT=${this.configuration.contractSourceRoot}/`],
                 optimizer: {
                     enabled: true,
-                    runs: 500
+                    runs: 200,
```
MicahZoltu, line 240: Why only 200 runs here?  Feels like you would want 1000s of runs.
Note: "runs" means "number of times the contract is expected to be executed".  By default, Solidity optimizes for deployment size, more runs prefers optimizations that increase deployment size in favor of decreasing runtime costs.

MicahZoltu (approved):
This PR brings me great joy.

Issue #5021, opened 2019-12-04
Comments covering content on mobile when scrolling


matt-bullock:
Happens when scrolling down then scrolling back up

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/202b8e9a-19a9-4c2d-809c-ca92e3a73695)

- [x] Fix the above issue

- [x] Remove dead space at the bottom of comments section. Should be 16px of spacing between bottom of comments section and bottom of screen

matt-bullock:
I'm still seeing the comments covering content. To replicate issue: scroll to very bottom of page then scroll up by touching anywhere other than the comments section and the comments section will stick to the bottom of the screen whilst everything else scrolls under it

Chwy5:
I want to remove the comments section from this screen only.....No reason to make a comment as you are entering an order

PR #5439, opened 2020-01-07
remove-comments-mobile-trade

JohnDanz:
added functionality to handle hiding/removing the comments component from the trade tab in mobile view only on the market view page
https://github.com/AugurProject/augur/issues/5021


bthaile (approved):
lgtm. this is a pane! right, ha

phoebemirman (approved):

PR #5441, opened 2020-01-08
0x gas estimation buffer for gnosis

robert-e-davidson3:
Properly ignoring already-matched orders.
Cleanup.
Adding 10% buffer to gnosis txs instead of adding gas buffer to Transaction.
Correctly calculating percentage.

### DIFF
--- a/packages/ethersjs-provider/src/EthersProvider.ts
+++ b/packages/ethersjs-provider/src/EthersProvider.ts
@@ -23,10 +23,7 @@ interface PerformQueueTask {
 export class EthersProvider extends ethers.providers.BaseProvider
   implements EProvider {
   gasLimit: ethers.utils.BigNumber | null = new ethers.utils.BigNumber(7500000);
-  gasEstimateIncreasePercentage: ethers.utils.BigNumber | null = new ethers.utils.BigNumber(
-    10
-  );
-
+  gasEstimateIncreasePercentage: ethers.utils.BigNumber | null = new ethers.utils.BigNumber(34);
   private contractMapping: ContractMapping = {};
   private performQueue: AsyncQueue<PerformQueueTask>;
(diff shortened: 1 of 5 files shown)


packages/gnosis-relay-api/src/GnosisRelayAPI.ts
```diff
@@ -183,7 +184,16 @@ export class GnosisRelayAPI implements IGnosisRelayAPI {
...
+      const relayTxEstimate: RelayTxEstimateResponse = result.data;
+
+      if (this.gasEstimateIncreasePercentage) {
+        relayTxEstimate.safeTxGas = relayTxEstimate.safeTxGas.plus(
+          relayTxEstimate.safeTxGas.div(
+            new BigNumber(100).div(this.gasEstimateIncreasePercentage))
+        );
+      }
+
+      return relayTxEstimate;
```
robert-e-davidson3, line 196: This is the important change.

packages/ethersjs-provider/src/EthersProvider.ts
```diff
@@ -113,7 +114,8 @@ export class EthersProvider extends ethers.providers.BaseProvider
     let gasEstimate = await super.estimateGas(transaction);
     if (this.gasEstimateIncreasePercentage) {
       gasEstimate = gasEstimate.add(
-        gasEstimate.div(this.gasEstimateIncreasePercentage)
+        gasEstimate.div(
+          new ethers.utils.BigNumber(100).div(this.gasEstimateIncreasePercentage))
```
robert-e-davidson3, line 115: There was a bug in how we calculated percentages that wasn't apparent because the value used was 10.

joeykrug (approved):

joeykrug (approved):

Issue #5207, opened 2019-12-12
App/Nav: Mobile - logged in -See Chwy's comments below


cillianhunter0:
App bar: 
https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=425%3A51

- [ ] Match text styles to figma **HIGH**
- [ ] Match text spacing to figma **HIGH**
- [ ] Change the bell icon and color (same as desktop ticket)

Comparison: 
![image.png](https://images.zenhubusercontent.com/5c7591d7eec15745ffae99b1/f1c17160-81fc-4306-8daa-fb90893324c0)

Hamburger menu logged in: 
https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=12854%3A298319

- [ ] Update top row background color to match figma (same as logged out ticket) **HIGH**
- [ ] Replace the 'X' icon with the one used in figma + match color **HIGH**
- [ ] Update the (?) help icon to match figma **HIGH**
- [ ] Update text styles to match figma **Medium** 
- [ ] Add in logo and divider line below **HIGH**
- [ ] Add in the 'add funds' button **HIGH** 
- [ ] Match the color of the inactive navigation link to figma **HIGH** 
- [ ] Add in the create market button and divider line above **Low** if non trivial change the text from 'create' to 'create market'
- [ ] Match 'logout' text + icon style and color with figma **Low** 
- [ ] Match 'global chat' button component with figma **Low**  

Comparison
![image.png](https://images.zenhubusercontent.com/5c7591d7eec15745ffae99b1/219f9407-aa29-49a0-834b-12b9b328173f)

Chwy5:
All I need done from this ticket for Launch is:
1) Add in Funds Button
2) Make font color match so the choices are easier to read
3) Add In create Market button

matt-bullock:
The add funds button should be green. looks like an old mock in the ticket above.

https://www.figma.com/file/fLWVwmanAwetVZbujQquEi/Market-Page?node-id=2138%3A8445

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/1b007dbf-4d9c-4363-bd1b-b9462ffe1e5c)

For some reason the create market button was changed to a text button style. This wasn't a design request but it actually works fine now we have the add funds button and I suspect this was done to create space. If we do this on desktop we should carry through to mobile

[...]

Issue #5206, opened 2019-12-12
App/Nav: Mobile - logged out 


cillianhunter0:
App bar: 
https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=11719%3A90
- [x] Add in the help menu button **HIGH**

Comparison: 
![image.png](https://images.zenhubusercontent.com/5c7591d7eec15745ffae99b1/8ea0b251-09c8-41f9-834a-3766943a4536)

Hamburger menu: 
https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=12854%3A300448

- [ ] Update color of top row of the hamburger menu to match figma **HIGH**
- [ ] Add in the help button and login/signup buttons to the top row **HIGH**
- [ ] Add in logo with divider line below **Medium**
- [ ] Match the colour of the bottom row with the global chat button to figma **LOW** 

Comparison: 
![image.png](https://images.zenhubusercontent.com/5c7591d7eec15745ffae99b1/ee079d00-8c60-43ae-92f9-3585404879b7)

Chwy5:
moving to post launch for remainder

Issue #5202, opened 2019-12-12
App/Nav: Desktop - logged in 


cillianhunter0:
Figma: 
https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=635%3A30189

Left side:
- [ ] Match the test styles used in figma **HIGH**
- [ ] Match the alignment and spacing used in figma **HIGH**

Right side: 
- [ ] Change the 'help' and 'bell' icons to match figma (24px icons with color D7DDE0) **HIGH**
- [ ] Move the account menu to the right side, so the order is 'help', 'alerts' & 'account' **HIGH** 
- [ ] When the alerts menu has the red count, match the spacing to figma (currently it has a lot of room around the number in the build + match the text style of the number **HIGH** 
- [ ] Add the 'add funds' button **HIGH** (Note: this will be removed at tablet breakpoint)
- [ ] Match the test styles used in figma **HIGH**

Comparison: 
![image.png](https://images.zenhubusercontent.com/5c7591d7eec15745ffae99b1/14c003eb-67f7-4ea7-964c-76e25b3f1ef1)

Nav bar: 
- [ ] Match the create market button styling to figma button 
![image.png](https://images.zenhubusercontent.com/5c7591d7eec15745ffae99b1/8985e64a-83d6-41fe-a902-25a18eaae056)

Chwy5:
closing

Issue #5201, opened 2019-12-12
App/Nav: Desktop - logged out 


cillianhunter0:
- [x] Add the help menu when logged out **HIGH**
https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=9597%3A277770
- [x] Fix the spacing in the nav bar, it should be an even 12px above and below the nav links **HIGH** 
- [x] When hovering on links in the nav bar, the text is the only area that allows me to click, increase the size of the click area **HIGH** 

Comparison: 
![image.png](https://images.zenhubusercontent.com/5c7591d7eec15745ffae99b1/c7bfdc52-ee15-49e5-8602-2fda5f42a783)

Nav screenshot from build, it's not 12px above and below:
![image.png](https://images.zenhubusercontent.com/5c7591d7eec15745ffae99b1/edcc1130-ff6d-48e4-b7a2-c7796751d5e7)

JohnDanz:
Design actually has 11 px above and below text and v2 matches design exactly right now for the nav.

PR #5455, opened 2020-01-08
Fix mobile/desktop handling of Help Tooltip Toggle icon in Top Nav

jfarid27:
## Description

Adds appropriate cases for top nav Help Tooltip Toggle visibility.

To facilitate the change, I have removed some ambiguous Less rules and moved visibility control to purely JS using the `isMobile` state in the `appState` variables. It is currently a mix of JS and CSS, and I figured it would be best to move all visibility control to JS since you have to test at the JS level `loggedIn` anyway.

## Screenshots
- Desktop - Logged In:     SHOW

    Screenshot from 2020-01-08 14-28-03

- Desktop - Logged Out:  SHOW
    
    Screenshot from 2020-01-08 14-27-48

- Mobile - Logged In:       HIDE

    Screenshot from 2020-01-08 14-27-21

- Mobile - Logged Out:    SHOW
    Screenshot from 2020-01-08 14-27-33

## Issues

#5207 
#5206 
#5202 
#5201 

## Note

- This clears up some requirements in the above issues and does not close them out fully.
- This does not fix help icon styling. Punting that to a separate PR.

### DIFF
--- a/packages/augur-ui/src/modules/app/components/help-resources.styles.less
+++ b/packages/augur-ui/src/modules/app/components/help-resources.styles.less
@@ -5,12 +5,8 @@
   position: relative;
   display: flex;
   flex: 1;
-  max-width: 57px;
+  max-width: 56px;
   z-index: @above-all-content-above;
-
-  @media @breakpoint-mobile {
-    display: none;
-  }
 
   > span {
     padding: @size-18;
--- a/packages/augur-ui/src/modules/app/components/top-bar.styles.less
+++ b/packages/augur-ui/src/modules/app/components/top-bar.styles.less
@@ -21,19 +21,6 @@
       padding: @size-8 0;
     }
 
-
-    > div:last-of-type {
-      @media @breakpoint-mobile {
-        display: none;
-      }
-    }
-
-    > div > div:last-of-type {
-      @media @breakpoint-mobile {
-        display: none;
-      }
-    }
-
     > button {
       margin-right: @size-12;
     }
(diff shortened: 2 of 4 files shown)


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). ⚡️

**https://peek.run/0elhx35**

-------

JohnDanz:
Thanks for the PR @jfarid27, 

We are actually trying to move away from `isMobile` and would prefer a CSS only approach ideally. Eventually we plan to completely refactor out `isMobile` we just haven't done it completely yet and have been taking a more "if you see it, try and get rid of it from being a requirement" approach for now. It would be great to get this same effect without the use of `isMobile` while making the CSS less ambiguous if you can find a way to do that.

jfarid27:
@JohnDanz Could I talk you out of that? The issue here is you have to have show and hide logic on the JS tier anyways because you have to determine if the user is logged in. I actually think a major part of the confusion I had was needing to think about logic on the CSS side, as well as the JS side. Why not keep everything in JS-land?

It's definitely doable with JS and CSS, but it would only add to code complexity with JS and CSS logic. What would you think works best here?

JohnDanz:
> @JohnDanz Could I talk you out of that? The issue here is you have to have show and hide logic on the JS tier anyways because you have to determine if the user is logged in. I actually think a major part of the confusion I had was needing to think about logic on the CSS side, as well as the JS side. Why not keep everything in JS-land?
> 
> It's definitely doable with JS and CSS, but it would only add to code complexity with JS and CSS logic. What would you think works best here?

I still want to eventually get rid of isMobile, however it's not gone yet and this is a pretty soft request overall on my part so you have convinced me that it's fine. 👍 Approved.

JohnDanz (approved):

PR #5460, opened 2020-01-09
actually disable smoke test

robert-e-davidson3:
I refactored the smoke test but it still ran beforeAll despite being skipped, so I also removed it from jest.config.json.


bthaile (approved):
lgtm

robert-e-davidson3:
I added this to https://github.com/AugurProject/augur/pull/5463 so it could pass CI.

Issue #5433, opened 2020-01-07
templates with TEXT (freetext)


bthaile:
[Single Player's Name], [Single Candidate's Name], [Single Person's Name], [Single Horse's Name], [Single Person, Single Group or Movie Title], [Individual Movie Title], [Individual Stock or ETF Name], [Individual Stock or ETF Ticker Symbol]

Added additional Rule to include:  
This market is intended to be about a [Single Person, Single Group or Movie Title], if this is not the case, this market should settle as invalid.

PGA: Will Tiger Woods win the 2020 PGA Championship?
PGA: Will Tiger Woods make the cut at the 2020 PGA Championship?
Euro Tour: Will Rory Mcllroy win the 2020 Omega Dubai Dessert Classic?
Euro Tour: Will Rory Mcllroy make the cut at the 2020 Omega Dubai Dessert Classic?
LPGA: Will Lexi Thompson win the 2020 U.S. Women's Open?
LPGA: Will Lexi Thompson make the cut at the 2020 U.S. Women's Open?
Will American Pharoah win the 2020 Triple Crown?
Men's Singles Tennis: Will Roger Federer win the 2020 Wimbledon?
Men's Double Tennis: Will Juan Sebastin/Robert Farah win the 2020 Wimbledon?
Men's Soccer (Point Spread): Real Madrid to win by more than 1.5 goals over Manchester United?\nEstimated schedule start time: Sept 19, 2019 1:00 pm EST
Men's Soccer (O/U): Real Madrid vs. Manchester United: Total goals scored Over/Under 4.5?\nEstimated schedule start time: Sept 19, 2019 1:00 pm EST
Women's Soccer (Point Spread): Real Madrid to win by more than 1.5 goals over Manchester United?\nEstimated schedule start time: Sept 19, 2019 1:00 pm EST
Women's Soccer (O/U): Real Madrid vs. Manchester United: Total goals scored Over/Under 4.5?\nEstimated schedule start time: Sept 19, 2019 1:00 pm EST
NBA: Will Steph Curry win the 2019-20 NBA Most Valuable Player award?
WNBA: Will Maya Moore win the 2019-20 WNBA Most Valuable Player award?
NFL: Will Patrick Mahomes win the 2019-20 MVP award?
NCAA FB: Which team will win the 2020 Cotton Bowl: Tennessee Volunteers vs. Miami Hurricanes?
Will Donald Trump win the 2020 U.S. Presidential election?
[...]

PR #5466, opened 2020-01-09
update template source to add resolution rules and specific place hol…

phoebemirman:
…ders for text inputs

https://github.com/AugurProject/augur/issues/5433


packages/augur-tools/src/templates-source.ts
```diff
@@ -1392,6 +1426,9 @@ export const TEMPLATES = {
                     {
                       text: `If the game is NOT played or is not deemed an official game, meaning, less than 90% of the scheduled match had been completed, the market should resolve as "Unofficial game/Cancelled".`,
                     },
+                    {
+                      text: 'This market is intended to be about a Single Group, if this is not the case, this market should settle as invalid.'
```
bthaile: Single group doesnt makes sense here. Its a sports market, [single team verse single team] makes more sense. Also we didnt cover categorical markets. Good discovery.

phoebemirman: ok updated to teams

bthaile (approved):
lgtm

PR #5479, opened 2020-01-10
Add cancellation orders to transactions history

bthaile:
integrating with cancelledOrders to display them transactions history


robert-e-davidson3 (approved):
s'good

Issue #4494, opened 2019-10-29
`hasRecentlyDepletedLiquidity` needs to be populated on Markets in the Markets Derived DB


nuevoalex:
This was previously being calculated during the getter call. Instead it should be updated in the derived DB itself periodically. The unused code which used to calculate this value still exists though it should be cautiously used a reference only when doing this task, then deleted.

pgebheim:
Merged.

PR #5485, opened 2020-01-10
update liquidity/invalid filters

andrewdamelio:
## update liquidity/invalid filters

- Update MarketDB to use ZeroXOrders collections in orders to fix liquidity filters
- Update filters so we only count for the liquidity sorts/filters up to 3 orders per side of the book per outcome per user per market, filters aren’t working at all atm either
- hasRecentlyDepletedLiquidity` needs to be populated on Markets in the Markets Derived DB [#4494]
- cleanup/remove old unused liquidity code
- other minor updates

@nuevoalex This was all pretty new to me codebase wise, so i'm sure there is some room for doing things better here.


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). 💚

**https://peek.run/kq3h4v5**

-------

packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -87,37 +88,31 @@ export class MarketDB extends DerivedDB {
...
     // TODO Get ETH -> DAI price via uniswap when we integrate that as an oracle
     const ETHInAttoDAI = new BigNumber(200).multipliedBy(10**18);
 
+    const marketDataById = _.keyBy(marketsData, 'market');
+
     for (const marketId of marketIds) {
       const doc = await this.getOrderBookData(this.augur, marketId, marketDataById[marketId], reportingFeeDivisor, ETHInAttoDAI);
       // This is needed to make rollbacks work properly
-      doc['blockNumber'] = highestBlockNumber;
+      doc['blockNumber'] = highestSyncedBlockNumber;
```
nuevoalex: Do we even need this anymore?

andrewdamelio: I don't think so -- let me double check, i think i just missed that in cleanup

nuevoalex: More on this. It appears this function now only does a full one time sync with no update capability. How do we handle updating liquidity as new orders come in? Previously we would get the latest order states from the currentOrders DB which would handle decreases or complete fills to decrease a markets liquidity as well as any new orders. In this code I think we'd end up updating the world on every single call to this. Is this now only used for bulk sync?

packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -158,33 +151,46 @@ export class MarketDB extends DerivedDB {
   }
 
   async getOrderBook(marketData: MarketData, numOutcomes: number, estimatedTradeGasCostInAttoDai: BigNumber): Promise<OrderBook> {
-    const currentOrdersResponse = await this.stateDB.CurrentOrders.where("[market+open]").equals([marketData.market, 1]).and((order) => {
-      return order.amount > '0x00';
-    }).toArray();
+    const currentOrdersResponse = await this.stateDB.ZeroXOrders
```
nuevoalex, line 191: Why aren't we using an index anymore? This does a full DB pull for every market on every 0x order to come in currently

packages/augur-sdk/src/state/getter/Markets.ts
```diff
@@ -559,8 +559,13 @@ export class Markets {
       if (params.includeInvalidMarkets !== true && market.invalidFilter) return false;
       // Liquidity filtering
       if (params.maxLiquiditySpread) {
-        // TODO hasRecentlyDepletedLiquidity == true on ZeroPercent spread
-        if (params.maxLiquiditySpread !== MaxLiquiditySpread.OneHundredPercent) {
+        if (params.maxLiquiditySpread === MaxLiquiditySpread.ZeroPercent) {
+          // return hasRecentlyDepletedLiquidity on ZeroPercent spread
+          if (!market.hasRecentlyDepletedLiquidity) {
```
nuevoalex, line 565: I don't think this is right. hasRecentlyDepletedLiquidity needs to also consider when the market last passed the liquidity check. That's all that commented out code getting deleted in this PR. It was implemented in a very bad way originally and the issue to restore it correctly has been languishing.

nuevoalex (changes requested):
I believe we need some strategy for handling only updating markets with actual changes as well as making the orderbook update actually use an index.

Sort of as a general thing we should probably figure out a way to limit the times an update occurs as well. Consider that since 0x orders occur off chain we could be getting 10s of these every second. We don't want/need to regenerate the liquidity data for a market that often, so we should likely limit it to some cadence.

packages/augur-sdk/src/state/db/MarketDB.ts
andrewdamelio: Yeah my mistake! - fixing

packages/augur-sdk/src/state/db/MarketDB.ts
andrewdamelio: I thought i had this wired up to run every-time the zeroXOrders table gets updated
```this.augur.events.subscribe('DerivedDB:updated:ZeroXOrders', this.syncOrderBooks);```


Also i left blockNumber logic in b/c of the comment below it, wasn't sure if i should remove it
```
// This is needed to make rollbacks work properly
```

File: packages/augur-sdk/src/state/getter/Markets.ts (hunk 3 above)
Comment (andrewdamelio) [reply packages/augur-sdk/src/state/getter/Markets.ts:565]: added ``lastPassingLiquidityCheck`` and ``lastLiquidityCheck`` timestamps on MarketDB in order to calc this.

**Comment by andrewdamelio** [comment]
~Currently using a cadence of 30seconds before re-running liquidity checks on markets if we have active zeroXOrders. Does 30 seconds sound reasonable?~

File: packages/augur-sdk/src/state/db/MarketDB.ts [hunk 4]
```diff
@@ -63,14 +63,15 @@ export class MarketDB extends DerivedDB {
     };
 
     this.augur.events.subscribe('DerivedDB:updated:CurrentOrders', this.syncOrderBooks);
```
nuevoalex: This should be removed I think

packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -85,37 +86,43 @@ export class MarketDB extends DerivedDB {
...
-
     const reportingFeeDivisor = await this.augur.contracts.universe.getReportingFeeDivisor_();
     // TODO Get ETH -> DAI price via uniswap when we integrate that as an oracle
     const ETHInAttoDAI = new BigNumber(200).multipliedBy(10**18);
 
+    const marketDataById = _.keyBy(marketsData, 'market');
+
+    const now = Math.floor(Date.now() / 1000);
+
+    // only check market if it hasn't already been checked in the last 30s
```
nuevoalex: This check is good for the debouncing sort of behavior but will still update more markets than we need to check.

I think we can address this by actually looking at the order events that are provided when handling the subscription event here: `this.augur.events.subscribe('DerivedDB:updated:ZeroXOrders',` and get the unique market ids out of that to see what needs updating

packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -354,4 +412,38 @@ export class MarketDB extends DerivedDB {
...
+
+
+    if (marketData.lastPassingLiquidityCheck === 0) {
+      hadLiquidityInLast24Hour = false;
+    }
+    else if (lastPassingLiquidityCheck > oneDayAgo) {
+      hadLiquidityInLast24Hour = true;
+    }
+    else if (lastPassingLiquidityCheck <= oneDayAgo) {
+      hadLiquidityInLast24Hour = false;
```
nuevoalex, line 422: should this be true and the condition above be false?

nuevoalex (changes requested):
Looking really good! Had two suggestions and it needs some tests but I think this is it!

andrewdamelio:
@nuevoalex Updated ZeroXOrders to emit events we can use to hook into the MaretDB for liquidity.

Also updated tests to support zeroXOrders table for MarketDB since currentOrders / onChain aren't being used anymore and the tests wouldn't work anymore with these changes.

andrewdamelio:
@nuevoalex Updated ZeroXOrders to emit events we can use to hook into the MarketDB for liquidity.

Also updated tests to support ZeroXOrders table for MarketDB since currentOrders / onChain aren't being used anymore and the tests wouldn't work anymore with these changes.

packages/augur-sdk/src/api/Liquidity.ts
```diff
@@ -47,6 +47,26 @@ export interface GetLiquidityParams {
   spread?: number;
 }
 
+// Helpers
+const sortBids = (collection) => {
+  return collection.sort((a, b) => {
```
nuevoalex, line 52: this could just be `sortAsks` with a `reverse`

packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -62,15 +61,16 @@ export class MarketDB extends DerivedDB {
       'MarketMigrated': this.processMarketMigrated,
     };
 
-    this.augur.events.subscribe('DerivedDB:updated:CurrentOrders', this.syncOrderBooks);
+    this.augur.events.subscribe('DB:updated:ZeroXOrders', (orderEvents) => this.syncOrderBooks([orderEvents.market]));
+    this.augur.events.subscribe('DB:get:ZeroXOrders', (markets) => this.syncOrderBooks(markets));
```
nuevoalex, line 65: when does `DB:get:ZeroXOrders` happen?

packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -85,57 +85,55 @@ export class MarketDB extends DerivedDB {
     }
   }
 
-  syncOrderBooks = async (syncing: boolean): Promise<void> => {
+  syncOrderBooks = async (marketIds: string[]): Promise<void> => {;
```
nuevoalex, line 88: I think this still needs some sort of debouncing behavior on it such that we only recalc liquidity on some fixed cycle (could be as long as once per 5 sec or something)

packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -152,37 +150,84 @@ export class MarketDB extends DerivedDB {
...
+
+    marketOrderBookData.hasRecentlyDepletedLiquidity = await this.hasRecentlyDepletedLiquidity(marketData, marketOrderBookData.liquidity, marketOrderBookData.lastPassingLiquidityCheck);
     return marketOrderBookData;
   }
 
   async getOrderBook(marketData: MarketData, numOutcomes: number, estimatedTradeGasCostInAttoDai: BigNumber): Promise<OrderBook> {
-    const currentOrdersResponse = await this.stateDB.CurrentOrders.where("[market+open]").equals([marketData.market, 1]).and((order) => {
-      return order.amount > '0x00';
-    }).toArray();
+    const orderTypes = ['0x00', '0x01'];
```
nuevoalex, line 173: these could be constants outside of the loop (at the top of the file)

nuevoalex (approved):
Had some comments/suggestions but given theres another PR to do after this those can go in that if desired

packages/augur-sdk/src/state/db/MarketDB.ts
andrewdamelio: https://github.com/AugurProject/augur/pull/5485/files#diff-c11a07414c9132850ffeab97bd3d143aR217

packages/augur-sdk/src/state/db/MarketDB.ts
joeykrug: Imo should only recalculate once a minute or so, doesn’t need to be super frequent

PR #5488, opened 2020-01-11
Ignore orders that will soon expire

robert-e-davidson3:
added 0x order timeout cutoff to getter, to avoid orders that will soon expire


bthaile (approved):
lgtm

robert-e-davidson3:
I'm cutting this PR short since it's useful but not complete. We need to vary the expiry by gas price and also differ by which orders we show the user and which we take.

Issue #5500, opened 2020-01-12
Validity bond needs to be payable with vDAI.


MicahZoltu:
https://github.com/AugurProject/augur/blob/de76fd1ce689e0ec52e0bf2f07bd64d796185485/packages/augur-core/source/contracts/reporting/Market.sol#L102-L103

Part of the "vDAI everywhere" series of issues.

The validity bond needs to be payable with either DAI or vDAI or (ideally, but not required) a combination of both.  During Maker shutdown, it is possible (arguably likely) that DAI will dry up and all that is left is vDAI.  In such a case, in order for Augur to remain functional markets need to be able to be created with vDAI rather than just DAI.

Resolution: Take the sum of DAI + vDAI balance here, rather than just DAI balance.

Issue #5499, opened 2020-01-12
Consider: Rename `Market.isValid` to `Market.isFinalizedAsValid`.


MicahZoltu:
https://github.com/AugurProject/augur/blob/de76fd1ce689e0ec52e0bf2f07bd64d796185485/packages/augur-core/source/contracts/reporting/Market.sol#L594-L600

I just wasted a bunch of time on what I thought was a bug because this function's name didn't make it clear that it would fail if market isn't finalized (the calling code was implicitly depending on that failure to occur).

A rename to `isFinalizedAsValid` or similar would make it more obvious that this function is asserting finalization.

Issue #5498, opened 2020-01-12
Trading needs to support vDAI


MicahZoltu:
ZeroXTrade currently works exclusively with `cash`.  During global shutdown, anytime a complete set is sold it will generate vDAI, rather than DAI, that is transferred to the user.  This means that a trader may open a position with DAI but then close the position and walk away with vDAI that cannot be converted to DAI.  This means the user cannot then use that capital to open a new trade, they need to go somewhere else to trade vDAI for DAI before they can then open a new 0x trade.

Not only does this provide a poor user experience, over time DAI could become a scarce resource as Augur migrates DAI to vDAI with each complete set purchase/sale.  This could lead DAI to end up with a hefty price premium over the underlying claim on collateralized assets compared to vDAI.

Resolution: Add support for using vDAI in 0x trades.  I don't have a complete solution, but at the least the `cash.allowance` checks need to change to `CashSender.allowance` checks:
https://github.com/AugurProject/augur/blob/de76fd1ce689e0ec52e0bf2f07bd64d796185485/packages/augur-core/source/contracts/trading/ZeroXTrade.sol#L170

Under the hood I suspect more work will be required to make 0x work with mixed DAI/vDAI, but I'll leave that to an exercise for the reader.  😉

nuevoalex:
I believe this is just that the ZeroXTrade balanceOf function will become inaccurate over time. Trading does work with vDai already so shouldn’t be too bad

MicahZoltu:
"Inaccurate over time" could easily result in "unusuable over time" if DAI liquidity dries up while vDAI liquidity does not.  Since DAI => vDAI is one way during shutdown, and Augur is actively doing that one-way transition, it is very possible (I would posit even likely) that post-shutdown, Maker will end up with almost no DAI and some vDAI available.

MicahZoltu:
It is worth noting that vDAI doesn't have an allowance function.  There is a balance function and an "authorized to send" boolean.  This means that vDAI trading is either on or off.

It is also worth noting that vDAI is not an ERC20 token... which may cause problems with 0x's trading functions (there is no `.transfer`, `.transferFrom`, etc.)

nuevoalex:
Just noting it won’t result in direct loss of funds or anything. I agree we should make the balanceOf function handle shutdown.

CashSender handles doing the equivalent of transfer and transferFrom for vDai

Issue #5497, opened 2020-01-12
migrateMarketOut needs to do a Vat DAI approval.


MicahZoltu:
Problematic call stack:
https://github.com/AugurProject/augur/blob/de76fd1ce689e0ec52e0bf2f07bd64d796185485/packages/augur-core/source/contracts/reporting/Universe.sol#L399-L400
https://github.com/AugurProject/augur/blob/de76fd1ce689e0ec52e0bf2f07bd64d796185485/packages/augur-core/source/contracts/reporting/Universe.sol#L404-L407
https://github.com/AugurProject/augur/blob/de76fd1ce689e0ec52e0bf2f07bd64d796185485/packages/augur-core/source/contracts/reporting/Universe.sol#L699-L701
https://github.com/AugurProject/augur/blob/de76fd1ce689e0ec52e0bf2f07bd64d796185485/packages/augur-core/source/contracts/Augur.sol#L200-L202
https://github.com/AugurProject/augur/blob/de76fd1ce689e0ec52e0bf2f07bd64d796185485/packages/augur-core/source/contracts/CashSender.sol#L32-L34
https://github.com/AugurProject/augur/blob/de76fd1ce689e0ec52e0bf2f07bd64d796185485/packages/augur-core/source/contracts/CashSender.sol#L42-L45

`vat.hope` is effectively the same as `dai.approve` and needs to be called so that during global shutdown Vat DAI can be transferred.  Without this change, if `Universe` has a vDAI balance I believe migration will fail and Augur will halt since the `vat.move` function will revert due to not having permission.

This is possible to maliciously hit by doing `Universe.deposit` post-shutdown from an account that has vDAI but no DAI.  This will result in vDAI being sent to the Universe since `Augur.trustedCashTransfer` correctly transfers both DAI or vDAI (depending on availability).  This will result in `Universe.marketBalance` being incremented to include some vDAI.  Once in this state, a migration will attempt to migrate all of the DAI + vDAI and since vDAI is non-zero, the migration will fail and Augur will halt (since in a Fork Augur halts on the primary universe, and migration to child universe isn't possible).

MicahZoltu:
### Commentary

[...]

PR #5502, opened 2020-01-12
various MKR shutdown and cleanliness issues

nuevoalex:
Includes various closed nit PRs as well as:

#5500
#5499
#5498
#5497
#5496
#5494
#5495


packages/augur-core/source/contracts/reporting/Universe.sol
```diff
@@ -397,6 +397,7 @@ contract Universe is IUniverse, CashSender {
         withdraw(address(this), _cashBalance, msg.sender);
         openInterestInAttoCash = openInterestInAttoCash.sub(_marketOI);
         cash.approve(address(augur), _cashBalance);
+        vat.hope(address(augur));
```
epheph: Should we do the `hope` in the the initialization of universe or at the point of forking, since it doesn't need to happen for every market that moves?

MicahZoltu: I believe the `approve` is here so the approval is limited to the OI owned by this market.  This offers some protection if a malicious fork can somehow manage to steal all of the assets in Augur, it would be limited to being able to steal whatever assets it was approved for.  Vat DAI on the other hand cannot limit the losses to the approved amount since approval is all-or-nothing.  So the question I think is whether or not we want to split them up (which would be confusing) or keep them together (which results in unnecessary calls to vat.hope).

packages/augur-core/tests/test_cashSender_usage.py
```diff
@@ -11,8 +11,9 @@ def test_cashSender_usage():
...
-                    expectedOccurences = 1
-                notFound = data.count("cash.transfer") == expectedOccurences
-                assert notFound, "Contract %s has an unexpected cash.transfer in it" % file
\ No newline at end of file
+                notFound = data.count("cash.transfer") == 0
+                assert notFound, "Contract %s has an unexpected cash.transfer in it" % file
+                notFound = data.count("cash.balanceOf") == 0
+                assert notFound, "Contract %s has an unexpected cash.balanceOf in it" % file
+                notFound = data.count("cash.allowance") == 0
+                assert notFound, "Contract %s has an unexpected cash.allowance in it" % file
```
MicahZoltu: There may be value in adding a method for `cashApprove` which will approve both DAI and vDAI.  Then we can add protection for `cash.approve` calls here as well.

We should also add a check for `cash.transferFrom` (though maybe that is counted in `cash.transfer` (my Python is weak)?

MicahZoltu (approved):
A couple points of feedback that I encourage looking into, but nothing that would block this PR from getting merged.

packages/augur-core/source/contracts/reporting/Universe.sol
nuevoalex: It's called relatively infrequently (Only during a fork but for all unfinalized markets). That said it probably should go in initialization

packages/augur-core/tests/test_cashSender_usage.py
nuevoalex: yea cash.transferFrom is handled by this expression.

nuevoalex: for `cashApprove` thats a good point.

PR #5560, opened 2020-01-15
finish adding to list of holidays

phoebemirman:


packages/augur-tools/src/templates-lists.ts
```diff
@@ -8347,6 +8347,46 @@ export const HOLIDAYS = {
   NATIONAL_DAY_6: 'National Day 6',
   NATIONAL_DAY_7: 'National Day 7',
   NATIONAL_DAY_8: 'National Day 8',
+  AUSTRALIA_DAY: 'Australia Day',
```
bthaile, line 8350: doesn't look like `AUSTRALIA_DAY` is referenced anywhere. is this needed?

and the others ...

bthaile (approved):
lgtm

PR #5567, opened 2020-01-16
Audit issues 2

nuevoalex:


packages/augur-core/source/contracts/libraries/Ownable.sol
```diff
@@ -40,8 +40,9 @@ contract Ownable is IOwnable {
         if (_newOwner != address(0)) {
             onTransferOwnership(owner, _newOwner);
             owner = _newOwner;
+            return true;
         }
-        return true;
+        return false;
```
MicahZoltu: It feels like this function should `require(_newOwner != address(0))` rather than returning false on failing to do the thing you told it to do.

MicahZoltu (approved):

azure-pipelines.yml
```diff
@@ -101,7 +101,7 @@ jobs:
       matrix:
         node_10_x:
           node_version: $(PRIMARY_NODE_VERSION)
-    timeoutInMinutes: 90
+    timeoutInMinutes: 120
```
epheph, line 103: 😭

packages/augur-core/source/contracts/reporting/Market.sol
```diff
@@ -426,6 +426,7 @@ contract Market is Initializable, Ownable, IMarket, CashSender {
         IMarket _forkingMarket = universe.getForkingMarket();
         require(_forkingMarket.isFinalized());
         require(!isFinalized());
+        require(this != warpSync.markets(address(universe)));
```
epheph, line 429: Interesting! 🤯

epheph (approved):

PR #5577, opened 2020-01-16
Add logging

nuevoalex:


packages/augur-core/source/contracts/Augur.sol
```diff
@@ -68,6 +69,8 @@ contract Augur is IAugur, IAugurCreationDataGetter, CashSender {
     event NoShowBondChanged(address indexed universe, uint256 noShowBond);
     event ReportingFeeChanged(address indexed universe, uint256 reportingFee);
     event ShareTokenBalanceChanged(address indexed universe, address indexed account, address indexed market, uint256 outcome, uint256 balance);
+    event MarketRepBondTransferred(address indexed universe, address market, address from, address to);
```
epheph, line 72: Nit, but why not emit an IUniverse?

nuevoalex: I think its just from an older time and a desire to be explicit about whats actually returned. Like its more helpful to see a uint8 over an Enum value for example since the consumer of these is a parser in another context.

epheph (approved):

packages/augur-core/source/contracts/Augur.sol
MicahZoltu: I disagree with this assessment (about the low level type being better).  It means that an advanced tool cannot give as much information.  Also, for a human reading the code it is less information available that could be valuable.

There is a reason that all types aren't labeled as `word` and `length-prefixed-bytes` even though at a low level that is what every type in Solidity is.  Types are good, we should use them whenever possible.

nuevoalex: Explaining why I believe it was done not that I believe it’s entirely better. That said I don’t want to prioritize changing it everywhere at this point.

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). 😎

**https://peek.run/1reh3ro**

-------

packages/augur-core/source/contracts/Augur.sol
MicahZoltu: I can get on board with not changing it everywhere, though I think in this case as code was touched (new or modified) I would use interfaces instead of `address` and just recognize that the system is in a transiently inconsistent state while that migration is happening.  That being said, I do understand the desire for consistency and waiting to change patterns until you can do it everywhere at once.

MicahZoltu (approved):

PR #5581, opened 2020-01-16
fix issue with spaces on market titles and also one where some words …

phoebemirman:
…weren't being underlined on hover

https://app.zenhub.com/workspaces/augur-5bfed3424b5806bc2bf8d173/issues/augurproject/augur/5580


bthaile (approved):
that looks like it'll work, lgtm

PR #5583, opened 2020-01-16
update-ui-package

JohnDanz:
updated package.json in the ui to match node versions for the entire mono-repo instead of limiting to 8,9,10


bthaile (approved):
Ok

PR #5588, opened 2020-01-17
New param docker all

bthaile:
add new param to run docker all and deploy contracts with time control on. 

removed creating orders from create canned market call. 
changed docker all to use flash convenience methods, normal-all and fake-all.


scripts/one-docker-to-rule-them-all.sh
```diff
@@ -6,22 +6,37 @@ cleanup() {
   echo "stopping geth docker image"
   docker kill geth
   yarn workspace @augurproject/gnosis-relay-api kill-relay
+  docker volume prune --force
+  docker system prune --force
```
robert-e-davidson3: I worry about taking such destructive action inside a script that isn't obviously this destructive.

bthaile: ok, I'll remove

bthaile:
removed the prunes

robert-e-davidson3 (approved):
lgtm!

Issue #5165, opened 2019-12-10
 Design QA: Account Summary - Transactions


matt-bullock:
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/e1dcdf92-83bd-4525-936e-321f769424b8)

Once desktop changes are done #5164, this should push through to mobile.

- [x] mobile friendly buttons

Issue #5164, opened 2019-12-10
Design QA: Account Summary - Transactions


matt-bullock:
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/20fbe930-e600-41d2-9b12-7b6c0a4f2d0a)

- [ ] Make add funds button blue to standout

- [ ] Add background colour to sections

- [ ] Update type styles

matt-bullock:
looks done. closing

PR #5590, opened 2020-01-17
Account summary transactions

JohnDanz:
https://github.com/AugurProject/augur/issues/5164
https://github.com/AugurProject/augur/issues/5165

I know the tickets say do one thing but I spoke with matt directly here on this and this is ok.


bthaile (approved):
lgtm

PR #5593, opened 2020-01-17
sweep interest on reporting fee recalc

nuevoalex:


packages/augur-core/tests/test_warp_sync.py
```diff
@@ -34,11 +34,16 @@ def test_warp_sync(contractsFixture, augur, universe, reputationToken, warpSync)
...
     # Finalizing the warp sync market will award the finalizer REP based on time since it became finalizable
+    # This will also trigger a sweep of accumulated interest
     expectedFinalizationReward = warpSync.getFinalizationReward(market.address)
+    nextDisputeWindow = universe.getOrCreateNextDisputeWindow(False)
+    initialFeesBalance = cash.balanceOf(nextDisputeWindow)
     with PrintGasUsed(contractsFixture, "WS Market Finalization Cost", 0):
         with TokenDelta(reputationToken, expectedFinalizationReward, account, "REP reward not minted for finalizer"):
             assert market.finalize()
 
+    assert cash.balanceOf(nextDisputeWindow) > initialFeesBalance
```
robert-e-davidson3, line 45: Is it feasible to calculate the expected interest instead of only asserting that it's more than zero?

robert-e-davidson3 (approved):

packages/augur-core/source/contracts/WarpSync.sol
```diff
@@ -70,6 +75,13 @@ contract WarpSync is IWarpSync, Initializable {
             return;
         }
 
+        // In order to periodically sweep interest we do so here which will result in a sweep at least on recurring basis where the sweeper is compensated.
+        uint256 _timestamp = block.timestamp;
```
MicahZoltu, line 79: Should this be using Augur.time (or whatever it is called) rather than `block.timestamp`?  This would allow testing of the timing functionality.

MicahZoltu: If this doesn't end up using Augur time contract, then consider getting rid of the local variable.  I'm not confident (double check) but I think `block.timestamp` is similar cost as reading from memory, so this local variable assignment may not be buying anything.

nuevoalex: Ah yea should use the controlled time

PR #5595, opened 2020-01-17
gnosis gas cap

robert-e-davidson3:
I ran the 3rd party tests so the typical behavior is fine but I didn't test what actually happens when requested gas exceeds the limit.


robert-e-davidson3:
already in

PR #5627, opened 2020-01-21
use web3 provider if it exists

bthaile:
if there exists provider then pass that provider down to sdk. if not then user existing ethereum http url to create provider.


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). 💎

**https://peek.run/vv2h1x7**

-------

packages/augur-ui/src/modules/app/actions/init-augur.ts
```diff
@@ -217,54 +216,58 @@ export function connectAugur(
     if (loggedInAccountType === ACCOUNT_TYPES.TORUS) {
       preloadAccount(ACCOUNT_TYPES.TORUS);
     }
-
-    connect(
+    let provider = new JsonRpcProvider(env['ethereum-node'].http);
+    if (window.web3) {
+      provider = new Web3Provider(window.web3.currentProvider);
+    }
+    const sdk = await augurSdk.makeClient(
```
pgebheim: You're in an `async` function, so you can just `await` here.

pgebheim: Tom pointed out that they want this to happen in parallel with the other initialization as the app can init and then get an error where it puts up the error modal.

So this fine.

pgebheim (approved):

andrewdamelio (approved):
lgtm

PR #5654, opened 2020-01-22
Zerox trading flash

bthaile:
added market order book shaper. 
it creates orders based on internal configuration for passed in marketId. It's the start of order/trade perf testing.


packages/augur-tools/src/flash/scripts.ts
```diff
@@ -1565,7 +1634,7 @@ export function addScripts(flash: FlashSession) {
       if (this.noProvider()) return null;
       const user = await this.ensureUser(this.network, true);
 
-      await (await this.db).sync(user.augur, 100000, 0);
+      //await (await this.db).sync(user.augur, 100000, 0);
```
robert-e-davidson3:
```suggestion
```

packages/augur-tools/src/flash/scripts.ts
```diff
@@ -901,7 +970,7 @@ export function addScripts(flash: FlashSession) {
         );
       }
       this.log(`place order ${result}`);
-      // something hangs need to force exit.
+      // something is hanging. no missings await that I can see.
```
robert-e-davidson3:
```suggestion
```

packages/augur-tools/src/flash/scripts.ts
```diff
@@ -901,7 +970,7 @@ export function addScripts(flash: FlashSession) {
...
       process.exit;
```
robert-e-davidson3:
```suggestion
```

robert-e-davidson3: We do this in cli.ts now so it's unnecessary here. Also it's not actually calling exit here?

packages/augur-tools/src/flash/scripts.ts
```diff
@@ -878,7 +942,12 @@ export function addScripts(flash: FlashSession) {
           displayShares: new BigNumber(0),
           expirationTime,
         };
-        result = fillOrder ? await user.augur.placeTrade(params) : await user.placeZeroXOrder(params)
+
+        try {
+          result = fillOrder ? await user.augur.placeTrade(params) : await user.placeZeroXOrder(params)
+        } catch(e) {
+          this.log(e);
```
robert-e-davidson3, line 1012: Can you add a comment explaining why we're eating errors here?

packages/augur-tools/src/flash/scripts.ts
```diff
@@ -856,12 +919,13 @@ export function addScripts(flash: FlashSession) {
...
           new BigNumber('0.01')
         );
+        console.log('onChainPrice', onChainPrice.toString());
         const price = convertOnChainPriceToDisplayPrice(
           onChainPrice,
           new BigNumber(0),
           new BigNumber('0.01')
         );
-        console.log('order', String(args.amount), '@', price);
+        console.log('order', String(args.amount), '@', price.toString());
```
robert-e-davidson3: Should use `.toFixed()` instead, unless using the ethers bignumber.

packages/augur-tools/src/flash/scripts.ts
```diff
@@ -760,6 +761,67 @@ export function addScripts(flash: FlashSession) {
...
+      while(true) {
+        const timestamp = await this.user.getTimestamp();
+        for(let i = 0; i < orderBooks.length; i++) {
+          const orderBook: OrderBookShaper = orderBooks[i];
+          const marketId = orderBook.marketId;
+          const marketBook: MarketOrderBook = await this.user.augur.getMarketOrderBook({ marketId });
+          const orders = await orderBook.nextRun(marketBook, timestamp);
+          this.log(`creating ${orders.length} orders for ${marketId}`);
+          if (orders.length > 0) {
+            orders.map(order => user.placeZeroXOrder(order));
```
robert-e-davidson3: `map` -> `forEach`

Unless you want to deal with the resulting promises, in which case you'd want to do a `Promise.all(orders.map....)`.

Maybe add a `.catch(this.log)` to the end of each placeZeroXOrder so the errors get handled? (Shouldn't need to bind `this.log` since it's already bound)

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). 🌻

**https://peek.run/o9ph9ej**

-------

packages/augur-tools/src/flash/orderbook-shaper.ts
```diff
@@ -0,0 +1,130 @@
...
+    this.outcomes.map(o => {
+      orders = [...orders, ...this.createMissingOrders(o, currentConfig[o], orderBook[o])];
+    })
+
+    return this.createOrders(orders, timestamp);
+  }
+
+  createMissingOrders = (outcome: number, orderBookConfig: OrderBookConfig, bidsAsks: BidsAsks): SimpleOrder[] => {
+    let orders: SimpleOrder[] = []
+    Object.keys(orderBookConfig).map(type => {
```
robert-e-davidson3: Same deal as below, where `map` that doesn't return anything can be `forEach`. The efficiency gain is probably irrelevant but the semantics are clearer.

packages/augur-tools/src/flash/orderbook-shaper.ts
```diff
@@ -0,0 +1,130 @@
...
+  fiveMinutesInSeconds = new BigNumber(18000);
+
+  constructor (marketId: string, outcomes: number[] = [1, 2]) {
+    this.marketId = marketId;
+    this.outcomes = outcomes;
+  }
+
+  nextRun = (marketBook: MarketOrderBook, timestamp: BigNumber): ZeroXPlaceTradeDisplayParams[] => {
+    const marketOutcomesBook: MarketOutcomesBook = this.outcomes.reduce((p, o) => ({...p, [o]: config}), {})
+    const currentConfig = JSON.parse(JSON.stringify(marketOutcomesBook));
```
robert-e-davidson3: Comment explaining this? Also it needs to add type information since it's been destroyed.

packages/augur-tools/src/flash/orderbook-shaper.ts
```diff
@@ -0,0 +1,130 @@
...
+
+export class OrderBookShaper {
+  marketId: string = "";
+  orderSize: number = 100;
+  outcomes: number[] = [];
+  maxPrice: BigNumber = new BigNumber(1);
+  minPrice: BigNumber = new BigNumber(0);
+  numTicks: BigNumber = new BigNumber(100);
+  numOutcomes: 3;
+  fiveMinutesInSeconds = new BigNumber(18000);
```
robert-e-davidson3:
```suggestion
  readonly fiveMinutesInSeconds = new BigNumber(18000);
```

Or make it a `const` global in this file, maybe?

packages/augur-tools/src/flash/scripts.ts
bthaile: don't want to deal with promises. it's a fire and forget.

packages/augur-tools/src/flash/orderbook-shaper.ts
bthaile: removed currentConfig, it's not needed.

robert-e-davidson3 (approved):

PR #5676, opened 2020-01-23
wait for zeroX orders to get populated before creating them

bthaile:
wait for zeroX orders to get populated before creating them.

don't fund if account has enough balance

### DIFF
--- a/packages/augur-tools/src/flash/scripts.ts
+++ b/packages/augur-tools/src/flash/scripts.ts
@@ -853,7 +853,7 @@ export function addScripts(flash: FlashSession) {
         name: 'userAccount',
         abbr: 'u',
         description:
-          'User account to create orders, if not provider contract owner is used',
+          'User account to create orders, if not provided then contract owner is used',
       },
       {
         name: 'meshEndpoint',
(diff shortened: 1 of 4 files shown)


packages/augur-tools/src/flash/scripts.ts
```diff
@@ -846,45 +846,55 @@ export function addScripts(flash: FlashSession) {
...
-        description: 'Market ids separated by commas for multiple to create orders and maintain order book, ie 0x122,0x333,0x4444'
+        description:
+          'Market ids separated by commas for multiple to create orders and maintain order book, ie 0x122,0x333,0x4444',
       },
       {
         name: 'userAccount',
         abbr: 'u',
-        description: 'User account to create orders, if not provider contract owner is used'
+        description:
+          'User account to create orders, if not provider contract owner is used',
```
robert-e-davidson3:
```suggestion
          'User account to create orders, if not provided then contract owner is used',
```

packages/augur-tools/src/libs/contract-api.ts
```diff
@@ -714,7 +714,7 @@ export class ContractAPI {
   async fundSafe(safe=undefined) {
     safe = safe || await this.getOrCreateSafe();
 
-    await this.faucet(new BigNumber(1e21));
+    await this.faucet(new BigNumber(1e21), safe);
     await this.transferCash(safe, new BigNumber(1e21));
```
robert-e-davidson3:
```suggestion
```

Can transferring be removed now that it's fauceting directly to the safe?

robert-e-davidson3 (approved):

PR #5678, opened 2020-01-23
only fund gnosis safe if it needs it

robert-e-davidson3:


bthaile (approved):
Lgtm, I’ll merge this in my PR when this gets on master

PR #5726, opened 2020-01-28
Deployment verification

nuevoalex:


bthaile (approved):
nice constants file to organize the contracts. lgtm

packages/augur-core/source/libraries/DeploymentVerifier.ts
```diff
@@ -0,0 +1,161 @@
...
+        if (error) return error;
+
+        error = await this.verifyInternalContracts();
+        if (error) return error;
+
+        error = await this.verifyExternalAddresses();
+        if (error) return error;
+
+        error = await this.verifyInitializations();
+
```
robert-e-davidson3, line 58:
```suggestion
if (error) return error;

```

packages/augur-core/source/libraries/DeploymentVerifier.ts
```diff
@@ -0,0 +1,161 @@
...
+		const abi: AbiFunction = {"constant":true,"inputs":[],"name":"augur","outputs":[{"internalType":"contract IAugur","name":"","type":"address"}],"payable":false,"stateMutability":"view","type":"function"}
+		const result = await this.localCall(abi, [], options.sender)
+		return <string>result[0]
+    }
+    
+    public augurTrading_ = async (options?: { sender?: string }): Promise<string> => {
+		options = options || {}
+		const abi: AbiFunction = {"constant":true,"inputs":[],"name":"augurTrading","outputs":[{"internalType":"contract IAugurTrading","name":"","type":"address"}],"payable":false,"stateMutability":"view","type":"function"}
+		const result = await this.localCall(abi, [], options.sender)
+		return <string>result[0]
```
robert-e-davidson3, line 160: I'm not sure you need to specify `<string>` here since it's implied by the function's return type. But if you do then we've standardized on `result[0] as string` instead.

nuevoalex: ah this is just copying the format used by the GenericContractinterfaces

PR #5728, opened 2020-01-28
cache timestamp in DB.ts

robert-e-davidson3:
also cleanup


pgebheim (changes requested):
My first thought here is that this should really just be handled in memory...  but regardless there's lots of merge conflicts now because this is a week old.

Lets discuss tomorrow

PR #5818, opened 2020-01-31
created additional trading script with a few more parameters

bthaile:
getting this error in flash when trying to take existing orders. orders are returned from zeroX getter. trade blows up

trading script. there are some errors, prob need to increase time between trade but will be good for testing. 

found bug and fixed but now getting: 

image

seems there is a conversion issue from zeroX order values being passed into placeTrade
```
getter: getZeroXOrders called at 1580763743394: 21.176ms
Take Order 100 @ 0.55
getter: getZeroXOrders called at 1580763743430: 15.084ms
[]Logged orders
```


packages/augur-tools/src/flash/scripts.ts
```diff
@@ -1076,44 +1084,102 @@ export function addScripts(flash: FlashSession) {
...
     async call(this: FlashSession, args :FlashArguments) {
       const skipFaucet = args.skipFaucet as boolean;
-      let repeats = Number(args.repeats as string) || 1;
-      const max = Number(args.max as string) || 10;
-      const wait = Number(args.wait as string) || 1;
+      const address = args.userAccount ? String(args.userAccount) : null;
+      let marketId = args.market ? String(args.market) : null;
+      let limit = args.limit ? Number(args.limit) : 86400000; // go for a really long time
+      const direction = args.direction ? String(args.direction) : 'bid'
+      const outcome = args.outcome ? Number(args.outcome) : 2;
```
robert-e-davidson3: If args.outcome is zero then it'll default to 2.

packages/augur-tools/src/flash/scripts.ts
```diff
@@ -1076,44 +1084,102 @@ export function addScripts(flash: FlashSession) {
...
-      for (; repeats > 0; repeats--) {
-        let orders = flattenZeroXOrders(await user.getOrders(market.id));
-        if (orders.length > max) {
-          orders = orders.slice(0, max)
+      const market = marketId ? markets.find(m => m.id === marketId) : markets[0];
+
+      console.log('market props')
+      const orderType = direction === 'bid' || direction === 'buy' ? '0' : '1';
+      let loop = true;
+      while (loop) {
```
robert-e-davidson3: Why not use a for-loop?

bthaile: Could do that. Initially thought it would loop until cancelled then added the large limit value and keep the loop. 
For loop would be more readable. Will change

packages/augur-tools/src/flash/scripts.ts
```diff
@@ -1076,44 +1084,102 @@ export function addScripts(flash: FlashSession) {
...
+                )
+              : orders.sort((a, b) =>
+                  new BigNumber(a.price).minus(new BigNumber(b.price)).toNumber()
+                );
+
+          const order = sortedOrders[0];
+          console.log('Take Order', order.amount, '@', order.price);
+          const params: NativePlaceTradeDisplayParams = {
+            market: market.id,
+            direction: Number(direction) as 0 | 1,
```
robert-e-davidson3, line 1166: Isn't `direction` going to be `bid` or `ask`, not a number? Maybe use `orderType` here?

bthaile: Sure can use orderType since bid or ask can be passed in.  I’ll make that change

packages/augur-tools/src/libs/contract-api.ts
```diff
@@ -274,41 +276,10 @@ export class ContractAPI {
...
-        takerFee: order.takerFee,
-        expirationTimeSeconds: order.expirationTimeSeconds,
-        salt: order.salt,
-        makerAssetData: order.makerAssetData,
-        takerAssetData: order.takerAssetData,
-        makerFeeAssetData: order.makerFeeAssetData,
-        takerFeeAssetData: order.takerFeeAssetData,
-      });
-      signatures.push(order.signature);
+      this.augur.placeTrade(order);
```
robert-e-davidson3: The promises here aren't being handled. This may be why there are unhandled exception errors.

bthaile: I was thinking fire and forget. Does it need to be await’ed?

robert-e-davidson3: I'd bundle the promises together and return Promise.all(...) on them. The caller of this function can decide if they want to await or fire-and-forget. For automated tests it's especially good to await because you want to make sure the state is where it should be when you verify it.

bthaile: Ok, I’ll make that change, just in case tests pick this up. Now that ai think contract-api prob already has this functionality. Since I changed it to use augur placeTrade

packages/augur-tools/src/flash/scripts.ts
```diff
@@ -1076,44 +1084,100 @@ export function addScripts(flash: FlashSession) {
...
+            outcome: Number(outcome) as 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7,
```
robert-e-davidson3, line 1167: Don't need `Number(...)` since `outcome` is already a number here.

packages/augur-tools/src/flash/scripts.ts
```diff
@@ -1076,44 +1084,100 @@ export function addScripts(flash: FlashSession) {
...
       const markets = (await user.getMarkets()).markets;
-      const market = markets[0];
-
-      for (; repeats > 0; repeats--) {
-        let orders = flattenZeroXOrders(await user.getOrders(market.id));
-        if (orders.length > max) {
-          orders = orders.slice(0, max)
+      const market = marketId ? markets.find(m => m.id === marketId) : markets[0];
+
+      const direction = orderType === 'bid' || orderType === 'buy' ? '0' : '1';
```
robert-e-davidson3, line 1148: Could make `direction` into a number here so you don't need to cast it later. Would want to change line 1154 to match too.

robert-e-davidson3 (approved):

PR #5819, opened 2020-01-31
fixed gnosis safe batch tx issues

joeykrug:

PR #5820, opened 2020-01-31
minor Gnosis fixes

andrewdamelio:
- update Gnosis.ts, remove interval checks
- Update to use handleGnosis event instead of onNewBlock in UI.
- better support for switching metamask accounts
- handle SafeAlreadyExistsException infinite load modal

@joeykrug I also removed all the set intervals and using the block event as intended. Can you double check your fix for #5819 still works with this PR?

### DIFF
--- a/packages/augur-sdk/src/api/Gnosis.ts
+++ b/packages/augur-sdk/src/api/Gnosis.ts
@@ -207,11 +207,18 @@ export class Gnosis {
       if (error.exception && error.exception.indexOf('SafeAlreadyExistsException') === 0) {
         const restoredAddress = error.exception.match(/0x[a-fA-F0-9]{40}/)[0];
 
+        const status = await this.getGnosisSafeDeploymentStatusViaRelay({
+          owner,
+          safe: ethUtil.toChecksumAddress(restoredAddress)
+        });
+
+        this.safesToCheck = this.safesToCheck.filter(r => restoredAddress !== r.safe);
         this.safesToCheck.push({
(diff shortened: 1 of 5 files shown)


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). 🎇

**https://peek.run/m96hxlo**

-------

joeykrug:
> Can you double check your fix for #5819 still works with this PR?

It won't, b/c it used setInterval and just made the interval longer

> Still looking into why sometimes the safe ends up in created state after it has already been set as available

If you fix this, then my setInterval longer length isn't needed, it has like a race condition or something and so after the AVAILABLE log it'll sometimes occasionally log an additional CREATED which makes that thing on the market pop up that doesn't let you trade (says waiting for safe)

andrewdamelio:
> > Can you double check your fix for #5819 still works with this PR?
> 
> It won't, b/c it used setInterval and just made the interval longer
> 
> > Still looking into why sometimes the safe ends up in created state after it has already been set as available
> 
> If you fix this, then my setInterval longer length isn't needed, it has like a race condition or something and so after the AVAILABLE log it'll sometimes occasionally log an additional CREATED which makes that thing on the market pop up that doesn't let you trade (says waiting for safe)

Ok I'm looking into that last issue right now - once i'm confident this PR fixes all these gnosis race issues i'll ping you for a QA if you don't mind.

pgebheim (approved):

PR #5836, opened 2020-01-31
hot loading fixes again

JohnDanz:
fixed hotloading issue, moved closing of loading modal into the interval check for loading order book, this should make sure we dont overload the rendering. we need to figure out a more elegant solution in the future

### DIFF
--- a/packages/augur-ui/src/modules/market-charts/components/market-charts-pane/market-charts-pane.tsx
+++ b/packages/augur-ui/src/modules/market-charts/components/market-charts-pane/market-charts-pane.tsx
@@ -24,7 +24,6 @@ interface MarketChartsPaneProps {
   toggle: Function;
   tradingTutorial?: boolean;
   isMarketLoading?: boolean;
-  canHotload?: boolean;
   orderBook: OutcomeOrderBook;
 }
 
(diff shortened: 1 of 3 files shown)


JohnDanz:
the real issue here is we are trying to force too many re-renders at once, or so it seems. 

What this does is give sufficient time for the dom to properly mount before we trigger the highcharts. By not closing the marketLoading modal until we have triggered our second call to getOrderbook we are ensuring that everything has had time to properly render and mount the dom. Until the modal is closed, we don't have the price history chart render and therefor there is no more severe hang. This only effects direct navigation/hotloading of a market that has price history, any other situation will have an immediate response and load of the price history chart. This is overall a temporary solution to what seems to be a larger issue we will want to address post v2 around efficient re-renders. 


To reproduce the issue:

go to a market, create trade history on the market.

refresh the page. This should no longer hang for over a minute in what seems to be an infinite re-render loop.

on master the same steps should result in a significant render hang of close to a minute before the market page finally loads if it loads at all.

ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). 🎯

**https://peek.run/kq3h77z**

-------

pgebheim (approved):

PR #5837, opened 2020-01-31
added retries if we get 0 0x orders

joeykrug:


packages/augur-sdk/src/api/ZeroX.ts
```diff
@@ -207,6 +207,11 @@ export class ZeroX {
         response = undefined;
       }
     }
+    if (response.ordersInfos.length < 1) {
+      console.log("Mesh retrying to fetch orders");
+      await sleep(3000);
```
pgebheim: I think we should use a smaller timeout here for the first N tries (some limited back off), just because I think that generally we'll get a connection to a bootnode relatively fast and we don't wanna arbitrarily slow down the fetching of the initial order book.

pgebheim:
```suggestion
      await sleep(tries < 12 ? 1000 : 250);
```

pgebheim (changes requested):

pgebheim (approved):

PR #5908, opened 2020-02-04
more liquidity in canned markets + cleanup

robert-e-davidson3:


bthaile (approved):
small changes, is there more? 
approved

PR #5915, opened 2020-02-05
Convert timestamps on MarketData to numbers

pgebheim:
Includes some other related conversions


bthaile (approved):
lgtm, works in UI. fixed a trade history chart issue, nice!

PR #5953, opened 2020-02-05
emit controller new block event for MarketDB to process markets and u…

bthaile:
…pdate reporting state for none tx event transitions.

only partial fix. with this fix SDK does update market reporting state when event expiration passes. 

Still need to emit MarketUpdated event for the UI to get the updated market.


pgebheim:
Closing in lieu of:
https://github.com/AugurProject/augur/pull/5955

PR #5981, opened 2020-02-06
Liquidity Filter Updates

andrewdamelio:
- don't include orders that expire within 70s
- Add recently invalid markets to Recently Depleted Liquidity (but always show at the bottom of market lists)
- previous PR feedback updates
- ~check orders every 3mins instead of on order/cancel/fill~ In another PR


nuevoalex (approved):
👍

PR #6021, opened 2020-02-07
Add network gnosis zerox endpoints

bthaile:
add zerox and gnosis url to network configuration in augur-core. 
add new firehose flash script to create X orders in bursts.


pgebheim (approved):

pgebheim (changes requested):
We're asking for the network in scripts and  at the top level.

Instead let's include the  network on thesegs

packages/augur-tools/src/flash/scripts.ts
```diff
@@ -85,10 +80,7 @@ export function addScripts(flash: FlashSession) {
       flash.provider = this.makeProvider(this.network);
       const networkId = await this.getNetworkId(flash.provider);
       flash.contractAddresses = getAddressesForNetwork(networkId as NetworkId);
-      const mesh = args.meshEndpoint as string || undefined;
-      const endpoint = 'ws://localhost:60557';
-      const meshEndpoint = mesh ? mesh : endpoint;
-      await flash.ensureUser(this.network, useSdk, true, null, useZeroX ? meshEndpoint : undefined, useZeroX ? true : false);
+      await flash.ensureUser(this.network, useSdk, true, null, useZeroX, useZeroX);
```
robert-e-davidson3: Looks like the original had a bug. The final parameter is `useGnosis`.

robert-e-davidson3 (approved):

pgebheim (approved):

PR #6043, opened 2020-02-09
order form clear on submit.

bthaile:
clear order form when submit order. order form doesn't do anything with error even if one occurs.

no has to keep track of order so use null when calling `updatePendingOrderStatus` if there is an error.  doesn't look like hash is used to find the order anyway. only for verifying it's confirmed, which causes an error anyway. good to set it to null. 

I put `work-in-progress` on this b/c UI might need to disable order form submit button if user has order awaiting signing. discussion is going now, in #dev channel in discord.


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). 💚

**https://peek.run/36wh26e**

-------

andrewdamelio (approved):

PR #6053, opened 2020-02-10
fix category count bug

### LINKED ISSUE #5072: Categories on Market trading pg are not displaying correctly
On V2 finance shows 0 markets, but if i click it....

but if i click it, i see this...

same for crypto categories

andrewdamelio:
closes #5072


bthaile (approved):
works well, lgtm

PR #6056, opened 2020-02-10
a few flash tweaks, add missing await

bthaile:
1. add missing await, 
2. remove invalid orders for creating orders, it hides markets b/c of invalid filter.
3. increase default expiration from 3 min to 5 min.


robert-e-davidson3 (commented):
lgtm

robert-e-davidson3 (approved):

PR #6064, opened 2020-02-10
Use a queue to manage the signing order for safes

pgebheim:


justinbarry (approved):
LGTM. Still needs to be tested in an actual UI.

pgebheim:
Tested this in the actual UI:

1) Double click
2) Sign one order
3) Wait for it to fill
4) sign the other
5) txn fails (good!)
6) try a Faucet and it works
7) try another trade and it works

PR #6066, opened 2020-02-11
update heuristics and fix faucet

nuevoalex:


packages/augur-tools/src/libs/contract-api.ts
```diff
@@ -557,14 +557,17 @@ export class ContractAPI {
   }
 
   async faucet(attoCash: BigNumber, account?: string): Promise<void> {
-    // NOTE: Specifying account only works if its signer is available. So, only useful for gnosis.
-    account = account ||  await this.augur.getAccount();
-    let balance = await this.getCashBalance(account);
+    const realAccount = await this.augur.getAccount();
+    account = account || realAccount;
+    let balance = await this.getCashBalance(realAccount);
```
bthaile: why use `realAccount` here if account is passed in?

bthaile (approved):
lgtm

PR #6092, opened 2020-02-11
check liquidity every 3mins

andrewdamelio:
Only sync orderBook every 3mins instead of on ZeroX events.

### DIFF
--- a/packages/augur-sdk/src/state/db/MarketDB.ts
+++ b/packages/augur-sdk/src/state/db/MarketDB.ts
@@ -35,6 +35,8 @@ interface LiquidityResults {
   [liquidity: number]: number;
 }
 
+let liquidityCheckInterval = null;
+
 /**
  * Market specific derived DB intended for filtering purposes
  */
@@ -65,10 +67,27 @@ export class MarketDB extends DerivedDB {
       'MarketMigrated': this.processMarketMigrated,
     };
 
-    this.augur.events.subscribe('DB:updated:ZeroXOrders', (orderEvents) => this.syncOrderBooks([orderEvents.market]));
-    this.augur.events.subscribe('DB:get:ZeroXOrders', (markets) => this.syncOrderBooks(markets));
+    this.augur.events.subscribe('DB:updated:ZeroXOrders', (orderEvents) => this.markMarketLiquidityAsDirty(orderEvents.market));
     this.augur.events.subscribe(SubscriptionEventName.NewBlock, this.processNewBlock);
     this.augur.events.subscribe(SubscriptionEventName.TimestampSet, this.processTimestampSet);
+
+    // Don't call this interval during tests
+    if (process.env.NODE_ENV !== 'test') {
+      if (!liquidityCheckInterval) {
+        // call recalc liquidity every 3mins
+        const THREE_MINS_IN_MS = 180000;
+        liquidityCheckInterval = setInterval(async () => {
+          const marketsToCheck = await this.stateDB.Markets
+          .filter((order) => order.liquidityDirty)
+          .toArray();
+
+          const marketIdsToCheck = marketsToCheck.map(m => m.market);
+          if (marketIdsToCheck.length > 0) {
+            this.syncOrderBooks(marketIdsToCheck);
+          }
+        },THREE_MINS_IN_MS);
+      }
+    }
   }
 
   syncFTS = async (): Promise<void> => {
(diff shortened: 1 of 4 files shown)


packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -65,10 +67,19 @@ export class MarketDB extends DerivedDB {
...
     this.augur.events.subscribe(SubscriptionEventName.NewBlock, this.processNewBlock);
     this.augur.events.subscribe(SubscriptionEventName.TimestampSet, this.processTimestampSet);
+
+    // Don't call this interval during tests
+    if (process.env.NODE_ENV !== 'test') {
+      if (!liquidityCheckInterval) {
+        // recalc liquidity every 3mins
+        const THREE_MINS_IN_MS = 180000;
+        liquidityCheckInterval = setInterval(() => {
+          this.syncOrderBooks([]);
```
nuevoalex: This is syncing every single market every 3 minutes. I think we should also still be listening to those events and marking which markets need to have their liquidity recalculated. Then when this gets called pass in those market ids.

andrewdamelio: @nuevoalex ok that makes more sense, easy enough 👍

packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -113,13 +132,26 @@ export class MarketDB extends DerivedDB {
...
         doc['market'] = marketId;
+        doc['liquidityDirty'] = false;
         documents.push(doc);
       }
     }
 
     await this.bulkUpsertDocuments(documents);
   }
 
+  async markMarketLiquidityAsDirty(marketId: string) {
```
nuevoalex: What I would do instead actually is just keep this in memory and store the marketIds as a Set

packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -65,10 +67,27 @@ export class MarketDB extends DerivedDB {
...
     this.augur.events.subscribe(SubscriptionEventName.NewBlock, this.processNewBlock);
     this.augur.events.subscribe(SubscriptionEventName.TimestampSet, this.processTimestampSet);
+
+    // Don't call this interval during tests
+    if (process.env.NODE_ENV !== 'test') {
+      if (!liquidityCheckInterval) {
+        // call recalc liquidity every 3mins
+        const THREE_MINS_IN_MS = 180000;
+        liquidityCheckInterval = setInterval(async () => {
+          const marketsToCheck = await this.stateDB.Markets
```
nuevoalex: Noted below but if the market marketIds are just stored in memory as a set you can just pass those in here and destroy that set after.

packages/augur-sdk/src/state/db/MarketDB.ts
andrewdamelio: 👍  updating

nuevoalex (commented):
Overall good! Think the dirty market handling can be done in a less expensive way

packages/augur-sdk/src/state/db/MarketDB.ts
andrewdamelio: 👍  updating

packages/augur-sdk/src/state/db/MarketDB.ts
```diff
@@ -65,10 +68,24 @@ export class MarketDB extends DerivedDB {
...
-    this.augur.events.subscribe('DB:get:ZeroXOrders', (markets) => this.syncOrderBooks(markets));
+    this.augur.events.subscribe('DB:updated:ZeroXOrders', (orderEvents) => this.markMarketLiquidityAsDirty(orderEvents.market));
     this.augur.events.subscribe(SubscriptionEventName.NewBlock, this.processNewBlock);
     this.augur.events.subscribe(SubscriptionEventName.TimestampSet, this.processTimestampSet);
+
+    // Don't call this interval during tests
+    if (process.env.NODE_ENV !== 'test') {
+      if (!liquidityCheckInterval) {
+        // call recalc liquidity every 3mins
+        const THREE_MINS_IN_MS = 180000;
```
bthaile, line 79: why not use block number and just estimate number of blocks as being 3 minutes. then don't need to use setInterval.

could hook into `processNewBlock` and then check liquidity every 12 blocks

nuevoalex: Block time may change. the time interval is more accurate to our goal really

nuevoalex (approved):
👍

PR #6108, opened 2020-02-12
Add missing parseLogs call in blockstream

justinbarry:


packages/augur-sdk/src/state/db/AbstractTable.ts
```diff
@@ -63,7 +63,7 @@ export abstract class AbstractTable {
 
     const id = [];
     for (let idField of this.idFields) {
-      id.push(document[idField]);
+      id.push(`${document[idField]}`);
```
bthaile, line 66: Is this just fancy toString()

bthaile (approved):
Lgtm

PR #6189, opened 2020-02-14
addresses and upload block numbers is now one file

robert-e-davidson3:


robert-e-davidson3:
This is failing because the pop docker images need to be rebuilt to use `environments.json`.

robert-e-davidson3:
We're going to use SDKConfiguration config files so this is probably not going to be used.

pgebheim:
Being included in another pr by @robert-e-davidson3 -- closing

PR #6203, opened 2020-02-14
add sending queue for relayer

andrewdamelio:


nuevoalex (changes requested):
I think this would work more clearly if we used the AsyncQueue class for our queue instead as we do in other code locations like `EthersProvider.ts` or `ContractDependenciesGnosis.ts`. That way the SIG_QUEUE could be an actual queue that recieves tasks which include the `{ tx: relayTransaction, txMetadata }` data and will queue/process them as expected without any logic needed on our part for that. In the existing implementation for example if 2 items queue up the second one will get dropped.

andrewdamelio:
@nuevoalex with the AsyncQueue class, am i able to clear out and abort the rest of the queue, if i need to (relayer down)?

nuevoalex:
@andrewdamelio Yup! See the `kill` method here: https://caolan.github.io/async/v3/docs.html#QueueObject

packages/contract-dependencies-gnosis/src/ContractDependenciesGnosis.ts
```diff
@@ -49,6 +53,49 @@ export class ContractDependenciesGnosis extends ContractDependenciesEthers {
...
+            TransactionStatus.RELAYER_DOWN,
+          );
+        } else if (error === TransactionStatus.FAILURE) {
+          this.onTransactionStatusChanged(
+            request.txMetadata,
+            TransactionStatus.FAILURE,
+          );
+        }
+
+        await this._relayQueue.kill();
```
nuevoalex: this should only be done for the RELAYER_DOWN case I believe

nuevoalex (commented):
LGTM so far!

packages/contract-dependencies-gnosis/src/ContractDependenciesGnosis.ts
andrewdamelio: Correct -- my mistake - was just lazy testing :/

packages/contract-dependencies-gnosis/src/ContractDependenciesGnosis.ts
```diff
@@ -49,6 +53,50 @@ export class ContractDependenciesGnosis extends ContractDependenciesEthers {
...
+        } else if (error === TransactionStatus.FAILURE) {
+          this.onTransactionStatusChanged(
+            request.txMetadata,
+            TransactionStatus.FAILURE,
+          );
+        }
+        throw error;
+      }
+    } else{
+      // If the Relay Service is not being used so we'll execute the TX directly
```
robert-e-davidson3:
```suggestion
      // The Relay Service is not being used so we'll execute the TX directly
```

packages/contract-dependencies-gnosis/src/ContractDependenciesGnosis.ts
```diff
@@ -49,6 +53,50 @@ export class ContractDependenciesGnosis extends ContractDependenciesEthers {
...
+          );
+        } else if (error === TransactionStatus.FAILURE) {
+          this.onTransactionStatusChanged(
+            request.txMetadata,
+            TransactionStatus.FAILURE,
+          );
+        }
+        throw error;
+      }
+    } else{
```
robert-e-davidson3:
```suggestion
    } else {
```

nuevoalex (approved):
👍

PR #6218, opened 2020-02-15
Bound wit me

justinbarry:


packages/augur-sdk/src/state/db/BaseSyncableDB.ts
```diff
@@ -0,0 +1,92 @@
...
+    this.eventName = eventName;
+
+    db.notifySyncableDBAdded(this);
+    this.rollingBack = false;
+  }
+
+  async addNewBlock(blocknumber: number, logs: ParsedLog[]): Promise<number> {
+    // don't do anything until rollback is complete. We'll sync back to this block later
+    if (this.rollingBack) {
+      return -1;
```
bthaile, line 35: does this -1 make it to `provider.getLogs`? What the the source of the hexify issue? I think it might have gotten fixed recently

pgebheim: The case where that happens is just slightly different and was from a case where we'd double fetch the current block from the eth node, and sometimes fail. On failure it would return -1 and we'd roll forward with that as current block.

Also don't love the sentinel value here not sure if that can be changed.

justinbarry: The value isn't used. Could just be an early return.

pgebheim (approved):

Issue #6198, opened 2020-02-14
Rep balance errors


matt-bullock:
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/cfa42aff-ddc3-49b5-8d06-ea9496684c48)

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/732b5d7c-c734-47b5-a205-8aecf7f2fdde)

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/544a7a02-351f-4496-87ab-aa702f5a43b8)

there's an error in rep balances when supporting an outcome:
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/43affa46-3fca-4cf3-a6b0-05b81a5473d2)

similar issue happening with disputing an outcome
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/f0773350-c256-455f-9e30-550046dab878)

bthaile:
I have a PR to fix this issue of balances not updating.

matt-bullock:
Check through and DR and OR appears to be fixed.

Disputing is still having issues with adding pre-filled stake to outcomes. Some markets are working fine but below is an example where the amount is coming out of the users available rep balance and is showing on the market card under pre-filled stake but it's not showing up in the users 'My Total REP Staked'

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/8f0120d8-e270-4f06-89fa-55bb96cf7f3e)

bthaile:
Ok, could be UI isnt refreshing user data after a dispute or there is some error preventing the updating of user data. I can look.

bthaile:
I'm not seeing this issue. @matt-bullock can you confirm

matt-bullock:
Checked through all scenarios and rep balances are all being deducted/adding up correctly

PR #6220, opened 2020-02-15
propagate non-market updated events

bthaile:
I noticed that non-market updates aren't making it to the UI. 
Also Controller was trying to load markets for 0x00.... market address, I filtered that address out so fake market updated events aren't being emitted to the UI. 

Also noticed that TokenType enumerator indexes were off, removed tokens that were not the same as Augur contract. 

Also ignore empty marketIds values in log-handers.

how to repro issue:
1. start Ui
2. login
3. nav to dispute page and buy participation tokens, user's participation token balance never updates. 

- with PR it updates a little bit after the buy transaction is successful.

https://github.com/AugurProject/augur/issues/6198


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). 😀

**https://peek.run/zr5hjeo**

-------

packages/augur-sdk/src/state/Controller.ts
```diff
@@ -31,25 +31,31 @@ export class Controller {
...
+    });
+  };
+
+  private allEvents = async (blockNumber: number, allLogs: ParsedLog[]) => {
     // Grab market ids from all logs.
     // Compose applies operations from bottom to top.
     const logMarketIds = fp.compose(
       fp.compact,
       fp.uniq,
+      fp.filter(NULL_ADDRESS),
```
justinbarry: I'm not sure if this will work as expected. Try this:
```suggestion
      fp.filter(item => item === NULL_ADDRESS),
```

packages/augur-sdk/src/state/Controller.ts
```diff
@@ -31,25 +31,31 @@ export class Controller {
...
-      marketIds: logMarketIds
-    });
-
-    this.augur.events.emit(SubscriptionEventName.MarketsUpdated,  {
-      marketsInfo
-    });
-  };
+    if(logMarketIds.length > 0) this.updateMarketsData(logMarketIds as string[]);
+    // emit non market related logs
+    allLogs.forEach(l => l.market === NULL_ADDRESS && this.augur.events.emit(l.name, {...l}));
```
justinbarry: If the filter above works we only need to do this:
```suggestion
    allLogs.forEach(l => this.augur.events.emit(l.name, {...l}));
```

bthaile: that change will double send market updated events

packages/augur-sdk/src/state/Controller.ts
bthaile: my filter does work.

packages/augur-sdk/src/state/Controller.ts
bthaile: because market updated events are sent earlier in the code.

packages/augur-sdk/src/state/Controller.ts
bthaile: I can make the code more explicit. I'll do that

justinbarry (approved):
LGTM

PR #6245, opened 2020-02-18
negative gnosis tests

robert-e-davidson3:
also removed a skipped test since I couldn't get emitted events to pass from sdk to tests


pgebheim (approved):
This looks like a good set of tests. It would indeed be nice to get something more complex in there the nonce race conditions but this set is still good to merge.

robert-e-davidson3:
re-added skipped test because it was changed in master, though still skipped

PR #6248, opened 2020-02-18
Kovan Deploy 2020 02 17

pgebheim:


packages/augur-tools/src/flash/flash.ts
```diff
@@ -160,60 +160,76 @@ export class FlashSession {
...
+          console.log(`Safe ${safe}: ${safeStatus}`);
+          this.user.augur.setGasPrice(new BigNumber(90000));
+          this.user.setGnosisSafeAddress(safe);
+          this.user.setUseGnosisSafe(true);
+          this.user.setUseGnosisRelay(true);
+        } else if (approveCentralAuthority) {
+          await this.user.approveCentralAuthority();
+        }
+      } catch (e) {
+        throw e;
```
robert-e-davidson3, line 208: is this necessary?

packages/augur-tools/src/flash/flash.ts
```diff
@@ -160,60 +160,76 @@ export class FlashSession {
...
-      // at another time.
-      await this.user.augur.on(
-        SubscriptionEventName.NewBlock,
-        this.sdkNewBlock
-      );
+      console.log('------ Starting Server -------')
+      console.log('Syncing Enabled: Starting API Server')
+      try {
+        await this.user.augur.connector.connect(config);
+        this.api = (this.user.augur.connector as Connectors.SingleThreadConnector).api;
```
robert-e-davidson3, line 219: maybe reference the `connector` variable directly, since it's local?

robert-e-davidson3 (approved):

packages/augur-tools/src/flash/flash.ts
pgebheim: Currently when you run scripts that call ensureUser it tosses random logs into the console and isn't really understandable.

Seems better to try to make this pretty while in here.

Issue #6144, opened 2020-02-12
Betting UI - Search Input


JohnDanz:
https://www.figma.com/file/42QoTLQcbObu7esoTPjWRv/Augur-Bet---Design-System?node-id=561%3A4371

![Screen Shot 2020-02-12 at 1.11.37 PM.png](https://images.zenhubusercontent.com/57155267bf93871ccd99a48b/ca9535dc-8c93-4db3-a77d-9120abe2ad57)

Update the style of the search bar -- don't worry about the added search results window that appears to be part of this for now.

PR #6250, opened 2020-02-18
Betting UI: search input

rubyweb3:
https://github.com/AugurProject/augur/issues/6144

Focused:
Screen Shot 2020-02-18 at 5 09 59 PM

Inactive:
Screen Shot 2020-02-18 at 5 09 53 PM


packages/augur-ui/src/modules/filter-sort/components/filter-search.styles.less
```diff
@@ -33,22 +33,22 @@
...
     .text-14;
 
-    border: none;
-    color: var(--color-primary-text);
     line-height: normal;
     padding: 0;
     height: @size-20;
     margin: @size-8 0 @size-10 @size-10;
     text-overflow: ellipsis;
+    border: none !important;
```
JohnDanz: no `!important` please

packages/augur-ui/src/modules/filter-sort/components/filter-search.styles.less
```diff
@@ -15,15 +15,15 @@
...
   box-sizing: border-box;
   border-radius: @border-radius-default;
-  color: var(--color-secondary-text);
   display: flex;
   padding: 0;
   transition: min-width @animation-speed-normal @animation-default-type;
   width: 100%;
+  background: var(--color-dark-grey);
+  border: @size-1 solid var(--color-secondary-action);
+  color: var(--color-secondary-text);
```
JohnDanz: don't move these values to the bottom, it makes merging master in later more confusing. Also I don't think it's been enforced for a little bit but we have a linter that prefers alphabetical order for styling

packages/augur-ui/src/assets/styles/_colors.less
```diff
@@ -145,6 +147,7 @@
   --color-page-background-80: fade(#F5F5F8, 80%);
   --color-dark-grey: #D1D1D9;
   --color-dark-grey-80: fade(#D1D1D9, 80%);
+  --color-border-grey: #DEDEE4;
```
JohnDanz: I talked to yanel, the labels shouldn't be trusted and you have to look at the square. the square for `color-border-grey` is actually `#D1D1D9` which is something we have already with `--color-dark-grey` In general I want to avoid adding any new color labels since the trading UI already uses so many compared to betting that we shouldn't need to add more.

packages/augur-ui/src/modules/filter-sort/components/filter-search.styles.less
JohnDanz: Don't use `!important` please. also as I mentioned above, please don't re-arrange definitions since it causes more merge conflicts that take longer to resolve/we might miss things that are getting double defined.

rubyweb3: input is another small search in Search. It won't have any border.

rubyweb3: Updated. Thank you

JohnDanz (approved):

PR #6274, opened 2020-02-18
use torus provider as default

andrewdamelio:
- update torus to v1
- if window.web3 is not available, use torus as the web provider

### DIFF
--- a/packages/augur-ui/src/modules/app/actions/init-augur.ts
+++ b/packages/augur-ui/src/modules/app/actions/init-augur.ts
@@ -24,6 +24,8 @@ import {
   MODAL_TEST_BET,
   MODAL_TUTORIAL_INTRO,
   SIGNIN_LOADING_TEXT,
+  NETWORK_IDS,
+  NETWORK_NAMES,
 } from 'modules/common/constants';
 import { windowRef } from 'utils/window-ref';
 import { AppState } from 'store';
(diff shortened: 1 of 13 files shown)


ghost (commented):
This pull request has been automatically deployed to [FeaturePeek](https://dashboard.featurepeek.com). 🍪

**https://peek.run/zr5hj14**

-------

packages/augur-ui/src/modules/auth/actions/login-with-fortmatic.ts
```diff
@@ -10,13 +10,13 @@ import { getNetworkId } from 'modules/contracts/actions/contractCalls';
 import { windowRef } from 'utils/window-ref';
 import { AppState } from 'store';
 
-const getFormaticNetwork = (networkId: string): false | string   => {
+const getFormaticNetwork = (networkId: string): string   => {
```
bthaile: Can this be moved out to a helper to convert networkId to network name

packages/augur-ui/src/modules/app/actions/init-augur.ts
```diff
@@ -214,10 +217,39 @@ export function connectAugur(
       preloadAccount(ACCOUNT_TYPES.TORUS);
     }
 
-    let provider;
-    if (windowApp.web3) {
-      provider = new Web3Provider(windowApp.web3.currentProvider);
+    let provider = null;
+    let networkId = null;
+
+    if (process.env.ETHEREUM_NETWORK === 'kovan') {
```
bthaile: Can this be moved out to helper. Maybe with helper to convert networkId to network name

packages/augur-ui/src/modules/auth/actions/login-with-fortmatic.ts
pgebheim: There's already a mapping for this, yah

pgebheim (commented):
Currently this code is doing the Kovan / Mainnet checks inline -- but I want to point out that the name here for ETHEREUM_NETWORK isn't really a network, its the name of a config.

We are likely going to make a `demo` config that is running against Kovan. We are also likely to test against Rinkeby etc, so we shouldn't hardcode this to kovan, at minimum. I think we should put the network ID on the SDKConfiguration (in network.json / env) for anything except dev nets and then this can just be pulled off the config.

See: `isDevNetworkId` from augur-artifacts/index.ts

bthaile (changes requested):
consolidate logic for translating network/networkIds

bthaile (approved):
lgtm

pgebheim (approved):

PR #6295, opened 2020-02-19
Metamask login prompt bug

andrewdamelio:
- Show MM signing helper modal when we are restoring accounts that happen to be signed out.


bthaile (approved):
lgtm

Issue #6088, opened 2020-02-11
Can't cancel an open order on mobile but can on desktop view


matt-bullock:
Cancel button isn't doing anything

bthaile:
note:
tried testing on safari but didn't get very far, it seems to pass out under the strain. I think we are waiting for a new safari updated.

Chwy5:
closing

PR #6298, opened 2020-02-19
fix-cancel-order

JohnDanz:
https://github.com/augurproject/augur/issues/6088

the issue was a function wasn't setup correctly it appears. must have changed at some point.


bthaile (approved):
lgtm

PR #6303, opened 2020-02-19
clear liquidity state on logout

### LINKED ISSUE #6285: Switching accts From MM to portis doesnt clear out info
Create a new market and provide pre-liquidity using MM.  after Market is created Do NOT sign orders to get created, but switch acts to Portis.  I get notification to sign and approve orders in portis acct for market I created in MM.

I do NOT see the market listed under portfolio pg though under my created markets, so hopefully this is just a notification bug.

andrewdamelio:
fixes #6285


bthaile (approved):
Lgtm

Issue #6228, opened 2020-02-17
creating Pre-liq orders don't show up in portfolio page. 


bthaile:
Portfolio page needs to be refreshed after user creates pre-liq orders from portfolio page. 

1. create market with pre-liq
2. when market is created click on submit orders to create pre-liq orders
3. notice that orders do get created but don't show in open orders section in portfolio page

Chwy5:
this works - closing

PR #6305, opened 2020-02-19
call selector when orders changes

phoebemirman:
https://github.com/AugurProject/augur/issues/6228


bthaile (approved):
looks good, that will trigger the selector when user's open orders redux state changes.

PR #6314, opened 2020-02-20
Server.ts is now a flash command: sdk-server

robert-e-davidson3:


packages/augur-sdk/src/state/create-api.ts
```diff
@@ -157,6 +167,10 @@ export async function createServer(config: SDKConfiguration, client?: Augur, acc
     ...config
   };
 
+  if (!(config.ethereum && config.ethereum.http)) {
```
pgebheim: This isn't actually true -- this is only true if running in server mode (and thus this should be checked outside of createServer).

e.g. for augur in the UI all we need to do is pass in a Provider.

packages/augur-sdk/src/state/create-api.ts
```diff
@@ -108,7 +110,15 @@ export async function createClient(
   ): Promise<Augur> {
 
   const ethersProvider = provider || new EthersProvider( new JsonRpcProvider(config.ethereum.http), 10, 0, 40);
-  const addresses = config.addresses || getAddressesForNetwork(config.networkId);
+  const networkId = config.networkId || await ethersProvider.getNetworkId();
+
+  if (!(config.addresses || networkId)) {
+    throw Error('Config must include addresses or networkId because node did not provide networkId');
+  }
+  if (!(config.gnosis && config.gnosis.http)) {
```
pgebheim:
```suggestion
  if (config.gnosis?.enabled && !config.gnosis?.http) {
```

packages/augur-sdk/src/state/create-api.ts
```diff
@@ -147,7 +157,7 @@ export async function createClient(
 
 export async function createServer(config: SDKConfiguration, client?: Augur, account?: string): Promise<{ api: API, controller: Controller, sync: () => Promise<void> }> {
   console.log('Creating Server');
-  // Validate the config -- check that the syncing key exits and use defaults if not
+  // Validate the config -- check that the syncing key exists and use defaults if not
   config = {
     syncing: {
```
pgebheim, line 175: Should we combine `server` and `syncing` here and just have `syncing` be a part of the server configuration?

packages/augur-sdk/src/state/create-api.ts
pgebheim:
```suggestion
  if (!config?.ethereum?.http) {
```

packages/augur-tools/src/flash/scripts.ts
```diff
@@ -2116,6 +2119,65 @@ export function addScripts(flash: FlashSession) {
...
+        abbr: 'c',
+        description: 'serialized SDKConfiguration JSON',
+      },
+    ],
+    async call(this: FlashSession, args: FlashArguments) {
+      const configFile = args.configFile as string;
+      const config = args.config as string;
+
+      let configuration: SDKConfiguration;
+      if (configFile && config) {
```
pgebheim, line 2149: This is so wonderful. We should probably move this up a level to the global configs so flash can take ANY config file.

```flash -c myConfig.json -n MyNetworkEnvironment create-canned-markets```

Does't need to be in this patch. Thoughts? If you agree we can make a ticket.

File: packages/augur-tools/src/flash/scripts.ts [hunk 5]
```diff
@@ -2116,6 +2119,65 @@ export function addScripts(flash: FlashSession) {
...
+        startWSS: true,
+        ...(configuration.server || {})
+      };
+
+      const httpServer = endpointSettings.startHTTP && runHttpServer(app, endpointSettings);
+      const httpsServer = endpointSettings.startHTTPS && runHttpsServer(app, endpointSettings);
+      const wsServer = endpointSettings.startWS && runWsServer(api, app, endpointSettings);
+      const wssServer = endpointSettings.startWSS && runWssServer(api, app, endpointSettings);
+
+      await awaitUserInput('Running SDK server. Press ENTER to quit:\n');
```
pgebheim: Instead of doing this -- we should probabaly just bind for SIGINT. Hitting any key to exit a running server is weird on any platform.

```suggestion
      process.on('SIGINT', () => {
          httpServer.close();
          httpsServer.close();
          wsServer.close();
          wssServer.close();
      });
```

pgebheim (changes requested):

packages/augur-tools/src/flash/scripts.ts
pgebheim: Oh, this applies for the other cases where we do this too. But don't these things gracefully shut down on SIGINT anyway? Is there some concern that they wont and we need to clean up? Or is this because of interactive mode?

robert-e-davidson3: I did this because it felt wrong to `await` on the final never-ending promise, especially without telling the user what was going on. I agree that binding of SIGINT is a better solution.

packages/augur-sdk/src/state/create-api.ts
robert-e-davidson3: I moved this check to where we actually use `config.ethereum.http`, and only when we don't provide a provider.

packages/augur-sdk/src/state/create-api.ts
robert-e-davidson3: Are `sdk`, `syncing`, and `server` all related then?

packages/augur-tools/src/flash/scripts.ts
robert-e-davidson3: I agree!

https://github.com/AugurProject/augur/issues/6385

packages/augur-tools/src/flash/scripts.ts
robert-e-davidson3: https://github.com/AugurProject/augur/issues/6386

packages/augur-sdk/src/state/create-api.ts
pgebheim: yeah -- except `sdk` is kinda like a client side configuration (speaking TO an sdk server) while `server` is for the server from its config....

we can rename all of this to make it make sense

robert-e-davidson3: Added a bit about this to https://github.com/augurproject/augur/issues/6385.

pgebheim (approved):

PR #6315, opened 2020-02-20
Betting UI - loading ellipse icon

rubyweb3:
https://github.com/AugurProject/augur/issues/6316

Screen Shot 2020-02-20 at 2 32 03 PM (2)


packages/augur-ui/src/modules/common/icons.tsx
```diff
@@ -1150,30 +1150,39 @@ export const DoubleArrows = (
...
+    <svg width="16" height="16" viewBox="0 0 16 16" fill="none">
+      <path
+        d="M7.99998 14.4C11.5346 14.4 14.4 11.5346 14.4 7.99998C14.4 4.46535 11.5346 1.59998 7.99998 1.59998C4.46535 1.59998 1.59998 4.46535 1.59998 7.99998"
+        stroke="url(#paint0_linear)"
+        strokeWidth="1.4"
+        strokeMiterlimit="1.5728"
+        strokeLinecap="round"
+      />
+      <defs>
+        <linearGradient
```
rubyweb3: There are several ways to achieve linearGradient in LoadingEllipse in trading and betting UIs. Other options:
1. We don't need to support linearGradient in LoadingEllipse
2. We can use a different icon in Betting UI.

@JohnDanz What do you think about it?

JohnDanz: I think we should be using CSS to handle this instead of directly defining it in the SVG like we do for other SVGs.

so for this we could do something like

```
. CategoryRow {
> span:nth-of-type(2) {
 > svg LinearGradient > stop {
 stopColor: var(--color-primary-text);
}
```
something along those lines

JohnDanz (changes requested):
I'd prefer to use CSS to style this icon, not checking theme on the icon. Ideally theme checking in JS is done extremely rarely and mainly just for logical changes to view routing or if there are different calculations to be made, it shouldn't be used for styling changes.

packages/augur-ui/src/modules/common/icons.tsx
rubyweb3: Updated. It really works. Thank you.

JohnDanz (approved):

PR #6363, opened 2020-02-20
Update warp sync to use gateway

### LINKED ISSUE #6398: Warp Markets Should be flagged on Derived DB
This can be done by checking if the creator === the WarpSync contract address.

justinbarry:
closes #6398

### DIFF
--- a/packages/augur-sdk/src/state/db/WarpSyncDB.ts
+++ b/packages/augur-sdk/src/state/db/WarpSyncDB.ts
@@ -28,9 +28,6 @@ export class WarpSyncDB extends AbstractTable {
   }
 
   async getMostRecentWarpSync():Promise<WarpSyncDocument | undefined> {
-    console.log('await this.table.toArray()',
-      JSON.stringify(await this.table.toArray()));
-
     return this.table.orderBy('end.number').last();
   }
 }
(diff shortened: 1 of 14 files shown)


packages/augur-sdk/src/state/create-api.ts
```diff
@@ -74,7 +74,7 @@ export function buildSyncStrategies(client:Augur, db:Promise<DB>, provider: Ethe
 
     const endWarpSyncBlockNumber = await warpSyncStrategy.start();
     const staringSyncBlock = Math.max(await (await db).getSyncStartingBlock(), endWarpSyncBlockNumber || uploadBlockNumber);
-    const endBulkSyncBlockNumber = await bulkSyncStrategy.start(staringSyncBlock, currentBlockNumber);
+    const endBulkSyncBlockNumber = await bulkSyncStrategy.start(staringSyncBlock + 1, currentBlockNumber);
```
nuevoalex: this seems wrong. If the upload block number is used for example this will skip the logs from the initial upload.

packages/augur-sdk/src/state/db/WarpSyncDB.ts
```diff
@@ -0,0 +1,36 @@
...
+        begin,
+        end,
+        hash
+      }
+    ]);
+  }
+
+  async getMostRecentWarpSync():Promise<WarpSyncDocument | undefined> {
+    console.log('await this.table.toArray()',
+      JSON.stringify(await this.table.toArray()));
```
nuevoalex: probably want to remove this?

nuevoalex (approved):
One issue I think but otherwise looks good

packages/augur-sdk/src/state/create-api.ts
justinbarry: Love it. Done!

PR #6365, opened 2020-02-21
send events after new block event

nuevoalex:


bthaile (approved):
lgtm, fixes a few issues.

Issue #6227, opened 2020-02-17
MC: Template wrapped value formatted issues


bthaile:
Two issues.
1. the second name isn't underlined 
2. tooltip isn't lined up

![image.png](https://images.zenhubusercontent.com/5c00901a5cb43a45a18aeb7f/2b6ad7ce-d892-4c99-8cf8-6484b145a427)

Issue in the second name is wrapped which is messing up the underline and tooltip location

Chwy5:
fixed closing

PR #6388, opened 2020-02-21
fix word wrapping for yellow underline

phoebemirman:
https://github.com/AugurProject/augur/issues/6227


bthaile (approved):
beautiful, lgtm

Issue #6371, opened 2020-02-21
Scalar template Numeric Range Fixes


Chwy5:
Still need these (finance and crypto).....

For Crypto, and Finance scalar templates can we have it say in the box (instead of the 0 and 1):
Enter Min range value/ Enter Max range value

[x] For NCAA Football Scalar, make the numeric range 0-12
[x] WNBA Scalar 0-36
[x] Olympics Scalar make the range 0-250

-------------------

Chwy5:
looks good closing

PR #6404, opened 2020-02-24
show placeholder text instead of defaulted min/max prices for tempalt…

bthaile:
…es without default values

https://github.com/AugurProject/augur/issues/6371

all scalar templates without defaulted min/max prices should show placeholder text in min/max price inputs.

image

validations work
image


phoebemirman (changes requested):
Screen Shot 2020-02-24 at 9 49 02 AM
seems like some templates are missing the min value?

bthaile:
Falsey is the issue. I need to check for undefined

phoebemirman (approved):

Issue #6392, opened 2020-02-21
Bug: Buy/Sell text in notifications and alerts should match action the user took.


pgebheim:
**Describe the bug**
When placing a BUY or SELL order, or when filling a BUY or SELL order, the notification toast and bell alert show the wrong direction.

**To Reproduce**
1. Go to any market
2. Make a BUY order
3. Check bell alert, it will say "Selling N and 0.X" in the alert.

**Expected behavior**
The direction should match what the user did.

If owner of order creates an ask order to put on the book then get filled, alert/toast should say
`"Selling N at X"` in the alert

If user creates a bid order to take an order on the book then the alert/toast should say
`"Buying N at X"` in the alert

Chwy5:
Still looks incorrect....look at top toast message says sell and look at my position shows Long

![Screen Shot 2020-02-25 at 12.24.02 PM.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/053e51fd-3a99-4211-af3b-7e61a0252bc0)

Chwy5:
fixed - closing

PR #6410, opened 2020-02-24
fix order type in alerts

phoebemirman:
https://github.com/AugurProject/augur/issues/6392

### DIFF
--- a/packages/augur-ui/src/modules/alerts/actions/set-alert-text.ts
+++ b/packages/augur-ui/src/modules/alerts/actions/set-alert-text.ts
@@ -43,6 +43,7 @@ import {
   CREATEYESNOMARKET,
   APPROVE,
   BUY_INDEX,
+  HEX_BUY,
   SELL_INDEX,
   ZERO,
   ONE,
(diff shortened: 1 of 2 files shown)


bthaile (approved):
lgtm, oops have question. see below

packages/augur-ui/src/modules/alerts/actions/set-alert-text.ts
```diff
@@ -61,8 +61,7 @@ function toCapitalizeCase(label) {
 export function getInfo(params: any, status: string, marketInfo: MarketData) {
   const outcome = new BigNumber(params.outcome || params._outcome).toString();
   const outcomeDescription = getOutcomeNameWithOutcome(marketInfo, outcome);
-  let orderType = params.orderType === BUY_INDEX ? BUY : SELL;
-
+  let orderType = convertAttoValueToDisplayValue(createBigNumber(params.orderType)).toNumber() === BUY_INDEX ? BUY : SELL;
```
bthaile: is params.orderType in hex? could just test against `0x01` and `0x00`

bthaile (changes requested):
question, is orderType in hex?

bthaile (approved):
looks amazing, good one

Issue #5812, opened 2020-01-30
Filling multiple orders at once displaying as individual fills and not 1 fill in Filled orders on trading pg and portfolio pg


Chwy5:
i took 3 open orders in 1 trade...the My fills is showing them as 3 separate trades....No Fills should say 3 and when expanded, show the 3 individual trades....

Currently displays it as this....
![Screen Shot 2020-01-30 at 3.03.02 PM.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/e29671c2-e4b5-49b5-ac83-dfa69710043e)

bthaile:
I'm seeing this behavior. If 2 users nibble at my user's existing order

![image](https://user-images.githubusercontent.com/3970376/75185809-5675b480-570c-11ea-9394-9ee7aea6790d.png)

I do see a refresh issue in filled order section of portfolio.

Chwy5:
this is still not working right when selling existing multiple orders

bthaile:
@Chwy5  when you say "when selling existing multiple orders" do you mean your existing order gets filled by multiple orders from other ppl? So you are not taking orders, you created the original order?

Chwy5:
So what happens if there is a 40 bid for 10, a 35 bid for 10, and 30 bid for 10 on the book....and I come along and sell 30 at 30?  Does this show as filled 30 shares at 35 with 3 fills?

bthaile:
this is the scenario I tested. 
User A puts large 1000 share order on the book
User B takes 10 shares in one transaction
then User B takes 20 shares in second transaction
the screenshot is from User A's filled orders. 
this looks correct to me.

![image](https://user-images.githubusercontent.com/3970376/75890253-b860a900-5df3-11ea-9224-5ed1c1a7158e.png)

Chwy5:
so your example is correct....there is a second scenario from the taker's perspective

there may be 200 shares offered at 60 (but those 200 shares are made up of 10 different offers all at the same price).... If as the take I went and filled all of them at once I would expect to see a fill of 1 order of 200 and in the # of Fills see the individual breakdown....otherwise we are going to flood the My Fills section with a bunch of small orders

Chwy5:
fixed-closing

PR #6412, opened 2020-02-24
fix issue where portfolio filled orders doesn't automatically update …

bthaile:
…when user is filled

https://github.com/AugurProject/augur/issues/5812

portfolio filled-orders wasn't updating automatically when user's order gets filled and user is on portfolio page. 

to repro:
have 2 UIs open with different users logged in. MM and portis for example.
have user one create order on market and nav to Portfolio page
have second user take the order, should see filled-orders update when order gets filled. Filling an order takes awhile. just FYI.


phoebemirman (changes requested):
getting this direct loading a market
Screen Shot 2020-02-24 at 12 39 30 PM

bthaile:
updated to default filled orders to [] if no fills exist and/or if user isn't logged in.

phoebemirman (approved):

PR #6414, opened 2020-02-24
remove-connect-modal/options

andrewdamelio:
Chatted with Design, we are going to remove the connectModal since it's not being used.


bthaile (approved):
yea remove code, lgtm

Issue #6281, opened 2020-02-19
Make Min/Max range visable for scalar markets on Mobile


Chwy5:
I cant see the Min/Max range for a scalar market on mobile, Making it not tradeable....

https://cdn.discordapp.com/attachments/378030793899114516/679680262191185920/Screenshot_20200219-082533_Chrome.jpg

matt-bullock:
done. closing

PR #6419, opened 2020-02-25
fix-scalar-mobile-minmax

JohnDanz:
fixed an issue where min-max wasn't shown for scalar markets on mobile unless you expanded the ticket
https://github.com/AugurProject/augur/issues/6281


bthaile (approved):
lgtm

Issue #6299, opened 2020-02-19
affiliate social sharing 


bthaile:
for all sharing links add user's account address (safe address) as a query parameter 
`affiliate` and make sure that address gets consumed by other user.

to verify share market and verify your address is the affiliate in URL.

cillianhunter0:
@bthaile what am i looking for here? (i don't see GS address in the URL)

cillianhunter0:
I'm not seeing my address in the URL or the copy 'affiliate' 

Copy market: 
**It's only copying the market ID - it should copy the full URL 
0x939840b6a28D159Ac8c393C8386e721391271f14

FB share link: 
https://v2.augur.net/?fbclid=IwAR3Yrq4zIlBC-_vPRDUAm9wDlmK__Sr_sae-8_UusPrSZ6rEb_Gd9I5H2EU#!/market?id=0x939840b6a28D159Ac8c393C8386e721391271f14

Twitter share link: 
https://v2.augur.net/#!/market?id=0x939840b6a28D159Ac8c393C8386e721391271f14

JohnDanz:
I'm seeing the affiliate on my end in v2:

Twitter share:
https://twitter.com/intent/tweet?text=Which%20city%20will%20have%20the%20highest%20median%20single-family%20home%20price%20on%20Wed%20Apr%2015%202020%20according%20to%20http://www.demographia.com?&url=https%3A%2F%2Fv2.augur.net%2F%23!%2Fmarket%3Fid%3D0x09277F8964794BA562Af53712a4B3486980e74CA&affiliate=0x995Bf32C10FA6f2b3680912300b985cc7c768DF0

Facebook:
https://www.facebook.com/sharer/sharer.php?t=Which%20city%20will%20have%20the%20highest%20median%20single-family%20home%20price%20on%20Wed%20Apr%2015%202020%20according%20to%20http://www.demographia.com?&u=https%3A%2F%2Fv2.augur.net%2F%23!%2Fmarket%3Fid%3D0x09277F8964794BA562Af53712a4B3486980e74CA&affiliate=0x995Bf32C10FA6f2b3680912300b985cc7c768DF0

as for the copy, it's copy market ID, if you mean on the market header I can change that in a PR

cillianhunter0:
I'm seeing the affiliate in the market URL, and can access the twitter and facebook share from the market card. also includes the affiliate. 

[...]

PR #6430, opened 2020-02-25
update-copy-market

JohnDanz:
updated copy market button on market page, updated titles for share buttons, copy button, and favorites button to have tooltips on mouseover.
https://github.com/AugurProject/augur/issues/6299


bthaile (approved):
lgtm

PR #6434, opened 2020-02-25
use provided node if on query param

### LINKED ISSUE #6429: User can specify ethereum_node_http in URL param and override window.web3 for fetching data.
if `ethereumNodeHttp` is specified in init-augur, don't set useWeb3Provider.

Testing concerns: When a user logs in, and we set their provider isn't this gonna override all the READ side of things and have it read from that provider not this node url? Is that a problem? I'm not even sure.

Or does this work just fine because we have an EthersProvider + a Signer? Signer goes through web3, Provider over this endpoint? I suppose that's the same as its always been.

andrewdamelio:
closes #6429 

if ethereum_node_http is specified in query param, use it as provider.

e.g
```http://localhost:8080?ethereum_node_http=https://eth-kovan.alchemyapi.io/jsonrpc/1FomA6seLdWDvpIRvL9J5NhwPHLIGbWA```

```http://localhost:8080/?ethereum_node_http=http://broken.com```

### CODE REVIEW

File: packages/augur-ui/src/config/network.json [hunk 1]
```diff
@@ -27,7 +28,7 @@
   "kovan": {
     "networkId": "42",
     "ethereum": {
-      "http": "https://eth-kovan.alchemyapi.io/jsonrpc/1FomA6seLdWDvpIRvL9J5NhwPHLIGbWA",
+      "http": "",
```
pgebheim: We should make this key optional on SDKConfiguration and then remove it from here.

packages/augur-ui/src/config/network.json
```diff
@@ -42,14 +43,15 @@
...
+        "verbosity": 1,
         "enabled": true
       }
     }
   },
   "demo": {
     "networkId": "42",
     "ethereum": {
-      "http": "https://eth-kovan.alchemyapi.io/jsonrpc/1FomA6seLdWDvpIRvL9J5NhwPHLIGbWA",
+      "http": "",
```
pgebheim: ditto

packages/augur-ui/src/config/network.json
```diff
@@ -64,6 +66,7 @@
         "ws":  "wss://demo.augur.net/0x-ws"
       },
       "mesh": {
+        "verbosity": 1,
```
pgebheim, line 67: woot woot.

packages/augur-ui/src/modules/app/actions/init-augur.ts
```diff
@@ -226,11 +226,23 @@ export function connectAugur(
...
     let networkId = env['networkId'];
 
-    // Unless DEV, use the provider on window if it exists, otherwise use torus provider
-    if (networkId && !isDevNetworkId(networkId)) {
+    if (env['ethereum'].http) {
+      // Use node provided in the ethereum_node_http param
+      try {
+        provider = new JsonRpcProvider(env['ethereum'].http);
+      } catch(error) {
+        dispatch(
```
pgebheim, line 234: nice touch!

pgebheim (changes requested):
Mind fixing up SDKConfiguration? Will probably need to check that `ethereum.http` exists in one place in augursdk or create-client as well.

pgebheim (approved):
Looks good

Issue #5968, opened 2020-02-06
Disputing: Mobile is missing checkbox for showing only markets in my portfolio 


matt-bullock:
From build:

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/580968d8-fbe4-4290-a654-70e5215eba45)

Design:

https://www.figma.com/file/aAzKHh4cA6OT2t7WFv2BQ7fB/Reporting-and-Disputing?node-id=375%3A83243

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/5d241fa8-ce5d-4586-bdc5-2b2437d1e838)

matt-bullock:
closing

PR #6435, opened 2020-02-25
show disputing checkbox on mobile

phoebemirman:
https://github.com/AugurProject/augur/issues/5968


bthaile (approved):
nice!, I knew that was going to be a big change.

Issue #5238, opened 2019-12-13
No alerts confirming REP migration  


cillianhunter0:
There's currently no confirmation alert when the REP is migrated to V2. Nothing happens...

Ran it by @Chwy5  show same spinner proposed for notifications when transaction is being processed. 

![image](https://user-images.githubusercontent.com/3970376/75048037-806d7380-548d-11ea-8c5d-44f0952aedd8.png)

show spinning icon when migration transaction is being processed.

matt-bullock:
Shows up in the bell dropdown but not as a toast. Is toast required?

cillianhunter0:
I'm not seeing this in the bell dropdown. 

We should also enable the toast for this, it's an important transaction, we should make the confirmation state as visible as possible.

The button state should also update using the new transaction feedback pattern (processing, confirmed etc)

matt-bullock:
toast and bell now there

PR #6436, opened 2020-02-25
Alert migrate v1 v2 rep

bthaile:
add alert when token minted event comes in from sdk

image

alert for 
https://github.com/AugurProject/augur/issues/5238


bthaile (commented):
Looks like processing button will work.

phoebemirman (approved):

PR #6440, opened 2020-02-25
Default `includeWarSyncMarkets` `getMarkets` param to default to false

justinbarry:
In the vast majority of cases the warp sync market is of no interest
to clients of the sdk.


bthaile (approved):
Lgtm

Issue #4376, opened 2019-10-23
Filter Settings: save user's preferences for market list filters


bthaile:
Filter Settings: save user's preferences for market list filters

add filterSettings to loginAccount and store to localstorage. so UI can rem the user's preferences when on market list page.

Chwy5:
What is the status on this?

Chwy5:
fixed closing

PR #6441, opened 2020-02-25
save users marketList filters

andrewdamelio:
Filter Settings: save user's preferences for market list filters #4376


bthaile (approved):
lgtm

Issue #6395, opened 2020-02-21
User should be logged in as fast as possible


pgebheim:
Currently, the code which calls the wallet SDK login functions waits until state.isConnected is set -- which isn't updated until SDK:Ready is received, signaling that all data has synced.

We should log in before all data is synchronized, and make sure that any tasks that are triggered by being logged in are (a) Deffered until SDK:Ready IFF that task needs access to the getters, (b) execute correctly outside of all data being sycned.

FYI:
manually tried to login during hotloading:

this error occurs because `marketDoc` is undefined in this statement. 
`const minPrice = new bignumber_js_1.BigNumber(marketDoc.prices[0]);`

Markets db hasn't been populated. Can't get user data until db is fully populated.

error that occurs
![image.png](https://images.zenhubusercontent.com/5c00901a5cb43a45a18aeb7f/2c0af764-5e49-4808-aa11-9117d369d6f9)

pgebheim:
Closed #6467

PR #6467, opened 2020-02-26
remove auto-login modals

andrewdamelio:
#6395 -> Tom did most of the work in regards to the ticket.

- Don't show login modals when auto log-in users
- If MetaMask is locked, show signing modal


bthaile (approved):
this works. tested with:
1. MM logged in then refresh (auto-login happens)
2. portis logged in then refresh (auto-login happens)
3. logged out and refreshed, user wasn't logged in. 

all looks good.

PR #6473, opened 2020-02-26
need to change 24 hr unrealized when market last trade changes not if…

bthaile:
… user trades in last 24 hr

noticed issue where user has to trade within last 24 hours in order for 24 hr unrealized to be calculated.


nuevoalex (approved):
LGTM

Issue #5898, opened 2020-02-04
Claim stake and fees notification and modal: if DAI is 0 show 0 rather than dash


matt-bullock:
Build:

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/57904d86-0446-4a13-84db-190abf42f294)

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/352b1ef4-65f7-4272-afbf-c5a5f85be4f7)

matt-bullock:
closing

PR #6476, opened 2020-02-26
change-null-dai-notifcations

JohnDanz:
changed claims from - to 0 when there is no dai
https://github.com/AugurProject/augur/issues/5898


bthaile (approved):
Lgtm

PR #6478, opened 2020-02-26
add pending data to buying participation tokens

phoebemirman:


packages/augur-ui/src/modules/reporting/actions/participation-tokens-management.ts
```diff
@@ -26,6 +29,11 @@ export const purchaseParticipationTokens = (
     );
     return callback(null, gas);
   }
-  buyParticipationTokens(universeId, amount);
+  dispatch(addPendingData(BUY_PARTICIPATION_TOKENS, BUY_PARTICIPATION_TOKENS, TXEventName.Pending, BUY_PARTICIPATION_TOKENS));
+  buyParticipationTokens(universeId, amount).then(() => {
+    dispatch(removePendingData(BUY_PARTICIPATION_TOKENS, BUY_PARTICIPATION_TOKENS))
```
bthaile: don't use `then` to remove pending queue use log-handler `handleTokensMintedLog`
use same pattern as migrate V1 REP to V2.

packages/augur-ui/src/modules/reporting/common.tsx
```diff
@@ -1222,11 +1221,12 @@ export const ParticipationTokensView = (
           "The total amount of unclaimed Dai you've earned through reporting"
         }
       />
-
-      <PrimaryButton
+      <ProcessingButton
         disabled={!hasRedeemable}
```
bthaile: disabling is off, the button should be disabled if there is a pending queue entry with status `pending`

bthaile (approved):
lgtm

PR #6483, opened 2020-02-27
add include warp sync flag for disputing and reporting page

bthaile:
show warp sync market in open reporting and on disputing page.

image


bconfortin (approved):
lgtm

Issue #6469, opened 2020-02-26
processing buttons


bthaile:
https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=8223%3A4544

![image](https://user-images.githubusercontent.com/3970376/75471869-608bf300-5958-11ea-903c-94f4698dccf4.png)

for all buttons that user clicks on to run a transaction change button to `processing ...` and disable. 
buttons to convert:

Market trading page:
report (designated reporter and open reporting) this should be done but need to use new processing button
claim proceeds. 
finalize (if used on market trading page)

Reporting page:
all buttons for DR and open reporting (use new processing button)

Disputing
- Disputing buttons on cards (disable all buttons while one is processing)
- Participation redeem and buy

If there is a notification that opens the modal like buttons above. Also change the button to `processing ...` in later effort the spinner will be used in specific cases. 

proposed to Design:
```
what if we do this:
happy path:
1. show `pending....` on button and disable when user kicks off a tx

success:
3. revert the text and activate the button when UI gets SDK event and all values are updated in UI

failure:
3a. if tx fails update button text to Failure and make button red with (X) to clear error so that user can tray again.

refresh:
if tx is still processing button needs to say `processing ...` and be disabled.
```
not going to handle timeout or long running transactions, just show `processing ...`
~~timeout:~~
~~3b. auto clear `processing ...` if tx hasn't confirmed after 20 blocks~~

dev work:
```
what if we do this:
happy path:
1. add pending queue entry 
2. when tx is pending update pending Queue with transaction hash and block number

success:
3. if tx is successful update pending queue entry to Success
4. clear pending queue entry when specific event comes into log-handler 

failure:
3a. if tx fails update pending queue entry status to Failure
4a. processing button changes to red with (X) button for user to clear entry from pending queue

refresh:
check hash, if confirmed remove from pending queue. if not confirmed then button should say `processing...`
```

not going to handle timeout or long running transactions, just show `processing ...`
~~timeout:~~
~~3b. auto clear `processing ...` if tx hasn't confirmed after 20 blocks~~

matt-bullock:
Need to get my local fixed before I can test most of these

matt-bullock:
looks done. closing

PR #6492, opened 2020-02-27
update processing btn states

phoebemirman:
https://github.com/AugurProject/augur/issues/6469


packages/augur-ui/webpack.config.js
```diff
@@ -15,7 +15,7 @@ const DeadCodePlugin = require("webpack-deadcode-plugin");
 // DEVELOPMENT CONFIG
 if (!process.env.DEBUG_BUILD && process.env.NODE_ENV === "development") {
   baseConfig = merge(baseConfig, {
-    // devtool: "eval-source-map",
+    devtool: "eval-source-map",
```
bthaile: I think this causes dev process to really slow down. prob don't want to check this in.

bthaile (approved):
lgtm

Issue #6486, opened 2020-02-27
REP migration update 


cillianhunter0:
https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=14936%3A224

IF user has V1 REP in GS, THEN show option 1
IF user has V1 REP in wallet, THEN show option 2
IF user has V1 REP in GS AND WALLET, THEN show option 1, when complete, THEN show option 2 

![image.png](https://images.zenhubusercontent.com/5c7591d7eec15745ffae99b1/2252fc21-7f91-4098-9926-1729c5df5beb)

cillianhunter0:
Spotted an issue that's confusing. 

When you have v1 rep in your GS, the tooltip says: 
"You have V1 REP in your wallet. Migrate it to V2 REP to use it in Augur V2."
You don't, it's in your 'augur account address' etc. 

We can reduce the confusion here by having two versions of this: 

1: IF user has v1 rep in their GS (or IF user has v1 rep in both GS and wallet) 
"You have V1 REP in your Augur account address. Migrate it to V2 REP to use it in Augur V2."

2: IF user has v1 rep in their wallet 
"You have V1 REP in your wallet. Migrate it to V2 REP to use it in Augur V2."

cillianhunter0:
@andrewdamelio is there any way i can test the modal with legacy v1 rep in my MM wallet? Not sure how i can send legacy rep from GS to my wallet?

cillianhunter0:
Confirmation state of button is broken: 
![image.png](https://images.zenhubusercontent.com/5c7591d7eec15745ffae99b1/d0dd1bcd-3e19-49cb-bcd3-c29bed8b3fd4)

Typo in the version of the modal when v1 rep is in wallet: 
first word of second paragraph "then" -> "When"

cillianhunter0:
Failed state of button is broken: 
![image.png](https://images.zenhubusercontent.com/5c7591d7eec15745ffae99b1/5f9795f8-6c60-4cc2-a7d8-f67bc7d7b496)

Confirm state of button is still broken in chrome. (you can remove the X from this button if it will fix the problem, as it's not needed in this case. this doesn't apply to the fail state) 
and
It disappears too quickly. The confirmed state should show for 3 seconds before being removed.
[...]

Issue #4898, opened 2019-11-25
[Dev] Support wallets that have REP for migration even if they're operating with a Gnosis Safe


pgebheim:

joeykrug:
v2 launch needed

PR #6494, opened 2020-02-27
Show custom migration modal for non-safae wallets with V1 REP

andrewdamelio:
## Show custom migration modal for non-safae wallets with V1 REP

To Test:
1) Disable Gnosis in network.json
2) Use the V1 faucet
3) Enable Gnosis in network.json
4) Login with Gnosis
5) You should see one of the below paths depending if you have V1 REP in either/or/both Safe address or non-safe address.

IF user has V1 REP in GS, THEN show option 1
IF user has V1 REP in wallet, THEN show option 2
IF user has V1 REP in GS AND WALLET, THEN show option 1, when complete, THEN show option 2

#### OPTION 2

Relates to..
- Support wallets that have REP for migration even if they're operating with a Gnosis Safe #4898
- REP migration update #6486

### DIFF
--- a/packages/augur-tools/src/flash/scripts.ts
+++ b/packages/augur-tools/src/flash/scripts.ts
@@ -253,19 +253,33 @@ export function addScripts(flash: FlashSession) {
         abbr: 't',
         description: 'Account to send funds (defaults to current user)',
         required: false
+      },
+      {
+        name: 'useLegacyRep',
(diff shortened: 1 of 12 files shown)


bthaile (approved):
lgtm, fixed a typo and added legacy REP option to rep-faucet flash script.

PR #6495, opened 2020-02-28
add pending dispute

phoebemirman:


bthaile (approved):
lets run with this as an iteration so we can keep testing with all pending stuff on master

PR #6503, opened 2020-02-28
bump Torus and use safe address when using onramp

andrewdamelio:
- Bump up torus version
- Fiat onramp should use the users wallet/safe address.

Screen Shot 2020-02-28 at 12 22 09 PM


bthaile (approved):
lgtm

PR #6510, opened 2020-02-28
add warp sync hash to universe details in getter, add emit event when…

bthaile:
… warp sync hash has been computed/updated


justinbarry (approved):
LGTM

phoebemirman (approved):

PR #6512, opened 2020-02-28
remove finalize notification, not needed any more

bthaile:
finalize market notification isn't needed market creator gets a claim REP and fees notification which would auto-finalize the market.


andrewdamelio (approved):
LGTM

PR #6518, opened 2020-03-01
Robert/sdk config refactor

robert-e-davidson3:


azure-pipelines.yml
```diff
@@ -443,10 +443,12 @@ jobs:
...
           docker images
-          docker tag $(IMAGE_NAME) augurproject/$(IMAGE_NAME):"${SOURCE_BRANCH}:${CONTRACT_SHA}"
-          docker push augurproject/$(IMAGE_NAME):"${SOURCE_BRANCH}:${CONTRACT_SHA}"
-        env:
-          SOURCE_BRANCH: $(variables['Build.sourceBranch'])
+          docker tag $(IMAGE_NAME) augurproject/"$(IMAGE_NAME):${SOURCE_BRANCH}:${CONTRACT_SHA}"
+          docker push augurproject/"$(IMAGE_NAME):${SOURCE_BRANCH}:${CONTRACT_SHA}"
+          if [ "$SOURCE_BRANCH" == "master" ]; then
+            docker tag $(IMAGE_NAME) augurproject/"$(IMAGE_NAME)"
+            docker push augurproject/"$(IMAGE_NAME)"
```
pgebheim: I still think this needs to go once this is merged in or else we'll end up with a big lag between contract changes and people being able to use them (1hr or whatever the post master merge time is)

pgebheim (approved):

azure-pipelines.yml
robert-e-davidson3: Oh oops. Actually that will never run because that task specifically excludes the `master` branch.

PR #6519, opened 2020-03-01
Update unrealized 24hr

bthaile:
create rollup 24hr percentage from individual positions to market positions to total portfolio.


packages/augur-sdk/src/state/getter/Users.ts
```diff
@@ -689,6 +693,26 @@ export class Users {
       }
     );
 
+    const endTime = await augur.contracts.augur.getTimestamp_();
```
nuevoalex, line 696: Instead of doing a call here you could just use Dexie.maxKey in the query for the endTime since the goal is to get everything in that direction

nuevoalex (approved):
One suggestions that will improve getter time, otherwise looks good!

packages/augur-sdk/src/state/getter/Users.ts
bthaile: endTime is used to calculate start time. Is there a dexie way to saying Dexie.maxKey minus X seconds?

nuevoalex: ah right, nevermind then. I thought there was somewhere we cached block time and only updated it on new blocks but not really a huge deal as its one call.

Issue #5986, opened 2020-02-06
Warp Sync UI changes


bthaile:
general direction for managing Warp Sync market in UI. market reporting state:

 - [x] for open reporting need to populate outcome input from sdk call to get hash
 - [x] for disputing need to populate outcome input from sdk call, just like open reporting.
 - [x] Allow users to put in hash value into outcome input. 
 - [x] Remove min and max price range in reporting and disputing form

current issue with disputing and reporting form. 
![image](https://user-images.githubusercontent.com/3970376/75615848-d5089280-5b0e-11ea-9c05-f825bb574fe3.png)

Open: filter out in all getters. 
In-Reporting: 
 - Open Reporting
market shows up in-reporting from getters. 
in reporting form, auto populated input with result of calling SDK method. 

- Disputing
if market's tentative winning outcome is same as users sdk hash filter out. (is this too resource intensive)
if user's SDK hash is different than tentative winning outcome market shows in disputing section.
in dispute form, auto populated input with result of calling SDK method. 

-   finalization
market shows in resolve section. ~~User can finalize the market.~~  user will get claims notification then they claim which will finalize the market.

Design
https://www.figma.com/file/aAzKHh4cA6OT2t7WFv2BQ7fB/Reporting-and-Disputing?node-id=3782%3A504

nuevoalex:
The UI should indicate that they will be compensated in REP for Open Reporting / Finalizing the market as well.

pgebheim:
- Disputing
  - This isn't too resource intensive and we *want* to do it all the time anyway because that is how we will have all clients pinning the warp sync state. However, I don't think we should filter it out. If someone is trolling this market we want people to be aware of it and to correct that, even if they are currently the same as the tentative winner.

@petervecchiarelli -- I don't recall if we have a help center section about Warp Sync, but I think we should write a good one and liberally link to it from the dispute cards for warp sync markets.

Also, I think we should have a special icon for warp markets so that its not displayed as a scalar because although it is a scalar it doesn't behave that way in the UI at all.

For the market's range / price I think we need to somehow just have it say it's resolving to an VALUE and not show the range at all because that's just confusing.

pgebheim:
This should be ready to work now with the pre-requisites in.

matt-bullock:
Happy path looks to be working ok. Going to close this

PR #6528, opened 2020-03-02
warp sync reporting form

phoebemirman:
https://github.com/AugurProject/augur/issues/5986


bthaile (approved):
lgtm, need to make sure user can report with hash.

PR #6552, opened 2020-03-03
filter out warp sync market if ten winning outcome is equal to curren…

bthaile:
…t warp sync hash, fix blank title  parsing error with warp sync market. checkpoint on allowing users to report on warp sync market when in DR

1. fix parsing title of warp sync.
2. filter out warp sync market from disputes page if ten win outcome is same as current warp synch hash
3. unit test for testing hash
4. allow user to report on warp sync market during designated reporting state.


justinbarry (approved):
LGTM

phoebemirman (approved):

Issue #6086, opened 2020-02-11
Trade history: price and quantity columns aren't lined up with labels


matt-bullock:
![Kapture 2020-02-11 at 17.11.11.gif]

Build:
![Kapture 2020-02-11 at 17.11.11.gif](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/3f4a06c6-700f-4cec-896b-349f15c7c81f)

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/332b26ac-a30f-4347-92cb-88fe6d669384)

matt-bullock:
done

PR #6555, opened 2020-03-03
align trading history

andrewdamelio:
Trade history: price and quantity columns aren't lined up with labels #6086


bthaile (approved):
lgtm

PR #6558, opened 2020-03-03
Updates for menus

JohnDanz:
various updates around menus/top menu and side

updated styles for betting/sports book on mobile
updated text of menu titles


bthaile (approved):
rubber stamp

Issue #5660, opened 2020-01-23
Blue tap highlight colour on outcome selector - showing on mobile


matt-bullock:
Is this needed as it's very distracting. If anything it should only appear over the button and not the whole menu

![Kapture 2020-01-23 at 15.31.51.gif](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/e6a957e9-9f3f-4094-be41-93a871d4e74e)

matt-bullock:
fixed. closing

PR #6581, opened 2020-03-04
remove blue highlight

andrewdamelio:
Blue tap highlight colour on outcome selector - showing on mobile #5660
BEFORE:

AFTER:


bthaile (approved):
Lgtm

Issue #6564, opened 2020-03-04
Mobile order book issue - offer side is auto scrolling to highest offer


matt-bullock:
In order to recreate, go to this market and you'll see the first time you come here the offer side loads as it should and scrolls as normal

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/9056d502-1454-413a-934b-0f04a90f7dda)

After scrolling, leave the page by clicking on market info tab or viewing the trades for another outcome and then returning to the first outcome. You'll find the offer side has either scrolled up to show the highest offer or will not scroll but then loads in showing the highest offer like so:

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/2ed100b4-1036-488b-9ab8-2ead3df107c0)

Chwy5:
This needs to show the Best offer (lowest offer available to be taken)

matt-bullock:
I'm testing outcome London in [this market](https://v2.augur.net/#!/market?id=0x09277F8964794BA562Af53712a4B3486980e74CA) on my mobile using chrome. It loads in showing the best offer but after a split second it is scrolling up to show the highest offer @JohnDanz

matt-bullock:
Unable to load larger order books on my phone currently. Need to wait for something to get fixed before I can test

matt-bullock:
fixed now. Closing

PR #6586, opened 2020-03-04
fix-orderbook-scrolling

JohnDanz:
fixed orderbook scrolling issue

https://github.com/AugurProject/augur/issues/6564

### DIFF
--- a/packages/augur-ui/src/modules/market-charts/components/order-book/order-book.tsx
+++ b/packages/augur-ui/src/modules/market-charts/components/order-book/order-book.tsx
@@ -31,7 +31,6 @@ interface OrderBookSideProps {
   setHovers: Function;
   type: string;
   marketType: string;
-  scrollToTop: boolean;
   hoveredSide?: string;
   hoveredOrderIndex?: number;
   showButtons: boolean;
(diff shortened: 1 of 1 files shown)


packages/augur-ui/src/modules/market-charts/components/order-book/order-book.tsx
```diff
@@ -70,23 +69,20 @@ class OrderBookSide extends Component<OrderBookSideProps, {}> {
   static defaultProps = {
     fixedPrecision: 4,
     pricePrecision: 4,
-    scrollToTop: false,
   };
 
   componentDidMount() {
-    if (this.props.scrollToTop) this.side.scrollTop = this.side.scrollHeight;
+    const { type } = this.props;
+    this.side.scrollTop = type === BIDS ? 0 : this.side.scrollHeight;
```
bthaile, line 76: what about ask section of orderbook? is that not an issue?

JohnDanz: this covers both now instead of passing a prop -- so if it's bids scroll to the top of the side (best bid) if it's asks scroll to the bottom (scroll height)

bthaile (approved):
lgtm

PR #6618, opened 2020-03-05
removing single-char commands from common flash options so they dont …

robert-e-davidson3:
…overwrite more specific options


bthaile (approved):
love it.

PR #6622, opened 2020-03-06
Update defaults

pgebheim:


joeykrug (approved):

robert-e-davidson3 (approved):
Might need to change the config the tests use too, in `packages/augur-tools/src/libs/TestEthersProvider.ts:29`.

pgebheim:
@robert-e-davidson3 Can you add that test commit?

robert-e-davidson3:
> Might need to change the config the tests use too, in `packages/augur-tools/src/libs/TestEthersProvider.ts:29`.

Turned out to be unnecessary.

Issue #6575, opened 2020-03-04
Betting/Sports UI - Update Footer


JohnDanz:
https://www.figma.com/file/42QoTLQcbObu7esoTPjWRv/Augur-Bet---Design-System?node-id=568%3A2

![Screen Shot 2020-03-04 at 11.35.40 AM.png](https://images.zenhubusercontent.com/57155267bf93871ccd99a48b/163b7f19-b800-45a1-8f64-fbbc67c38fe5)

Update the footer for spotsbook/betting ui -- we might also need to add it to pages where it's not currently shown in the trading UI.

PR #6625, opened 2020-03-06
Betting/Sports UI - Update Footer

rubyweb3:
https://github.com/AugurProject/augur/issues/6575

Screen Shot 2020-03-06 at 4 56 16 PM


rubyweb3:
@JohnDanz  Is there a way to get Augur UI: 6.1.4?

PR #6633, opened 2020-03-06
claim proceeds processing

phoebemirman:


packages/augur-ui/src/modules/events/actions/add-update-transaction.ts
```diff
@@ -44,8 +44,7 @@ const ADD_PENDING_QUEUE_METHOD_CALLS = [
   MIGRATE_FROM_LEG_REP_TOKEN,
   REDEEMSTAKE,
   BATCHCANCELORDERS,
-  TRADINGPROCEEDSCLAIMED,
-  CLAIMMARKETSPROCEEDS
```
bthaile, line 48: removing `CLAIMMARKETSPROCEEDS` here does Notification `view` button show processing if user clicks `claim all` ?

I think this can be left in.  It should not conflict with individual claims. the notification button will just show `processing` while the individual claim is being processed. it will disable the notification button though.

phoebemirman: it does. we only want it if multiple markets have been claimed so had to remove it there

bthaile: I think this can be left in. It should not conflict with individual claims. the notification button will just show processing while the individual claim is being processed. it will disable the notification button though.

phoebemirman: and then they can't get back to the modal to make more claims if they are in the process with one?

bthaile: could have the notification button not disable.

bthaile: just another param to pass into processingButton

phoebemirman: its debatable

bthaile (approved):
lgtm, we'll run with it.

PR #6635, opened 2020-03-06
Fixed bootstrapList, zeroX.rpc.enabled, and typing

robert-e-davidson3:


packages/augur-tools/src/flash/scripts.ts
```diff
@@ -466,8 +472,18 @@ export function addScripts(flash: FlashSession) {
 
   flash.addScript({
     name: 'create-canned-markets',
-    async call(this: FlashSession) {
-      const user = await this.ensureUser();
+    options: [
+      {
+        name: 'use-gnosis',
```
pgebheim, line 477: @bthaile -- what are we calling this with the new GSN stuff.... would be like... use-proxy-wallet?

pgebheim:
Only issue I see on this is docker:all --dev errors....

pgebheim (approved):

PR #6637, opened 2020-03-06
Refactor for multiple environments/states

williamdanger:
Moved everything into `modules` and `envs`. To deploy a cluster now, `cd infra/envs/v2/ && make apply`


azure-pipelines.yml
```diff
@@ -1,20 +1,20 @@
 variables:
```
williamdanger, line 1: Passed this through a YAML linter. Meat and potatoes below

azure-pipelines.yml
```diff
@@ -653,6 +653,58 @@ jobs:
           AWS_ACCESS_KEY_ID: $(AWS_KEY_ID)
           AWS_SECRET_ACCESS_KEY: $(AWS_SECRET_KEY)
 
+
```
williamdanger, line 656: Review starts here. This pushes to the v2 s3/Cloudfront bucket.

pgebheim (approved):

PR #6645, opened 2020-03-07
add soonest order's expiration time to top level of orderbook.

bthaile:
set soonest to expire order's expirationTimeInSeconds on orderbook object returned from getter.


robert-e-davidson3 (approved):
nice!

Issue #6601, opened 2020-03-05
Sports Betting - Update Modal Styles


JohnDanz:
https://www.figma.com/file/42QoTLQcbObu7esoTPjWRv/Augur-Bet---Design-System?node-id=17%3A12

Update/check that all in use modals are styled correctly. This could be a shorter task than I've estimated but be thorough. Also in the process, identify modals that are no longer in use or are extremely rarely used/are very similar to another type of modal. This way we can clean up some older components.

rubyweb3:
Close?

PR #6649, opened 2020-03-09
Sports Betting - Update modal mesasge styles

rubyweb3:
https://github.com/AugurProject/augur/issues/6601

Some updated modals are:
Screen Shot 2020-03-09 at 5 50 54 PM
Screen Shot 2020-03-09 at 5 50 45 PM


packages/augur-ui/src/modules/modal/common.tsx
```diff
@@ -426,18 +417,6 @@ export const CallToAction = (props: CallToActionProps) => (
   <h3>{props.callToAction}</h3>
 );
 
-export const SelectableTable = (props: SelectableTableProps) => (
```
rubyweb3, line 429: SelectableTable is not used anymore.

JohnDanz (approved):

PR #6652, opened 2020-03-09
fix migration button layout break

andrewdamelio:
https://github.com/AugurProject/augur/issues/6486#issuecomment-595156783

### DIFF
--- a/packages/augur-ui/src/modules/app/components/top-nav/top-nav.styles.less
+++ b/packages/augur-ui/src/modules/app/components/top-nav/top-nav.styles.less
@@ -101,15 +101,6 @@
       }
     }
 
-    > .MigrateRep {
-      button {
-        .mono-10-bold;
-        text-transform: uppercase;
-      }
-      position: absolute;
-      right: calc(96px + @size-48);
-    }
-
     > .BettingUI {
       position: absolute;
       right: @size-32;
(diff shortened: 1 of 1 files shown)


bthaile (approved):
not sure about `@media all and (min-width: @tablet-min) and (max-width: 800px) {` usage. don't we have a breakpoint for all layouts.

bthaile (changes requested):
`@media all and (min-width: @tablet-min) and (max-width: 800px) {` don't we breakpoints to cover this?

packages/augur-ui/src/modules/app/components/top-nav/top-nav.styles.less
```diff
@@ -170,6 +161,14 @@ button.AddFunds {
   }
 }
 
+.MigrateRep {
+  > button {
+    @media all and (min-width: @tablet-min) and (max-width: 800px) {
```
JohnDanz, line 166: why this ad-hoc media breakpoint? we shouldn't be deviating from the mocks and if the mocks have odd font-sizes between breakpoints the design should be adjusted instead. Also don't set `font-size` directly, use one of the helpers like `.text-9` or whatever (8 seems too small, I don't think we should be using anything that low as that's outside of the typography we have laid out, another thing to bring up to design). 

This is most likely an oversight with the design rather than a hard request.

edit:

looking at the issue referenced in the OP of this PR I would suggest we adjust the spacing on that topnav instead of reducing button text. see my comment on the issue.

JohnDanz (changes requested):

Issue #5892, opened 2020-02-04
Expired timestamp should show the date the market expired


matt-bullock:
Where this needs updating: My Created Markets in Portfolio

Build:

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/ca299f21-4870-4903-84b5-2959761abb92)

Design:

https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=2640%3A29211

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/a2a1762e-66b4-4f8b-9b72-b7311a7b5035)

Chwy5:
fixed - closing

PR #6660, opened 2020-03-09
Market awaiting finalization doesnt have countdown anymore

bconfortin:
#5892 

Markets awaiting finalization were showing up with a countdown timer on the portoflio page (my created markets section).

### DIFF
--- a/packages/augur-ui/src/modules/common/progress.tsx
+++ b/packages/augur-ui/src/modules/common/progress.tsx
@@ -209,13 +209,13 @@ export const CountdownProgress = (props: CountdownProgressProps) => {
 
     timeLeft = time.timestamp - currentTime.timestamp;
     countdown = (countdownBreakpoint || OneWeek) >= timeLeft && timeLeft > 0;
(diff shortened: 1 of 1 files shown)


packages/augur-ui/src/modules/common/progress.tsx
```diff
@@ -209,13 +209,13 @@ export const CountdownProgress = (props: CountdownProgressProps) => {
 
     timeLeft = time.timestamp - currentTime.timestamp;
     countdown = (countdownBreakpoint || OneWeek) >= timeLeft && timeLeft > 0;
-    valueString = countdown
-      ? `${daysRemaining}:${
+    valueString = !countdown || label === 'Expired'
```
bthaile: Dont use Expired, 
If the market is awaiting finalization or finalized then there is no countdown timer.

bthaile (changes requested):
Need to not use countdown timer if market is in awaiting finalization or finalized.

bthaile (approved):
Lgtm

PR #6664, opened 2020-03-09
add individual pending items for claiming stake

phoebemirman:


bthaile (approved):
That was easy. Does it work when there are multiple transacations needed to claim a lot of contracts?

Issue #4664, opened 2019-11-07
Breaking Changes to the MetaMask Inpage Provider


andrewdamelio:
https://medium.com/metamask/breaking-changes-to-the-metamask-inpage-provider-b4dde069dd0a

pgebheim:
Done for V1

joeykrug:
is this done for v2?

andrewdamelio:
To be done still - shouldn't be a big effort though

joeykrug:
let's get it done

PR #6670, opened 2020-03-09
Update MetaMask to use eth_requestAccounts

andrewdamelio:
- Update MetaMask to use eth_requestAccounts [ #4664]
- prevent wallet sdks from overwriting  injected web3s


bthaile (approved):
lgtm

PR #6673, opened 2020-03-10
Bruno/add plugin configuration

bconfortin:
Main purpose of this commit:
- Added environment configuration to disable orbit and facebook comments

Unrelated very annoying bugs that popped up in console:
- Fixed a bug with TopNav map function not having "key" prop
- Fixed a bug with MarketCards React Child being passed a function instead of a component


packages/augur-artifacts/src/index.ts
```diff
@@ -70,6 +70,10 @@ export interface SDKConfiguration {
   },
   uploadBlockNumber?: number,
   addresses?: ContractAddresses,
+  plugins?: {
+    chat?: '3box'|'orbit'|undefined,
```
robert-e-davidson3: Undefined is redundant here because you used a question mark.

packages/augur-ui/src/modules/global-chat/containers/global-chat.ts
```diff
@@ -1,24 +1,14 @@
...
-import GlobalChat from 'modules/common/global-chat';
-import { closeModal } from 'modules/modal/actions/close-modal';
-import { AppState } from 'store';
-import { ThunkDispatch } from 'redux-thunk';
-import { Action } from 'redux';
+import { GlobalChat } from 'modules/global-chat/components/global-chat.tsx';
 
-const mapStateToProps = (state: AppState) => ({
+const mapStateToProps = (state) => ({
+  whichChatPlugin: state.env.plugins && state.env.plugins.chat,
```
robert-e-davidson3: Can use the `?.` operator here.

robert-e-davidson3:
The 3rd party test failure is due to a bug in `docker:all --detach`

bconfortin:
> The 3rd party test failure is due to a bug in `docker:all --detach`

How do I make the test pass?

robert-e-davidson3:
> > The 3rd party test failure is due to a bug in `docker:all --detach`
> 
> How do I make the test pass?

In scripts.ts, the docker-all script has this line commented out. Uncomment it out:
```
await spawnSync('yarn', ['build']); // so UI, CI, etc will have the correct addresses
```

robert-e-davidson3 (approved):

PR #6681, opened 2020-03-10
orderbook-update

JohnDanz:
cleaned up and converted orderbook to a pure component, this should allow us to use a new useEffect we plan to move here from market view in another pr


bthaile (approved):
Lgtm

PR #6682, opened 2020-03-10
disable zeroX mesh for googlebot

andrewdamelio:
In the UI don't use Mesh if the userAgent is googlebot -- to hopefully unblock SEO efforts


packages/augur-ui/src/services/browser-mesh.ts
```diff
@@ -95,17 +96,19 @@ export async function createBrowserMesh(
...
 
-  const meshConfig = createBrowserMeshConfig(
-    config.ethereum.http,
-    web3Provider,
-    Number(config.networkId),
-    config.zeroX.mesh.verbosity || 5,
-    config.zeroX.mesh.useBootstrapList,
-    config.zeroX.mesh.bootstrapList
-  );
+  if (!isGoogleBot()) {
```
pgebheim: I think you should just put this check up one level where we actually create/read the `config` and just force `config.mesh.enabled = false`, then ALL of the places that check for mesh will be covered.

So, e.g. somewhere around here:
https://github.com/AugurProject/augur/blob/master/packages/augur-ui/src/modules/app/actions/init-augur.ts#L219

pgebheim (changes requested):

packages/augur-ui/src/services/browser-mesh.ts
andrewdamelio: Makes way more sense 👍

pgebheim (approved):

PR #6685, opened 2020-03-10
Changing colors

bconfortin:
Was trying to simulate an error that Matt got but was unsuccessful.
Noticed the modal's background color was wrong so I fixed it.


bthaile (approved):
lgtm

Issue #6687, opened 2020-03-10
when on disputing page and dispute time runs up, market card does not update and sticks around


phoebemirman:
<img width="1108" alt="Screen Shot 2020-03-10 at 12 50 31 PM" src="https://user-images.githubusercontent.com/6775839/76353344-c91a8e80-62cd-11ea-99d1-39300676f34e.png">

matt-bullock:
@phoebemirman I have to manually push time to test this as using fake time docker doesn't move time until pushed. It disappears from the disputing page when I do this but I'm not sure what would happen if the time was automatically ticking over. Presumably the card would disappear from the disputing page?

I'm closing unless you think this needs any further testing

Issue #6686, opened 2020-03-10
when time is up, markets don't move from upcoming to designated reporting or from designated to open reporting while on the page


phoebemirman:
![Screen Shot 2020-03-10 at 12.46.22 PM.png](https://images.zenhubusercontent.com/574e216bbf93871ccd9be5fb/b08d7d22-d719-42eb-aaff-de4222a959c5)

steps:
1. create market and make self DR
2. push time so market is in DR phase and be on reporting pg
3.see it does not move

PR #6692, opened 2020-03-10
add refresh when reporting state changes

phoebemirman:
https://github.com/AugurProject/augur/issues/6686
https://github.com/AugurProject/augur/issues/6687


bthaile (approved):
lgtm

Issue #6691, opened 2020-03-10
Market Page should error if it cannot load the market ID


pgebheim:
We should check the value of the market id in the market-view

https://github.com/AugurProject/augur/blob/master/packages/augur-ui/src/modules/market/containers/market-view.ts#L46

Make sure to checksum format it (use getAddress from ethers.js/utils) and error if its not a valid address or our utils `formatAddress`

Also, should error if the market is not found (e.g. if the hotloading can't load any market data).

Currently UI blows up in SDK

![image](https://user-images.githubusercontent.com/3970376/81417643-1cb02980-9111-11ea-93dd-eac1456e05bb.png)

There are two situations
1. marketId is not a valid ethereum address
2. marketId doesn't exists as a market

not a marketid but valid ethereum address
![image](https://user-images.githubusercontent.com/3970376/81576880-69dd0700-936e-11ea-9033-3114be8e7430.png)

matt-bullock:
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/3c79d133-bedb-471f-841b-21838ef26d4b)

I saw the error once but unable to bring it up again

bthaile:
@matt-bullock  I think the above only happens if user goes to a market trading page then changes the marketId to make an invalid market address. 

if the user clicks on a link with bad marketId address the market loads correctly with `market not found` modal. 
For example these links should be bad, first one is a bad ethereum address and the seconds is a valid address but not a valid marketId

<https://v2.augur.net/#!/market?id=0x057BDc6949450a897769F8c1597c84cf28Ff6CDa>
<https://v2.augur.net/#!/market?id=0xEA674fdDe714fd979de3EdF0F56AA9716B898ec8>

matt-bullock:
@bthaile ok these links show the message correctly so I'll close

PR #6694, opened 2020-03-10
Handle case where market ID isn't checksumed case

pgebheim:
refs #6691


bthaile (approved):
sure why not. just in case someone fat finger the marketId in.

PR #6714, opened 2020-03-11
Performance Testing Setup Script

### LINKED ISSUE #6813: Create performance testing kit
- [x] 1000 orders per second (in one market and 100/s across 10 markets)
 - [x] 100s of markets on the site
 - [x] test with user accounts that've traded in 100s of markets
 - [x] and with accounts that've done 1000s of trades and have 100+ open positions

robert-e-davidson3:
Fixes #6813


packages/augur-core/source/libraries/HelperFunctions.ts
```diff
@@ -6,9 +6,9 @@ export async function sleep(milliseconds: number): Promise<void> {
     return new Promise<void>(resolve => setTimeout(resolve, milliseconds));
 }
 
-export async function resolveAll(promises: Iterable<Promise<any>>) {
+export async function resolveAll<T>(promises: Iterable<Promise<T>>) {
     let firstError: Error|null = null;
-    for (let promise of promises) {
+    for (const promise of promises) {
```
pgebheim, line 11: Why is this serial?

packages/augur-tools/src/flash/scripts.ts
```diff
@@ -2002,6 +1995,18 @@ export function addScripts(flash: FlashSession) {
         description: 'Do not stop dockers after running command and do not wait for user input before exiting',
         flag: true,
       },
+      {
+        name: 'do-not-run-geth',
+        abbr: 'G',
```
pgebheim, line 2008: Love this

robert-e-davidson3:
> Ooo now now all of these are settable during webpack build?

Yup. The big one was GSN_ENABLED since our perf testing doesn't use GSN, to avoid it as as bottleneck. But now you can set whichever without needing to modify the environment file.

pgebheim (approved):

packages/augur-core/source/libraries/HelperFunctions.ts
robert-e-davidson3: It isn't serial, actually. The promises are already executing. Awaiting them just makes sure each one finishes.

Issue #6713, opened 2020-03-11
Missing order book


Chwy5:
created a market added liquidity...logged out. Refreshed. Swithced accounts and went to that market...clicked on multiple orders in the outcome table and the order book is empty for all of them....the outcomes I click on do show in the order ticket

![3-11ob.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/8245c016-3009-4e1b-9e9b-af4d20c84941)

Chwy5:
fixed - closing

PR #6715, opened 2020-03-11
fix orders in order book showing up

phoebemirman:
https://github.com/AugurProject/augur/issues/6713


bthaile (commented):
I think the change should be `const outcomeOrderBook = ownProps.orderBook || {}`

bthaile (approved):
lgtm

PR #6716, opened 2020-03-11
Added Paul's function to window

bconfortin:
Captura de Tela 2020-03-11 às 17 24 16


packages/augur-ui/src/modules/app/actions/init-augur.ts
```diff
@@ -263,6 +263,77 @@ export function connectAugur(
   };
 }
 
+function showIndexedDbSize() {
```
bthaile: put this in a different file. init-augur is busy, imo
create a file and put in util dir.

bthaile (dismissed):
simple organization change request.

pgebheim (approved):

pgebheim (approved):

PR #6723, opened 2020-03-12
Gh 6310 warp sync checkpoints must be generated based on market end time

### LINKED ISSUE #6310: Warp Sync: Checkpoints must be generated based on Market End Time
Need to generate a warp sync hash once we're 30 blocks past the last block before the market end time.

justinbarry:
closes #6310


packages/augur-test/src/tests/state/getter/markets/misc.test.ts
```diff
@@ -120,16 +120,29 @@ describe('State API :: Markets :: ', () => {
     beforeEach(async () => {
       await john.initializeUniverseForWarpSync();
 
-      const someHash = 'QmNLei78zWmzUdbeRB3CiUfAizWUrbeeZh5K1rhAQKChD2';
+      // This hash won't do anything because it is the IPFS for
```
nuevoalex: comment seems unfinished?

nuevoalex (approved):
👍

packages/augur-test/src/tests/state/getter/markets/misc.test.ts
justinbarry: Ha. Indeed it is.

PR #6726, opened 2020-03-12
Clean up configs and read from new config files

pgebheim:
First pass at fixing up the terraform configs to point to the new environment configs.


williamdanger (approved):
Awesome work! Everything looks good/sane to me.

Issue #6571, opened 2020-03-04
open orders: failed tag shouldn't hide close button. Please add back in


matt-bullock:
Build screenshots:

market page:

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/affbf84e-2420-47dd-aa40-55fa1d70d1a6)

portfolio:

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/c615a009-1bc8-47f1-aab6-9ef4dc0f38ec)

matt-bullock:
looks good. closing

PR #6730, opened 2020-03-12
use existing buttons for showing pending label and use failed dismis…

bthaile:
…sible button for failed open orders or failed trades

https://github.com/AugurProject/augur/issues/6571

Screen Shot 2020-03-12 at 10 56 14 AM

Screen Shot 2020-03-12 at 10 56 07 AM

Screen Shot 2020-03-12 at 11 09 01 AM


phoebemirman (changes requested):
should we make processing btn take the option to pass in status?

phoebemirman (approved):

PR #6731, opened 2020-03-12
fix-pagination

JohnDanz:
updated an issue with pagination where if you switched to all on the markets list you couldnt go back to 10 per page because the dropdown would not render anymore


bthaile (approved):
lgtm

PR #6747, opened 2020-03-13
use users local time to show smooth timers

bthaile:
smooth timers

### DIFF
--- a/packages/augur-ui/src/modules/common/progress.tsx
+++ b/packages/augur-ui/src/modules/common/progress.tsx
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import React, { useState, useEffect } from 'react';
 import Styles from 'modules/common/progress.styles.less';
 import * as format from 'utils/format-date';
 import classNames from 'classnames';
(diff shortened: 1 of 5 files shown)


packages/augur-ui/src/modules/common/progress.tsx
```diff
@@ -167,22 +163,34 @@ export const MarketProgress = (props: MarketProgressProps) => {
...
+
+  useEffect(() => {
+    const interval = setInterval(() => {
+        setCurrentTime(seconds => seconds + 1);
+      }, 1000);
+    return () => clearInterval(interval);
+  }, [currentTime]);
+
+  return currentTime;
+}
```
JohnDanz, line 182: Fire.

JohnDanz (approved):
only suggestion would be pulling the custom hook into a new `hooks` folder somewhere so we can start developing a standard practice for things like this and it can be re-used more easily instead of knowing it's in `progress.tsx` (it should just be in something like `common/hooks/use-timer.ts` or maybe in a more general file called like `time.ts` for time related hooks). can work that out later as we see what hooks we end up making/needing but the general idea of keeping them out of just component style files would probably be for the best since they are logic that is re-used and don't need to be tied to a specific component

PR #6755, opened 2020-03-13
Mobile tweaks to filters and sorting

bconfortin:
On mobile, when you click the filters/sorting, you now have to click on the "Apply" button in order to actually apply those.


bthaile (commented):
looks good to me. I'd like @andrewdamelio  to take a peek at it.

andrewdamelio (changes requested):
I noticed an issue when canceling a filter selection.

To Reproduce:
1) On Desktop view set all filters to All, url params will be ```spread=100&maxFee=1&templateFilter=all```
2) Now switch to mobile view and change the filters but don't click apply, click "X"
3) The page will crash due to the params being wrong, looks like they are getting jumbled up ```spread=hide&maxFee=100&templateFilter=1&showInvalid=all```


which is causing the below error

image

bconfortin:
@andrewdamelio Thanks for catching that. Lack of attention on my part while copying/pasting variables.

andrewdamelio (approved):
LGTM now!

PR #6758, opened 2020-03-13
Add config flag for WarpSync

justinbarry:


packages/augur-artifacts/src/configuration.ts
```diff
@@ -28,6 +28,7 @@ export interface SDKConfiguration {
     writeArtifacts?: boolean,
     externalAddresses?: ExternalAddresses,
   },
+  useWarpSync: boolean,
```
robert-e-davidson3: Should be optional.

```suggestion
  useWarpSync?: boolean,
```

packages/augur-sdk/src/state/create-api.ts
```diff
@@ -57,11 +74,9 @@ export function buildSyncStrategies(client:Augur, db:Promise<DB>, provider: Ethe
...
     blockAndLogStreamerSyncStrategy.listenForBlockRemoved(logFilterAggregator.onBlockRemoved);
 
-    // Check on each new block to see if we need to generate a checkpoint.
-    client.events.on(SubscriptionEventName.NewBlock, async (newBlock) => {
-      const block = await provider.getBlock(newBlock.lastSyncedBlockNumber);
-      await warpController.onNewBlock(block);
-    });
+
+    if(config.useWarpSync) {
+    }
```
robert-e-davidson3: Missing code here?

packages/augur-tools/src/flash/flash.ts
```diff
@@ -60,7 +60,11 @@ export class FlashSession {
     public accounts: Account[],
     public network?: string,
     public config?: SDKConfiguration,
-  ) {}
+  ) {
+    if(config) {
+      config.useWarpSync = false;
+    }
+  }
```
robert-e-davidson3: Better to do this in `cli.ts` because we mess with config there already.

robert-e-davidson3 (changes requested):

packages/augur-tools/src/flash/flash.ts
bthaile: what about adding something to `packages/augur-artifacts/src/configuration.ts` to turn it on so it'll be off by default

robert-e-davidson3: It can be set to `false` in the default config then anywhere that wants to use it can set it to true with mergeConfig.

packages/augur-sdk/src/state/create-api.ts
justinbarry: Good call, This was moved to the event handler above. Removed.

packages/augur-artifacts/src/environments/test.json
```diff
@@ -15,20 +15,18 @@
...
     "enableFaucets": true,
     "normalTime": false,
-    "contractInputPath": "/home/achapman/augur/packages/augur-artifacts/build/contracts.json",
-    "writeArtifacts": true,
-    "externalAddresses": {}
+    "writeArtifacts": false
   },
   "gsn": {
-    "enabled": true
+    "enabled": false
```
bthaile, line 21: doesn't GSN need to be enabled?

packages/augur-artifacts/src/environments/test.json
```diff
@@ -15,20 +15,18 @@
...
   },
   "zeroX": {
     "rpc": {
       "enabled": false,
       "ws": "ws://localhost:60557"
     },
     "mesh": {
-      "enabled": true,
+      "enabled": false,
```
bthaile, line 29: is there a reason to have this disabled?

bthaile (commented):
a few comments/questions

bthaile (approved):
lgtm

robert-e-davidson3 (approved):

PR #6759, opened 2020-03-13
change disputing flash scripts so sdk isn't needed

bthaile:
pass in all needed parameters to dispute/report into flash scripts so SDK isn't needed


packages/augur-tools/src/flash/scripts.ts
```diff
@@ -1834,52 +1804,31 @@ export function addScripts(flash: FlashSession) {
...
+        name: 'payoutNumerators',
+        abbr: 'p',
+        description: 'payout numerators of child unverse.',
+        required: true,
       },
     ],
     async call(this: FlashSession, args: FlashArguments) {
       if (this.noProvider()) return;
-      const user = await this.ensureUser();
+      this.config.gnosis.enabled = false;
```
robert-e-davidson3: It probably won't happen but this will fail if `config.gnosis` is undefined. Use `mergeConfig`:

```suggestion
      this.config = mergeConfig(this.config, { gnosis: { enabled: false }});
```

robert-e-davidson3 (approved):

PR #6770, opened 2020-03-15
fix issues where processing cancel order resulted in error and pendin…

bthaile:
…g queue status wasn't updated. fix issue where failed cancel order didn't show failed in UI. added ability to dismiss failure.

fixes this error

image

fixes issue where failed dismissible button doesn't show

image

show multiple failures for batch cancel orders
image


nuevoalex (approved):
👍

PR #6773, opened 2020-03-15
need to just remvoe pending cancel orders if they are successful

bthaile:
need to just remove pending cancel orders if they are successful


bthaile:
closing this, the code is correct. issue is SDK isn't emitting order event, so UI doesn't know to update user's open orders.

Issue #6709, opened 2020-03-11
gsn: Initializing Account via creating open order flow


matt-bullock:
If a user goes to enter an open order we trigger a warning message in the order ticket. They will need to initialise their account before they can place the open order

(All copy to be finalised)

https://www.figma.com/file/fLWVwmanAwetVZbujQquEi/Market-Page?node-id=3862%3A188115

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/2f236730-4798-416b-92cd-9cd9ef3b5067)

Clicking the initialise account button brings up this modal:

https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=15124%3A280324

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/456536b1-38b7-4be9-8447-deb38072ac87)

This transaction will require a toast alert and bell message

matt-bullock:
This is working. closing and creating a separate ticket for adding in the processing button and confirmed buttons

Issue #6708, opened 2020-03-11
Add Copy to learn more for gsn: add screen to onboarding that introduces the idea of charging a one off initial fee to initialize the users account and join the network


matt-bullock:
Add this copy to "learn more":

Augur runs on a peer-to-peer network which requires micro-fees to be paid in a digital currency called Eth for actions such as making trades.

In order to facilitate the ability for you to both trade and pay for transactions using only one currency, DAI ($), a small one-time fee will be automatically paid on your behalf when joining Augur. This fee is not collected by Augur.

--------------------------------------------------------------------------------------------------------------

This should come after the Add Dai to your account screen

Desktop:

https://www.figma.com/file/RMXc7jUWiC6amxz9Q2JK5L/Onboarding-and-Wallet-UX?node-id=2113%3A26

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/1b501751-0bc1-4f26-b002-5d4bb4c3890c)

Mobile:

https://www.figma.com/file/RMXc7jUWiC6amxz9Q2JK5L/Onboarding-and-Wallet-UX?node-id=2117%3A302

matt-bullock:
looks good. We need to add a link to the learn more. @petervecchiarelli do you have place in the help section this can link to? Feel free to suggest any edits to the copy in this onboarding screen

petervecchiarelli:
can someone give me figma access or show me what before/after steps are for more context?

matt-bullock:
@petervecchiarelli you can see the before and after steps in the signup flow

petervecchiarelli:
Can link here: https://augur.gitbook.io/help-center/trading-ui-1/signing-up#gas-station-network

How about something like this:

Augur runs on a peer-to-peer network which requires micro-fees to be paid in a digital currency called Eth for actions such as making trades. 

In order to facilitate the ability for you to both trade and pay for transactions using only one currency, DAI ($), a small one-time fee will be automatically paid on your behalf when joining Augur. This fee is not collected by Augur.

matt-bullock:
Just waiting for copy to be added to the help section. @petervecchiarelli please close when done

Add this copy to "learn more":

Augur runs on a peer-to-peer network which requires micro-fees to be paid in a digital currency called Eth for actions such as making trades.

In order to facilitate the ability for you to both trade and pay for transactions using only one currency, DAI ($), a small one-time fee will be automatically paid on your behalf when joining Augur. This fee is not collected by Augur.

[...]

PR #6787, opened 2020-03-16
add onboarding modal/initialize wallet button

andrewdamelio:
gsn: Initializing Account via creating open order flow #6709
gsn: add screen to onboarding that introduces the idea of charging a one off initial fee to initialize the users account and join the network #6708


bthaile (approved):
lgtm

PR #6790, opened 2020-03-16
Gsn list

nuevoalex:
- increase nonce gap
 - send "reason"with tx failure updates (to be used to help user know to retry with UI message)
 - remove manual setting of gas price in favor of dynamic movement
 - expose actual payment calculation and use in UI for tx fee estimate


bthaile (approved):
lgtm

Issue #6610, opened 2020-03-05
Sports Betting - My Account Overview


JohnDanz:
https://www.figma.com/file/42QoTLQcbObu7esoTPjWRv/Augur-Bet---Design-System?node-id=1227%3A0

![Screen Shot 2020-03-05 at 12.30.13 PM.png](https://images.zenhubusercontent.com/57155267bf93871ccd99a48b/609f4dd0-6685-444d-9c08-8ece471af662)

Update account overview styling for desktop and mobile to match design

PR #6793, opened 2020-03-17
Update Sports Betting - My Account Overview

rubyweb3:
https://github.com/AugurProject/augur/issues/6610
Update:
- Background
- Border
- Header

Screen Shot 2020-03-17 at 6 07 55 PM

### DIFF
--- a/packages/augur-ui/src/modules/account/components/account-view.styles.less
+++ b/packages/augur-ui/src/modules/account/components/account-view.styles.less
@@ -142,18 +142,49 @@
   .AccountView {
     > div {
       border: @size-1 solid var(--color-outline);
+      border-radius: @size-6;
+
       div[class*='BoxHeader'] {
         background: var(--color-table-header-50);
+        border-radius: @size-6;
 
         > div:first-of-type {
           background: transparent;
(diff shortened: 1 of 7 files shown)


packages/augur-ui/src/modules/account/components/account-view.styles.less
```diff
@@ -136,3 +136,56 @@
     }
   }
 }
+
+:root[theme="BETTING"],
+:root[theme="SPORTS"] {
+  .AccountView {
+    > div {
+      border: @size-1 solid var(--color-outline);
+      border-radius: @size-6;
```
JohnDanz: I would use `@border-radius-rounded` here (and all other @size-6 border-radius definitions) instead of 6 incase that ever changes.

JohnDanz (changes requested):
minor change request

packages/augur-ui/src/modules/account/components/account-view.styles.less
rubyweb3: Updated. Thank you.

JohnDanz (approved):

PR #6794, opened 2020-03-17
fix undefined .toNumber()

andrewdamelio:


packages/augur-ui/src/modules/auth/actions/update-assets.ts
```diff
@@ -63,7 +63,7 @@ function updateBalances(
     const eth = amounts[2];
     const legacyAttoRep = amounts[3].toString();
     const legacyAttoRepNonSafe = amounts[4].toString();
-    const rep = formatAttoRep(attoRep).roundedValue.toNumber();
+    const rep = formatAttoRep(attoRep).roundedValue && formatAttoRep(attoRep).roundedValue.toNumber();
```
pgebheim:
```suggestion
    const rep = formatAttoRep(attoRep).roundedValue?.toNumber();
```

bthaile (approved):
lgtm. weird that it would be null

Issue #6795, opened 2020-03-17
Getting wrong message when clicking on my own order in the outcome Table


Chwy5:
On the market trading pg....
Add open orders....on the outcome table, click on your own open order, order tackit filters in and shows Multiple Txs needed (this is the wrong message) should be getting "Consume own order" message

![3-17a.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/55b00af3-93ad-40dc-a27e-c91cd47e0cda)

Does work correctly if clicking the same order on the order book....issue is just through the outcome table

Chwy5:
fixed - closing

PR #6798, opened 2020-03-17
Add self trade prop outcome list

bthaile:
show consuming own order when user clicks in outcome list and will consume their own order.

image

https://github.com/AugurProject/augur/issues/6795


andrewdamelio (approved):
LGTM!!

Issue #6751, opened 2020-03-13
forking modals are missing processing tags


phoebemirman:
Migrating and releasing Rep buttons should get processing buttons so user knows when the tx is going through or failed

matt-bullock:
done. closing

PR #6807, opened 2020-03-17
add processing tag for migrating rep button

phoebemirman:
https://github.com/AugurProject/augur/issues/6751


bthaile (approved):
looks good. processing button makes things easy!

Issue #6801, opened 2020-03-17
Trading tutorial is skipping the last step and jumping back to the welcome screen


matt-bullock:
After clicking next on the my fills tooltip, it should go to the positions tooltip and then the 'Congratulations on making your first test trade!' screen. Instead it skips this and goes back to the 'Welcome to our test market' screen.

matt-bullock:
fixed. closing

PR #6808, opened 2020-03-17
fix trading tutorial skipping step

phoebemirman:
https://github.com/AugurProject/augur/issues/6801


bthaile (approved):
oh, trading tutorial gets left behind. good one

PR #6819, opened 2020-03-18
Add compression to warp sync

justinbarry:
Trying again with a significantly more populare npm package. Confirmed the UI loads with this one.


pgebheim (approved):
Failing test -- but otherwise code looks good!

PR #6822, opened 2020-03-18
bunch of UI gas gsn updates/fixes

andrewdamelio:
- when GSN is enabled have AugurSDK use GSN gasPrice
- show current GSN gas price in UI / remove gasPrice controls
- add getEthForDaiRate to contractCalls
- update ethToDaiRate/gasPrice each block
- fix/clean up displayGasInDai -- was returning incorrect gas estimates in DAI
- fix incorrect order expiration


packages/augur-ui/src/modules/events/actions/log-handlers.ts
```diff
@@ -195,6 +197,10 @@ export const handleNewBlockLog = (log: Events.NewBlock) => async (
       loadAnalytics(getState().analytics, blockchain.currentAugurTimestamp)
     );
   }
+  // update ethToDaiRate/gasPrice each block
+  dispatch(getEthToDaiRate());
+  dispatch(registerUserDefinedGasPriceFunction());
```
bthaile, line 202: are we going to allow the user to define a gas price?

bthaile (approved):
lgtm, we'll need to figure out what to show user if we add a power user mode and user uses ETH as transaction fee. I think we are mostly covered for that scenario but it hasn't been tested.

PR #6825, opened 2020-03-18
add checkbox to notification

phoebemirman:


bthaile:
already seen notifications. Checkbox didn't check when clicked

image

bthaile (approved):
looks good, I'll have a follow up PR when emitting market updated when liquidity gets updated in SDK.

Issue #6609, opened 2020-03-05
Sports Betting - Update style of Depth Chart


JohnDanz:
https://www.figma.com/file/42QoTLQcbObu7esoTPjWRv/Augur-Bet---Design-System?node-id=1034%3A93

![Screen Shot 2020-03-05 at 12.28.06 PM.png](https://images.zenhubusercontent.com/57155267bf93871ccd99a48b/6b21c432-0165-4fe7-aecb-150974ed8c76)

we need to update the depthChart to have the same colors/style as noted here. Don't worry about changing functionality or anything like that, just update the colors from green/orange to red/blue colors and white background, updated text styling etc.

PR #6828, opened 2020-03-19
Sports Betting - Update style of Depth Chart

rubyweb3:
https://github.com/AugurProject/augur/issues/6609

Screen Shot 2020-03-19 at 7 31 29 PM
Screen Shot 2020-03-19 at 7 31 23 PM
Screen Shot 2020-03-19 at 7 31 15 PM

### DIFF
--- a/packages/augur-ui/src/modules/market-charts/components/depth/depth.tsx
+++ b/packages/augur-ui/src/modules/market-charts/components/depth/depth.tsx
@@ -3,11 +3,12 @@ import * as d3 from 'd3';
 import ReactFauxDOM from 'react-faux-dom';
 import memoize from 'memoizee';
 import { createBigNumber } from 'utils/create-big-number';
-import { ASKS, BIDS, BUY, SELL, ZERO } from 'modules/common/constants';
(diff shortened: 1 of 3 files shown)


packages/augur-ui/src/modules/market-charts/components/depth/depth.tsx
```diff
@@ -3,11 +3,12 @@ import * as d3 from 'd3';
 import ReactFauxDOM from 'react-faux-dom';
 import memoize from 'memoizee';
 import { createBigNumber } from 'utils/create-big-number';
-import { ASKS, BIDS, BUY, SELL, ZERO } from 'modules/common/constants';
+import { ASKS, BIDS, BUY, SELL, ZERO, THEMES } from 'modules/common/constants';
```
JohnDanz: Revert all the changes to depth.tsx, this ticket is for styling only and shouldn't touch this file. I know this probably needs changes but it's going to be a separate ticket because it will require re-drawing on theme switch among other things.

JohnDanz (changes requested):

JohnDanz (approved):

Issue #6809, opened 2020-03-17
Manually entering a trade into the order ticket needs to trigger "Consuming own order"


Chwy5:
If user has an order on the book/outcome table and they manually enter an order into the order ticket, they currently can consume their own order....need to trigger the "Consume own Order" flag in this scenerio

Chwy5:
fixed - closing

PR #6832, opened 2020-03-19
moved self trade to sdk, pass in takerAddress to simulate trade

bthaile:
https://github.com/AugurProject/augur/issues/6809

was able to remove UI calculation for selfTrade now that it has moved to sdk.


nuevoalex (approved):
👍

PR #6833, opened 2020-03-19
Cache 'latest' getBlockByNumber calls to reduce load on eth node

justinbarry:


packages/augur-sdk/src/state/create-api.ts
```diff
@@ -120,7 +120,7 @@ export async function createClient(
     connector,
     zeroX,
     enableFlexSearch
-  );
+  );{}
```
nuevoalex, line 123: intentional?

packages/ethersjs-provider/src/EthersProvider.ts
```diff
@@ -58,17 +58,43 @@ export class EthersProvider extends ethers.providers.BaseProvider
...
     this.performQueue = queue(
       (item: PerformQueueTask, callback: (err: Error, results: any) => void) => {
         const _this = this;
         retry(
           { times, interval },
           async () => {
-            if (item.message === "send") {
-              return await _this.providerSend(item.params.method, item.params.params);
+            if (item.message === 'send') {
+              const {method, params: [blocktag, includeHeaders]} = item.params;
```
nuevoalex, line 70: can this all go in a function so the core queue processing is still easy to read?

nuevoalex (approved):
Two minor things but the code looks good!

PR #6847, opened 2020-03-19
Pg/0x 9.2.1, Clean up Terraform

pgebheim:


robert-e-davidson3:
Should update the 0x-docker flash script to use the same 0x version. Or delete that script, if we don't want to use it going forward.

robert-e-davidson3 (approved):

Issue #6627, opened 2020-03-06
Visual Issues with compact disputing cards


matt-bullock:
Icons need to sit bottom left so they don't overlap with total dispute stake

Outcome listed under tentative should be primary text colour. The grey looks odd

Build
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/394b21cd-4ce5-4b97-aee1-e717419714e2)

Design
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/0d07f3ae-e986-4b25-a832-b7af009d2f09)

matt-bullock:
looks good. closing

Issue #6240, opened 2020-02-17
categorical market cards look broken in compact view


matt-bullock:
Vol and event expiration should be at the bottom

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/e45268dd-f321-49f9-aefa-71a869494d66)

matt-bullock:
these still look broken on local and kovan. I pulled the latest release today. Let me know if  it's not been merged yet and I'll check again

matt-bullock:
@phoebemirman I'm still seeing this on local. I set it up fresh today but maybe I'm doing something wrong?

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/a4b9b3be-07a1-423b-8cca-e65734049236)

matt-bullock:
closing

PR #6857, opened 2020-03-20
fix compact market cards

phoebemirman:
https://github.com/AugurProject/augur/issues/6240

https://github.com/AugurProject/augur/issues/6627


bthaile (approved):
lgtm

Issue #6824, opened 2020-03-18
gsn: initializing account - add processing and confirmed button states 


matt-bullock:
https://www.figma.com/file/fLWVwmanAwetVZbujQquEi/Market-Page?node-id=3881%3A0

Disabled button looks faint on the light background but it's only a one off

![image](https://user-images.githubusercontent.com/3970376/77080482-51ccb500-69c7-11ea-8abf-5f40e9b06c7d.png)

use the existing dismissible banner.

matt-bullock:
looks good. closing

PR #6861, opened 2020-03-20
update init wallet button

andrewdamelio:
gsm: initializing account - add processing and confirmed button states #6824

Also fixes this issue:
Screen Shot 2020-03-19 at 4 29 48 PM


packages/augur-ui/src/modules/auth/actions/update-sdk.ts
```diff
@@ -81,10 +85,10 @@ export const createFundedGsnWallet = () => async (
   getState: () => AppState
 ) => {
   try {
+    dispatch(addUpdatePendingTransaction(CREATEAUGURWALLET, TXEventName.Pending));
```
bthaile, line 88: this can be taken care of in add-update-transactions if there is a method name to track. this isn't needed here
add to add-update-transactions
```
const ADD_PENDING_QUEUE_METHOD_CALLS = [
  BUYPARTICIPATIONTOKENS,
  REDEEMSTAKE,
  MIGRATE_FROM_LEG_REP_TOKEN,
  BATCHCANCELORDERS,
  TRADINGPROCEEDSCLAIMED,
  MIGRATEOUTBYPAYOUT,
  FORKANDREDEEM
];
```
collection and it will automatically be tracked.

packages/augur-ui/src/modules/events/actions/add-update-transaction.ts
```diff
@@ -103,6 +107,9 @@ export const addUpdateTransaction = (txStatus: Events.TXStatus) => async (
       methodCall !== CANCELORDER &&
       methodCall !== PUBLICFILLORDER
     ) {
+      if (methodCall === CREATEAUGURWALLET) {
```
bthaile: I don't think this is needed if the create augur wallet transaction is tracked in pending queue. see comment above.

bthaile (changes requested):
a few comments about tracking transaction and setting wallet status

bthaile (approved):
Lgtm

Issue #6631, opened 2020-03-06
Issues with market info on markets in reporting


matt-bullock:
Images from build. 

The Event expiration should be shown here with the arrow to expand to show more market details

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/627f3c2e-d6e0-490d-a26f-1d742ca06f6a)

Now when clicking the arrow you see two arrow. This shouldn't happen - only need the one arrow here

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/eaceac12-199b-4bf5-b0ad-66612730e053)

Clicking the arrow shows the spacing on the event expiration is broken:
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/7feac52c-80af-4199-aa3f-7e8058e87fc4)

Design:

https://www.figma.com/file/fLWVwmanAwetVZbujQquEi/Market-Page?node-id=449%3A953

If it's easier, show the event expiration and total volume as default but just make sure everything lines up

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/d923de43-d06c-4c2c-a7e0-5c2dca2ea320)

matt-bullock:
This is opening as an overlay on mobile. It should push content down like when you expand the market stats of the open markets @phoebemirman

matt-bullock:
done. closing

PR #6862, opened 2020-03-20
fix double arrows on market view reporting and spacing on expiration …

phoebemirman:
…time

https://github.com/AugurProject/augur/issues/6631


bthaile (approved):
lgtm

PR #6880, opened 2020-03-23
show processing for claim proceeds button on market trading page

bthaile:
image


phoebemirman (changes requested):
you shouldn't need tot include CLAIMMARKETSPROCEEDS there, you should change market header reporting to use market id instead of TRANSACTIONS

phoebemirman (approved):

Issue #6888, opened 2020-03-23
report form lets you report with no funds


phoebemirman:
should be a banner there

![Screen Shot 2020-03-23 at 1.10.53 PM.png](https://images.zenhubusercontent.com/574e216bbf93871ccd9be5fb/77f6e63a-7ea3-48ab-9ba2-9b8149403eb1)![Screen Shot 2020-03-23 at 1.11.01 PM.png](https://images.zenhubusercontent.com/574e216bbf93871ccd9be5fb/a605cd10-cc9d-48ba-b74b-824a4c6dad28)

Chwy5:
@matt-bullock Should we have a message box appear saying Like we do for trading:

Insufficient Balance 
 You do not.......

Chwy5:
Was this for Open reporting or a Designated reporter, reporting on their market.  Anyone can report in open reporting even if they do not own rep, but they would still need dai to pay gas costs

matt-bullock:
This looks like it's for Open Reporting where the user would only need dai for the tx cost. The form looks somewhat confusing because here because it's totalling up the initial reporter stake and the pre-filled stake which isn't correct as you don't pay anything for the initial reporter winning stake. The pre-filled stake box is also not needed here.

This is the original design. We could follow this for clarity and add a line about the winning stake to the grey text underneath 'Outcomes'
![image](https://user-images.githubusercontent.com/39496225/77922830-d9f16b00-72a1-11ea-96a7-a4668940dcb3.png)

matt-bullock:
~@phoebemirman can you confirm if the no funds referred to REP or DAI~ Just checked and it's letting me report when I have no dai to cover the tx fee because the ETH in my account is covering it

matt-bullock:
Now has message 'Insufficient Funds to complete transaction'. Closing

PR #6891, opened 2020-03-23
add message for not enough funds in reporting form

phoebemirman:
https://github.com/AugurProject/augur/issues/6888


bthaile (approved):
lgtm

Issue #6892, opened 2020-03-24
GSN wallet initialisation: Make the buy sell button active as soon as initialization is confirmed.


matt-bullock:
Not wait for the user to close the message

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/4e1c49b8-0554-4030-888d-915886fa806a)

matt-bullock:
I'm no longer seeing the processing or confirmed states

![Kapture 2020-03-26 at 12.58.43.gif](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/662a5be3-f23c-44c1-8609-eb9648377980)

Getting this error after trying to place the order. Not sure if the two are linked

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/e556666c-0493-4bf4-923b-56c0b777a870)

andrewdamelio:
@matt-bullock give this another test now on Kovan -- let me know if its working as expected.

matt-bullock:
@andrewdamelio need to wait for v2 to be up and running again

matt-bullock:
@andrewdamelio it's working ok on Local, does that mean it's good to close?

matt-bullock:
Closing this as it's working how it should

PR #6898, opened 2020-03-24
set CREATED on confirmation

andrewdamelio:
GSN wallet initialisation: Make the buy sell button active as soon as initialization is confirmed. #6892


bthaile (approved):
Lgtm

PR #6899, opened 2020-03-24
One big commit instead of several smaller ones

bconfortin:


packages/augur-ui/src/main.tsx
```diff
@@ -5,7 +5,7 @@ import { HashRouter } from 'react-router-dom';
 
 import MainErrorBoundary from 'modules/app/components/main-error-boundary';
 
-import store from 'store';
+import store from 'appStore';
```
bthaile, line 8: why did `store` change to `appStore` ?

also store is an alias in webpack.common.config.js
`      store: path.resolve(PATHS.APP, "store"),`

bthaile (changes requested):
I need a good reason why 'store' changed to 'appStore'

packages/augur-ui/src/main.tsx
bconfortin: `store` changed to `appStore` because it was in conflict with the package named `store` from npm: <https://www.npmjs.com/package/store>

This was one of the errors as to why 3box wasn't working. They use the `store` npm package but the name resolution for the package was in conflict with our folder. In a zoom call, me, Justin and John tested and agreed in changing it to `appStore` because it's the store from the app, not from the npm package.

bthaile: does alias need to change in webpack common file?

bconfortin: When Justin made the change, he removed the following aliases:

```
      assets: path.resolve(PATHS.APP, "assets"),
      config: path.resolve(PATHS.APP, "config"),
      modules: path.resolve(PATHS.APP, "modules"),
      reducers: path.resolve(PATHS.APP, "reducers"),
      services: path.resolve(PATHS.APP, "services"),
      store: path.resolve(PATHS.APP, "store"),
      utils: path.resolve(PATHS.APP, "utils"),
```

I believe he removed the other ones because they weren't being used.

bthaile (approved):
lgtm

PR #6912, opened 2020-03-24
Optimize Webpack module strategy + 0x

### LINKED ISSUE #6871: User should be able to use app before 0x wasm bundle is loaded
To accomplish this:

- [ ] Switch to the 0x-mesh-browser-lite module 
- [ ] Include the wasm bundle we want in the augur src tree && make sure webpack copies it into build
- [ ] After warp sync is done / when bulk sync starts, also load the browser mesh by using WebAssembly.instantiateStreaming(...); This should already be doable just by doing this in createBrowserMesh from browser-mesh.ts -- after it is loaded, instantiate like normal

pgebheim:
Here I am optimizing some of the Webpack build settings in order to
package vendor javascript (node_modules) separately from core augur
code. Additionally, this fixes #6871 by using browser-mesh-lite and
initiating the wasm loader at the time the mesh browser is started.


packages/augur-ui/config/webpack.common.config.js
```diff
@@ -58,17 +61,18 @@ module.exports = {
         test: /\.tsx?$/,
         loader: "ts-loader",
         options: {
+          experimentalWatchApi: true,
```
pgebheim, line 64: Note: This makes webpack use the same watch api as tsc, which works a lot better.

justinbarry (approved):
LGTM. We need to put the word out about the node upgrade.

PR #6913, opened 2020-03-24
dont call approveToTrade for GSN wallets

andrewdamelio:
If GSN is enabled, don't call approveAll before certain actions. There is no need to do this for GSN since the approval will happen when the wallet gets initialized during its first transactions.

this applies to wallets in "FUNDED_NEED_CREATE" state.

### DIFF
--- a/packages/augur-ui/src/modules/contracts/actions/contractCalls.ts
+++ b/packages/augur-ui/src/modules/contracts/actions/contractCalls.ts
@@ -306,13 +306,12 @@ export function getDai() {
   return contracts.cashFaucet.faucet(new BigNumber('1000000000000000000000'));
 }
 
-export function fundGsnWallet(gsnWalletAddress: string) {
+export function fundGsnWallet() {
   const amount = new BigNumber('1000000000000000000000');
   const { contracts } = augurSdk.get();
 
   augurSdk.client.setUseRelay(false);
   contracts.cashFaucet.faucet(amount).then(() => {
-    contracts.cash.transfer(gsnWalletAddress, amount);
     augurSdk.client.setUseRelay(true);
   })
 }
--- a/packages/augur-ui/src/modules/contracts/actions/contractCalls.ts
+++ b/packages/augur-ui/src/modules/contracts/actions/contractCalls.ts
@@ -306,12 +306,13 @@ export function getDai() {
   return contracts.cashFaucet.faucet(new BigNumber('1000000000000000000000'));
 }
 
-export function fundGsnWallet() {
+export function fundGsnWallet(gsnWalletAddress: string) {
   const amount = new BigNumber('1000000000000000000000');
   const { contracts } = augurSdk.get();
 
   augurSdk.client.setUseRelay(false);
   contracts.cashFaucet.faucet(amount).then(() => {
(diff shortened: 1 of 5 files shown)


packages/augur-ui/src/modules/contracts/actions/contractCalls.ts
```diff
@@ -306,13 +306,12 @@ export function getDai() {
   return contracts.cashFaucet.faucet(new BigNumber('1000000000000000000000'));
 }
 
-export function fundGsnWallet(gsnWalletAddress: string) {
+export function fundGsnWallet() {
   const amount = new BigNumber('1000000000000000000000');
   const { contracts } = augurSdk.get();
 
   augurSdk.client.setUseRelay(false);
   contracts.cashFaucet.faucet(amount).then(() => {
```
bthaile: prob need to get rid of `cashRaucet` call since it won't work on mainnet.

packages/augur-ui/src/modules/contracts/actions/contractCalls.ts
```diff
@@ -306,13 +306,12 @@ export function getDai() {
...
-    contracts.cash.transfer(gsnWalletAddress, amount);
```
bthaile: this is needed for testing on kovan, when signing wallet has ETH but no DAI. this button faucets DAI and sends to GSN Wallet

packages/augur-ui/src/modules/modal/containers/modal-gsn-faucet.ts
```diff
@@ -21,7 +21,7 @@ const mapStateToProps = (state: AppState) => {
 
 const mapDispatchToProps = (dispatch: ThunkDispatch<void, any, Action>) => ({
   closeModal: () => dispatch(closeModal()),
-  fundGsnWallet: (address) => fundGsnWallet(address),
+  fundGsnWallet: () => fundGsnWallet(),
```
bthaile: suggest leaving this

packages/augur-ui/src/modules/contracts/actions/contractCalls.ts
```diff
@@ -306,13 +306,12 @@ export function getDai() {
   return contracts.cashFaucet.faucet(new BigNumber('1000000000000000000000'));
 }
 
-export function fundGsnWallet(gsnWalletAddress: string) {
+export function fundGsnWallet() {
```
bthaile: suggest leaving this b/c of comment below

bthaile (changes requested):
suggest leaving the DAI faucet and transfer. Needed for signing wallets that have ETH but no cash on kovan.

packages/augur-ui/src/modules/contracts/actions/contractCalls.ts
andrewdamelio: Tested, seemed to work fine without this call

bthaile: Did you point UI at kovan. 
paste address next to link to kovan faucet in Transactions section of Account Summary
then when the button activates click it to transfer cash to GSN wallet?

bthaile (approved):
lgtm

Issue #6895, opened 2020-03-24
PnL isnt being updated correctly on a resolved market


Chwy5:
I have a losing position in a resolved market yet my PnL shows 0.00. I am long 30 of "yes" at .25....The market is Resolved as "No"....My realized PnL is showing 0.00

![3-24 starting balance on resolve outcome.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/daa202e5-4e13-463f-b415-030317658e36)

there was an existing bid on that book for 10 yes at .1...which i sold, then my Pnl updated, but still not correct
shouldnt my unrealized be included in realized since this market is resolved

![3-24updated.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/2977f94b-c847-44b9-ba96-07a942092fcc)

Also noticing the unrealized calculation is being done off the last trade price, which it shouldnt.....it should be based (in this example off 0), since the market is resolved

bthaile:
We'll need to add losing positions to user's realized P/L

Chwy5:
Still doesnt appear to be working....
realized doesnt seem to be working....I was short the over, 100 at .50 and the Over was the winning outcome ( I should have a negative realized)
![3-29real.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/b715cdbb-4281-4101-9a64-2bb60dc9401c)


![3-29r2.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/9c7bbf64-421e-4b4f-978e-7399b7eed5e9)

Chwy5:
still off

Chwy5:
Still not working....
Here is my starting balances in a resolved market (look at my 30-day starting balance) - The winning outcome for the market is "No"
![4-1start.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/f3c3479a-5cc1-4f83-9eec-91bae578eee8)

I am now buying 10 shares of outcome "No" at .31 
![4-1A.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/0f182bb0-bf3a-4b38-a68f-20f586f7bcff)

Expected behavior is my 30day P/L should increase (Since I am buying is at .31 and it is resolved at 1.0)
except my 30day decreases
[...]

PR #6914, opened 2020-03-24
Unrealized to realized finalized

bthaile:
https://github.com/AugurProject/augur/issues/6895
when market is actually finalized show negative realized P/L for lose positions. 

image

1. push market to awaiting finalization.
2. put ask orders on losing outcome
3. have user take orders to have a long position on losing outcome. 
4. verify that P/L isn't updated because market has not been finalized.
5. have user with winning shares claim to force finalization of the market.
6. verify the user with losing shares shows a negative P/L in core stats.


andrewdamelio (approved):
Nice Job!

Issue #6925, opened 2020-03-25
back button not working on this page (for entertainment templates)


phoebemirman:
![Screen Shot 2020-03-25 at 11.51.34 AM.png](https://images.zenhubusercontent.com/574e216bbf93871ccd9be5fb/71ddea63-5c58-4a9c-b6e1-fc373866bdfc)

Chwy5:
fixed closing

Issue #6903, opened 2020-03-24
Category tier system update in templates that use "Event"


Chwy5:
In our current templates the 3rd Category sort should be the Event:

For example: [Wimbledon} Men's tennis; Andre Agassi vs Pete Sampres
1) Sports
2) Tennis
3) Wimbledon

phoebemirman:
this was added for the first tennis template

Chwy5:
Can we do it for the Multiple Choice ones as well.....make the "Event" the 3rd category

phoebemirman:
sure

Chwy5:
Can we do this for Golf and Horse Racing as well?

Chwy5:
fixed - closing

PR #6926, opened 2020-03-25
add mechanism to set categories based on template input

phoebemirman:
https://github.com/AugurProject/augur/issues/6903
https://github.com/AugurProject/augur/issues/6925


bthaile (approved):
lgtm

PR #6930, opened 2020-03-25
only show tx/approval banner if gsn is disable

andrewdamelio:
disable the below for GSN during market creation review
image


bthaile (approved):
lgtm

PR #6932, opened 2020-03-26
fork takes market not marketinfo

robert-e-davidson3:


bthaile (approved):
oops, missed these tests and so did CI. interesting

PR #6943, opened 2020-03-26
Fix race condition order form

bthaile:
fix order form issues. 
1. populate order form then manually clear quantity then change price. order form didn't updated quantity.
2. race condition when user fills in quantity and types price fast. 

scenario below user created bid at 0.20 price then gets consuming own error creating an ask order. that is correct for ask price 0.20 but not for price 0.25.


phoebemirman (changes requested):
Screen Shot 2020-03-26 at 12 31 20 PM
still seeing this

1. fill order form
2. clear quantity
3. edit price
4. add quantity
5. delete quantity

or

1. fill order form
2. edit price so there is error 
3. Type a non-error price
4. delete quantity


Also seeing the quantity get auto filled to a non valid number sometimes

Screen Shot 2020-03-26 at 12 36 45 PM

bthaile:
I talked to Chwy about the non-valid quantity and it's fine. User can click on one of the buttons to get a valid quantity.

phoebemirman (changes requested):
seems input behavior changed. won't let me delete quantity in the form

bthaile:
can't totally delete the input value. Lets get this in and I'll run the behavior with Chwy on v2.augur.net

phoebemirman (approved):

PR #6945, opened 2020-03-26
fix leading zeros

andrewdamelio:
image


bthaile:
why is there a leading zero?

andrewdamelio:
@bthaile  html input fields treat 02 === 2 so its valid number in the input field with type=number.  Doing a Number on the value beforehand just trims it away if the user provides it.

bthaile (approved):
lgtm

PR #6947, opened 2020-03-26
Prune Augur DB

justinbarry:
Modify warp sync process to: 

1. If the db is empty, load current hash.
    1. load curent hash
    2. Make note of current hash in db.
2. Check most recently loaded warp sync hash
    1. if the most recently synced block is older than 7 days. Proceed as if DB is empty.
    2. if more recent, bulk sync to head
    3. Calculate current warp sync and compare to reported value


pgebheim:
@justinbarry -- One test failing, `getAccountTimeRangedStats`

https://dev.azure.com/augurproject/augur/_build/results?buildId=16819&view=logs&j=11227f36-ea87-561e-ddf4-b897aed732fb&t=f2a058a2-bad7-5784-fef5-4a0602b99cf8&l=1527

packages/augur-sdk/src/state/db/DB.ts
```diff
@@ -265,6 +397,55 @@ export class DB {
     await this.marketDatabase.sync(highestAvailableBlockNumber);
   }
 
+  async prune(timestamp: number) {
+    // Discover the markets that whose time has come.
```
nuevoalex, line 401: remove "that"

packages/augur-sdk/src/state/db/DB.ts
```diff
@@ -265,6 +397,55 @@ export class DB {
...
+    // This should probably be calculated in the constructor.
+    const dbsToRemoveMarketsFrom = this.genericEventDBDescriptions
+      .filter(({ indexes, primaryKey }) =>
+        [...indexes, primaryKey].includes('market')
+      )
+      .map(({ EventName }) => EventName)
+      .filter((eventName) => eventName !== 'ShareTokenBalanceChanged');
+
+    await Promise.all(
+      [...dbsToRemoveMarketsFrom, 'Markets', 'MarketVolumeChangedRollup', 'MarketOIChangedRollup'].map(dbName =>
```
nuevoalex, line 419: Why do these need to be added explicitly? It looks like they should be included in `dbsToRemoveMarketsFrom` already

packages/augur-sdk/src/state/db/DB.ts
```diff
@@ -265,6 +397,55 @@ export class DB {
...
+        this[dbName]
+          .where('market')
+          .anyOf(marketIdsToRemove)
+          .delete()
+      )
+    );
+
+    // Now to deal with 'ShareTokenBalanceChanged' logs.
+    const rollupShareTokenBalanceChangedLogs = await this.ShareTokenBalanceChangedRollup.where(
```
nuevoalex, line 428: Why do we need to do special handling of the `ShareTokenBalanceChanged` table? Shouldn't just deleting anything with the matching market as normal be ok?

For the rollup this could also just be a delete on a `whereBetween` clause without querying any other table couldn't it?

nuevoalex (approved):
Some questions / suggestions around the actual prune, otherwise lgtm

packages/augur-sdk/src/state/db/DB.ts
justinbarry: The derived/Rollup dbs aren't part of the genericEventDBDescriptions which is what we are using to generate this list.

Issue #6961, opened 2020-03-27
Forking market looks visually broken on mobile


matt-bullock:
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/dcdf6074-0ffb-49a9-b3b3-10d37cfe7567)

matt-bullock:
Done. closing

Issue #6960, opened 2020-03-27
Forking message: remove huge spacing above message. 24px spacing above and below message should work


matt-bullock:
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/56c3b789-3c30-494a-9eb8-eb34bef1e10b)

matt-bullock:
done. closing

Issue #6959, opened 2020-03-27
Forking form: Move forking message above invalid market rules. Forking message should be top priority


matt-bullock:
It should use the same background styling as the invalid rules, same width, padding etc. 16px spacing between these two content boxes

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/87c64678-4602-444f-ae3e-963de7d90ffa)

matt-bullock:
done. closing

PR #6982, opened 2020-03-27
fix forking cards and banner on mobile, update forking form

phoebemirman:
https://github.com/AugurProject/augur/issues/6961
https://github.com/AugurProject/augur/issues/6960
https://github.com/AugurProject/augur/issues/6959


bthaile (approved):
lgtm

Issue #6958, opened 2020-03-27
Getting Liquidity Depleted Filter and Not seeing market show up in 10% spread filter that passes Lig spread filter in market creation


Chwy5:
Create Team A vs Team B Categorical Market
in pre -liquidity add
Buy 10 shares of team A at .6
Buy 10 shares of team B at .3
Buy 10 shares of tie/no winner at .1

create market....Liquidity depleted notification shows in My created Markets and cant find market on the market pg (though not sure if it should pass the 10% filter since no asks)

joeykrug:
> (though not sure if it should pass the 10% filter since no asks)

Markets without asks should pass the filter. It has to have a bid and ask on one outcome or more, or bids on all outcomes (ignoring invalid), or asks on all outcomes (ignoring invalid) to pass it.

10 shares is far too small for the fees, would need at least 100 or so, maybe more if gas price is super high.

I'd bet the problem here is that market creation is using some different algo or something or not accounting for fees perhaps? But the regular market list does? @andrewdamelio tagging too

Chwy5:
I think it is passing the market creation filter b/c adding pre-liquidity doesnt have a tx fee associated with it....It is free to place open orders (the tx fee in market creation gets tied to the market being created)

Chwy5:
still seeing this occur.....

If I buy up all the outcomes (except invalid) had all bids add up to 1.0, buying 10 shares each, market passed liq filter in market creation, and then can only be seen by Filtering by All in the liq spread on the market trading pg

Chwy5:
fixed - now fails market creation like it should - closing

PR #6983, opened 2020-03-27
different liquidity spread calculations for pre-liquidity and liquidity spread.

bthaile:
move filtering orders based on gas trade cost to `getLiquidityForSpread` it was in marketDb.ts.

This way market creation form using `getMarketLiquidityRanking` will return `hasLiquidity` using filtered orders just like marketDB.ts calculates liquidity spread.

https://github.com/AugurProject/augur/issues/6958


nuevoalex (approved):
🐐

Issue #6979, opened 2020-03-27
Bug trying to go back after you save a template


Chwy5:
I created a crypto yes/no market...entered the deatils into the template and saved it.  Then hit the back button on the bottom left of the screen

![3-27cry.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/b89d0e75-a8a4-454b-843c-a98f751ab884)

phoebemirman:
only seeing this on v2

Chwy5:
fixed - closing

PR #6986, opened 2020-03-27
dont clear when going back

phoebemirman:
https://github.com/AugurProject/augur/issues/6979


bthaile (approved):
let's see if this fixes v2.augur.net

Issue #6920, opened 2020-03-25
UI-Archived Finalized Market data (30-day)


Chwy5:
UI Changes Needed:

1.  In Header: Realized PnL change to 30 day PnL

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/c2e1a5e3-7e64-4d60-b6b5-ba778d1b035d)
---
2.  Acct Sum Pg:
        A. Remove  All-time
        B. Change 1 week to 7 days
        C. Change 1 month to 30 Days

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/dd50eb9a-7919-4003-930e-8b1dd86884b7)

---
3. Icon + Tooltip needed for resolved market cards when 30days has passed

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/0b2b8bcd-ffd4-4615-9aa1-f935e9f98718)
        A. Add it to Market trading pg - After 30 days add icon (with hover over icon" Data only saved for 30 days~, hit {Button}to retrieve it~"
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/6347eaa2-7bba-4494-ac23-5838127a18b1)
        B. Acct Sum Pg (if in watchlist)
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/4df3964a-5287-4a40-addc-bb77f0c35c2c)
        C. Portfolio pg

positions
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/b82e9a5a-ff3f-46ad-9420-7eb9c561cdac)

My created markets

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/d8782968-cd75-4c49-b65f-7da2449aabd3)

---

4. Transaction pg - Need messaging with button "Data only saved for 30 days after market finalized, ~ click here to retrieve archived data~"

https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=11880%3A0
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/12dc3532-3eba-408f-8374-9b5f0a6fdc45)

5.  Null State needed for Market trading pg.....(Trade History, Charts )

https://www.figma.com/file/fLWVwmanAwetVZbujQquEi/Market-Page?node-id=3905%3A194993

[...]

PR #7009, opened 2020-03-30
30day archive

phoebemirman:
https://github.com/AugurProject/augur/issues/6920

still have to do market page null state


packages/augur-ui/src/utils/convert-marketInfo-marketData.ts
```diff
@@ -40,6 +44,7 @@ export function convertMarketInfoToMarketData(
     endTimeFormatted: convertUnixToFormattedDate(marketInfo.endTime),
     creationTimeFormatted: convertUnixToFormattedDate(marketInfo.creationTime),
     categories: marketInfo.categories,
+    isArchived: archivedDuration && (Math.abs(archivedDuration.asDays()) > ARCHIVED_MARKET_LENGTH),
```
bthaile: I think conditional is off. if duration is in whole days then the market could get pruned between day 30 and 31 but the condition would only set isArchived after day 31.

easy fix would be `>=` or use seconds so when 30 days + 1 second elapses we'd set the archived flag.

bthaile (approved):
lgtm

Issue #7040, opened 2020-03-31
fills not showing up in market page


phoebemirman:
fills in portfolio:
![Screen Shot 2020-03-31 at 12.14.47 PM.png](https://images.zenhubusercontent.com/574e216bbf93871ccd9be5fb/7588eb9f-b79d-4ed7-964c-0bfa25b73795)

market page:
![Screen Shot 2020-03-31 at 12.14.40 PM.png](https://images.zenhubusercontent.com/574e216bbf93871ccd9be5fb/29b35ca6-8442-466c-a7da-a47074e12447)

market on kovan: 
Week 10: Which NFL Team will win: Denver Broncos vs. Pittsburgh Steelers?
Estimated scheduled start time: Mar 31, 2020 11:00 AM (PDT)
0xD80C49977C5a5F51CaD969fa1236f8566f2247ca

Chwy5:
fixed

Issue #6120, opened 2020-02-12
Mobile tooltips should scroll with the page


matt-bullock:
What they're currently doing on build:

![Kapture 2020-02-12 at 15.47.45.gif](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/e53ca3c0-9eb8-47df-87ed-17019d0630d3)

matt-bullock:
@phoebemirman Can't test on mobile due to v2 being down but I still see this on my up to date local when emulating mobile

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/4a840f7d-879f-49ee-b2d5-f82fc034231f)

matt-bullock:
They're not moving in line with the scrolling and instead disappear but I think this is fine. closing

PR #7042, opened 2020-03-31
fix fills table and have tooltips disappear on mobile scroll

phoebemirman:
https://github.com/AugurProject/augur/issues/6120
https://github.com/AugurProject/augur/issues/7040


bthaile (approved):
lgtm

Issue #6169, opened 2020-02-13
Category filters are resetting when user visits a market and returns via the back button


matt-bullock:
user filters by category > clicks on a market > clicks the back button > categories should still be active rather than cleared

How to reproduce

![Kapture 2020-02-13 at 11.45.53.gif](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/70fc7569-bdd6-4f8b-8c45-b4492e091636)

matt-bullock:
done. closing

PR #7044, opened 2020-03-31
keep selected categories around when navving around app

phoebemirman:
https://github.com/AugurProject/augur/issues/6169


bthaile (approved):
lgtm

Issue #7051, opened 2020-04-01
Wrong color shield on templates


Chwy5:
Need to confirm all tempaltes have the correct color shields

Manual input of outcomes (for Multiple choice) get Yellow shields

![4-1shield.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/6c0fa403-5c52-4697-8a0e-667cb0adcf3b)

Chwy5:
fixed - closing

PR #7059, opened 2020-04-01
give categorical markets with user input yellow shields

phoebemirman:
https://github.com/AugurProject/augur/issues/7051


bthaile (approved):
lgtm

PR #7061, opened 2020-04-01
Fix the block range chunking for syncing

pgebheim:


packages/augur-sdk/src/state/sync/BulkSyncStrategy.ts
```diff
@@ -20,30 +21,24 @@ export class BulkSyncStrategy extends AbstractSyncStrategy implements SyncStrate
     endBlockNumber: number,
   ): Promise<number> {
     if(!endBlockNumber) throw new Error('Cannot bulk sync forever. Please pass in an endblock');
+    if(startBlockNumber >= endBlockNumber) throw new Error('Starting point of bulk sync must be less than the endpoint')
```
justinbarry: Is there a convoluted set of circumstances where these could be *equal*?

packages/augur-utils/src/generators.ts
```diff
@@ -0,0 +1,5 @@
+export function* chunkRange(start, end, chunk) {
+   for (;start < end;start+=chunk) {
```
justinbarry:
```suggestion
   for (;start =< end;start+=chunk) {
```

image
The top has this patch, the second is as written here.

justinbarry (changes requested):
One dumb question and one feedback item.

packages/augur-sdk/src/state/sync/BulkSyncStrategy.ts
pgebheim: Perhaps? I suppose with your change below that would be fine

pgebheim:
```suggestion
    if(startBlockNumber > endBlockNumber) throw new Error('Starting point of bulk sync must be less than or equal to the endpoint')
```

justinbarry (approved):
LGTM

PR #7087, opened 2020-04-02
detect when user has completely losing position and subtract frozen f…

bthaile:
…unds of that position

 - when market is finalized and user has complete losing positions don't include frozen funds in display value in UI. 
 - added console logs for payment in dai and eth for debugging purposes. 
 - fixed flash script creating orders for scalar markets, bid and ask was backwards.

### DIFF
--- a/packages/augur-sdk/src/state/getter/Users.ts
+++ b/packages/augur-sdk/src/state/getter/Users.ts
@@ -965,9 +965,29 @@ export class Users {
       _.values(_.mapValues(profitLossResultsByMarketAndOutcome, _.values))
     );
 
+    const marketFinalizedResults = await db.MarketFinalized.where('market')
+      .anyOf(_.map(allProfitLossResults, 'market'))
+      .toArray();
+    const marketFinalizedByMarket = _.keyBy(marketFinalizedResults, 'market');
+    const marketsData = await db.Markets.where('market')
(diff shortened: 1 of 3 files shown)


packages/augur-sdk/src/state/getter/Users.ts
```diff
@@ -965,9 +965,29 @@ export class Users {
       _.values(_.mapValues(profitLossResultsByMarketAndOutcome, _.values))
     );
 
+    const marketFinalizedResults = await db.MarketFinalized.where('market')
+      .anyOf(_.map(allProfitLossResults, 'market'))
+      .toArray();
+    const marketFinalizedByMarket = _.keyBy(marketFinalizedResults, 'market');
+    const marketsData = await db.Markets.where('market')
+      .anyOf(_.map(allProfitLossResults, 'market'))
```
nuevoalex: this could just be the smaller set of markets from `marketFinalizedResults` instead I think

bthaile: ` .anyOf(_.map(allProfitLossResults, 'market'))` filter finalized market by profit and loss markets.
ah! I see need to use finalized markets instead. will change, 👍

nuevoalex (approved):
One suggestion otherwise LGTM

bthaile:
change made, thx 👍

PR #7098, opened 2020-04-03
Adjusted layout for other themes

bconfortin:
You have to enable 3box in order to test it.

Go to packages/augur-artifacts/src/environments/local.json and add the following lines of code to that json:

```
"plugins": {
  "chat": "3box",
  "comments": "3box"
}
```

### DIFF
--- a/packages/augur-ui/src/assets/styles/_colors.less
+++ b/packages/augur-ui/src/assets/styles/_colors.less
@@ -19,7 +19,7 @@
   --color-primary-text-40: fade(#D7DDE0, 40%);
   --color-primary-text-80: fade(#D7DDE0, 80%);
   --color-primary-text-90: fade(#D7DDE0, 90%);
-  
+
   --color-secondary-text: #9599A5;
   --color-secondary-text-30: fade(#9599A5, 30%);
   --color-secondary-text-50: fade(#9599A5, 50%);
@@ -125,6 +125,10 @@
   --color-outcome-six-40: fade(#FFF966, 40%);
   --color-outcome-seven: #FFFFFF;
   --color-outcome-seven-40: fade(#FFFFFF, 40%);
+
+  // 3box Comments
+  --color-upvote: #7EC9FF;
+  --color-downvote: #FD6266;
 }
 
 :root[theme="BETTING"],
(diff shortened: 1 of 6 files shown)


packages/augur-ui/src/assets/styles/_colors.less
```diff
@@ -156,7 +160,7 @@
   --color-module-background-action-focus: #FFFFFF; // no equivalent
   --color-page-background: #F5F5F8;
   --color-page-background-80: fade(#F5F5F8, 80%);
-  --color-dark-grey: #D1D1D9;
```
JohnDanz: remove these spaces ;)

packages/augur-ui/src/modules/global-chat/components/global-chat.styles.less
```diff
@@ -1,111 +1,138 @@
...
-.GlobalChat {
-  background: var(--color-dark-grey);
-  bottom: 0;
-  position: fixed;
-  right: 0;
-  z-index: @mask-modal;
-
-  > button {
-    background-color: var(--color-page-background);
+:root {
```
JohnDanz: Don't need these to be wrapped in Root, root is default -- so trading classes can remain the same except for colors `var(--` stuff. 

The betting/sports changes can remain below.

packages/augur-ui/src/modules/global-chat/components/global-chat.tsx
```diff
@@ -28,7 +30,7 @@ export const GlobalChat = ({ provider, whichChatPlugin }: GlobalChatProps) => {
           mute
           popupChat
           showEmoji
-          colorTheme="#0E0E0F"
+          colorTheme={theme !== THEMES.TRADING ? '#D1D1D9' : '#0E0E0F'}
```
JohnDanz, line 33: Is there any better way to do this with just CSS? we should avoid hard coding styling hex values into the html

packages/augur-ui/src/modules/market/components/market-view/market-view.styles.less
```diff
@@ -1,702 +1,709 @@
...
-.MarketView {
-  display: flex;
-  flex-direction: column;
-  flex-grow: 1;
-
-  &.Inactive {
-    pointer-events: none;
-    min-width: 100%;
-    padding-bottom: 40px;
+:root {
```
JohnDanz: remove the `:root` wrapping, this can go back to the way it was

JohnDanz (changes requested):
couple changes, we dont need to wrap things in `:root` we only need that for specific changes that appear only in betting/sports

packages/augur-ui/src/modules/global-chat/components/global-chat.tsx
bconfortin: Will submit a PR to 3box and will fix it as soon as it's published.

JohnDanz (approved):

Issue #7095, opened 2020-04-03
Issues with redeeming stake


matt-bullock:
account 1: staked 6251 rep on outcome Yes

Account 2: staked 3125 on outcome No

Winner: Outcome No

Screenshot shows the amount staked for each account
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/2aedb689-38d7-4f6d-b57b-63458bbb617d)

---
Account 1 gets a claim stake and fee notification. The modal only shows the users original stake. It is missing the stake they win from the losing side:

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/0bc71d56-e0ed-48de-bac9-b74470a6e798)

---

After account 1 claims, this is their REP balance: 12,500.9618

If I'm right it should be 3748.9307rep (available balance) + 6251 (Reclaiming their rep staked) + 3125 (the losers dispute stake) which should equal roughly 13124 rep

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/c6b636af-e919-4a7f-a385-d4ddcb599f37)

----
The stake redeemed figure in the bell is also incorrect
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/bd5ee2a8-3e85-45f2-a528-bb58bbad1f22)

phoebemirman:
the REP amount you end up with seems right 3748.9307+6251.4507+(6251.4507*.4)=12,550

but the alert seems off

matt-bullock:
Forgot it was only 40% ROI. Amount you end up with is correct then, but modal and bell message aren't showing the correct figure

matt-bullock:
@phoebemirman what I'm seeing in the claim stake and fees modal is still different from the bell alert

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/4eb89674-c300-4429-854a-2e7577c51cf4)

phoebemirman:
couldn't get those exact numbers but saw similar issue that I fixed. see if it helps

matt-bullock:
Correct stake amount and correct bell alert figures showing now. closing

PR #7099, opened 2020-04-03
load staked markets

phoebemirman:
part of https://github.com/AugurProject/augur/issues/7095

### DIFF
--- a/packages/augur-sdk/src/state/getter/Users.ts
+++ b/packages/augur-sdk/src/state/getter/Users.ts
@@ -273,14 +273,14 @@ export class Users {
       universe: params.universe,
     });
 
-    const stakedRepMarketIds = [];
+    let stakedRepMarketIds = [];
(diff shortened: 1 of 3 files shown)


bthaile:
There is a better approach. If data is missing from `await Augur.getUserAccountData` call then we need to add it into the SDK getter. the whole point of `getUserAccountData` is to reduce multiple calls to sdk getters.

bthaile (changes requested):
There is a better approach. If data is missing from `await Augur.getUserAccountData` call then add the needed data in the getter. The whole point of `getUserAccountData` is to reduce multiple calls to sdk getters. 


 `getUserAccountData` getter internally calls the same getter as loadAccountReportingHistory.

```
    userStakedRep = await Accounts.getAccountRepStakeSummary(augur, db, {
      account: params.account,
      universe: params.universe,
    });
```


All markets info should be returned from the getter call.

bthaile (approved):
Lgtm

PR #7103, opened 2020-04-03
remove total lost frozen funds when markets are finalized, leave froz…

bthaile:
…en funds for positions user can claim on

calculate profit based on winning payout numerators for the finalized market. 
if profit is less than frozen funds ignore the market.


nuevoalex (approved):
🥇

PR #7104, opened 2020-04-03
Bruno/3box optimization attempt

bconfortin:
The current 3box instance has been added to redux store, meaning that, once you load their chat, their comments should load much faster. Instead of around 10 seconds, it should be loading in 1-2 seconds.

You have to enable 3box in order to test it.

Go to `packages/augur-artifacts/src/environments/local.json` and add the following lines of code to that json:

```
"plugins": {
  "chat": "3box",
  "comments": "3box"
}
```


packages/augur-ui/src/reducers/index.ts
```diff
@@ -122,4 +124,5 @@ export interface AppStateInterface {
   reportingListState: ReportingListState;
   analytics: Analytics;
   orderBooks: OrderBooks;
+  initialized3box: any;
```
bthaile, line 127: I'd like to see typing. most other reducers have typing.

bthaile (approved):
nit pick but I'd think it's important to use typing for the easy stuff.

PR #7109, opened 2020-04-04
fixed realized calc when user hasn't claimed and market is finalized

bthaile:
fixed realized calc when user hasn't claimed and market is finalized.

fixed issue where array wan't getting combined correctly with user's REP staked on reporting and disputing  marketIds.


justinbarry (approved):
LGMT

Issue #6977, opened 2020-03-27
Sports Betting - My Account - Page Layout


JohnDanz:
https://www.figma.com/file/wogMuUDylgOHgTEYhZ3jgp/Augur-Bet---UI?node-id=3275%3A132693

![Screen Shot 2020-03-27 at 12.45.34 PM.png](https://images.zenhubusercontent.com/57155267bf93871ccd99a48b/f92dc8a5-8400-4c31-806b-26f1af341148)

Update the overall layout of the My Account page in betting-ui and sportsbook. This may require rendering overview/status separately if the theme isn't trading. This ticket should be done last in this epic most likely.

PR #7113, opened 2020-04-06
Sports Betting - My Account - Page Layout

rubyweb3:
https://github.com/AugurProject/augur/issues/6977

Screen Shot 2020-04-09 at 6 58 07 PM
Screen Shot 2020-04-09 at 6 57 55 PM
Screen Shot 2020-04-09 at 6 57 42 PM

### DIFF
--- a/packages/augur-ui/src/assets/styles/_typography.less
+++ b/packages/augur-ui/src/assets/styles/_typography.less
@@ -193,10 +193,12 @@
 @size-326: 20.375rem; // 326px
 @size-340: 21.25rem; // 340px
 @size-350: 21.875rem; // 350px
+@size-360: 22.5rem; // 360px
 @size-400: 25rem; // 400px
 @size-480: 30rem; // 480px
 @size-550: 34.375rem; // 550px
 @size-560: 35rem; // 560px
+@size-600: // 600px
 @size-720: 45rem; // 720px
 @size-744: 46.5rem; // 744px
 @size-750: 46.875rem; // 750px
--- a/packages/augur-ui/src/assets/styles/_typography.less
+++ b/packages/augur-ui/src/assets/styles/_typography.less
@@ -200,7 +200,6 @@
 @size-560: 35rem; // 560px
 @size-600: 37.5rem; // 600px
 @size-720: 45rem; // 720px
-@size-744: 46.5rem; // 744px
 @size-750: 46.875rem; // 750px
 @size-775: 48.4375rem; // 775px
 @size-800: 50rem; // 800px
--- a/packages/augur-ui/src/assets/styles/_typography.less
+++ b/packages/augur-ui/src/assets/styles/_typography.less
@@ -196,14 +196,11 @@
 @size-360: 22.5rem; // 360px
 @size-400: 25rem; // 400px
 @size-480: 30rem; // 480px
-@size-550: 34.375rem; // 550px
 @size-560: 35rem; // 560px
-@size-600: 37.5rem; // 600px
 @size-720: 45rem; // 720px
 @size-750: 46.875rem; // 750px
 @size-775: 48.4375rem; // 775px
 @size-800: 50rem; // 800px
-@size-1400: 87.5rem; // 1400px
 
 // font-weights
 @weight-light: 300; // 300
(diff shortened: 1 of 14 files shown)


packages/augur-ui/src/assets/styles/_typography.less
```diff
@@ -192,13 +193,18 @@
 @size-326: 20.375rem; // 326px
 @size-340: 21.25rem; // 340px
 @size-350: 21.875rem; // 350px
+@size-360: 22.5rem; // 360px
 @size-400: 25rem; // 400px
 @size-480: 30rem; // 480px
+@size-550: 34.375rem; // 550px
 @size-560: 35rem; // 560px
+@size-600: // 600px
```
JohnDanz: missing a value here

packages/augur-ui/src/modules/account/components/account-view.styles.less
```diff
@@ -140,10 +140,33 @@
...
+      margin: @size-32;
+    }
+  }
+
   .AccountView {
+    margin: @size-8 @size-32;
+    display: flex;
+    flex-flow: column wrap;
+    align-content: space-between;
+    min-height: @size-1400;
```
JohnDanz: I would prefer to use grid layout for this page, flex is workable but grid is preferred for layouts that are more complex like this one.

packages/augur-ui/src/modules/account/components/account-view.styles.less
```diff
@@ -170,23 +191,50 @@
...
+    }
+
+    > footer {
+      min-height: @size-40;
     }
   }
+
+  /* Force new columns */
+  .AccountView::before,
+  .AccountView::after {
```
JohnDanz: I would rather use grid to avoid things like this

JohnDanz (changes requested):
Let's switch to grid for the layout, also remember we will be adding another list of items (Created Markets) so grid will allow us to more easily integrate a new component then trying to re-adjust with flex.

packages/augur-ui/src/assets/styles/_typography.less
rubyweb3: Updated.

packages/augur-ui/src/modules/account/components/account-view.styles.less
rubyweb3: Removed.

packages/augur-ui/src/modules/account/components/account-view.styles.less
rubyweb3: Updated to grid layout. Please let me know what you think about it. Thanks.

packages/augur-ui/src/modules/account/components/account-view.styles.less
```diff
@@ -140,53 +140,85 @@
 
 :root[theme="BETTING"],
 :root[theme="SPORTS"] {
-  .AccountView {
-    > div {
-      border: @size-1 solid var(--color-outline);
-      border-radius: @border-radius-rounded;
+  .AccountViewWrap {
```
JohnDanz: we shouldn't need `AccountViewWrap`, the HTML should change very little in `account-view.tsx` in terms of its structure. we would want ot just modify `AccountView`.

packages/augur-ui/src/modules/account/components/account-view.tsx
```diff
@@ -113,6 +115,35 @@ export default class AccountView extends React.Component<
...
+                    <Overview hideHeader={false} />
+                    <AugurStatus hideHeader={false} />
+                  </div>
+                  <div>
+                    <Transactions />
+                    <Favorites />
+                  </div>
+                </div>
+                <TermsAndConditions />
+              </div>
```
JohnDanz, line 134: I'd simplify this to just a div with `AccountView` and everything inside of it with no added divs for row separation. Use grid to place the pieces where they belong.

JohnDanz (changes requested):

packages/augur-ui/src/modules/account/components/account-view.styles.less
rubyweb3: Removed.

packages/augur-ui/src/modules/account/components/account-view.tsx
rubyweb3: Updated. Some UIs have not been updated yet. I will work on it.

JohnDanz (approved):

Issue #7084, opened 2020-04-02
user should see an explation of how transaction fees work during initial transaction


pgebheim:
Currently, during the "Initialize Wallet" transaction, we pop up a modal explaining initializing a wallet means + why there is a fee to do so.

We will need to extend this modal in the following ways:
- [x] It should show up on the FIRST transaction that anyone ever does (initialize wallet or anything else).
- [x] It should have the following text (or similar)
> Augur is a peer-to-peer system, and certain actions require paying a small fee to other users of the system. The cost of these fees will be included in the total fees displayed when taking that action. Trades, Creating Markets, and Reporting on the market outcome are examples of such actions.
> Augur will **reserve** $N of your funds in order to pay these fees, but your total balance can be cashed out at any time. To see the total amount reserved for fees, click on the Account menu.

pgebheim:
From @Chwy5 

Yeah we should do this now, the UX is a little confusing
1)We should remove the gsn wallet number from the dropdown and just say Account and show the phone number or email address and use Account metamask/Web3 if using that. (can still get the acct number through add funds)
2) We should add Eth Reserve (?) in the Account Menu
3) Add tooltip copy: Augur will reserve $(N) of your funds in order to pay these fees, but your total balance(Available Dai + Eth Reserves) can be cashed out at any time.

andrewdamelio:
@pgebheim do we know what this value will be? Is it static or will i need to get this from a helper contract method.

```reserve $N ```

pgebheim:
Currently when doing a txn it is passed in -- and right now its set to 0.1 ETH.

@nuevoalex -- Where's that constant defined? @andrewdamelio -- we should use the constant*eth2daiRate as that $N

nuevoalex:
It's defined in ContractDependenciesGSN currently as a top level constant

matt-bullock:
All looks good when I create an open order. But if I take an order off the books I'm not seeing the popup modal

andrewdamelio:
@matt-bullock can you retest the above and let me know if its still an issue on your end?

> All looks good when I create an open order. But if I take an order off the books I'm not seeing the popup modal

matt-bullock:
@andrewdamelio no longer seeing that issue. Closing out

PR #7117, opened 2020-04-06
show init wallet modal on first tx

andrewdamelio:
### user should see an explanation  of how transaction fees work during initial transaction #7084

- It should show up on the FIRST transaction that anyone ever does (initialize wallet or anything else).
 -- create market
 -- take trade 
 -- Dispute
 -- Report
 -- Buy PP

- It should have the following text (or similar)
image


bthaile (approved):
more random pop ups. ok, for now. I think we need a better design.

Issue #7074, opened 2020-04-02
Initializing account needs a bell message


matt-bullock:

matt-bullock:
done. closing

PR #7125, opened 2020-04-06
add bell message for init wallet

andrewdamelio:
Initializing account needs a bell message #7074

@bthaile let me know if there is a better way to setup alerts 🤔


bthaile (approved):
lgtm

Issue #5034, opened 2019-12-04
Orderbook empty states - add offer/bid buttons on MOBILE


matt-bullock:
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/73aebfc6-586b-4ffc-8a38-990dbd4bbbed)

https://www.figma.com/file/fLWVwmanAwetVZbujQquEi/Market-Page?node-id=2435%3A157294

- [ ] Replace current text with buttons

- [ ] pressing these will bring the order ticket into view

matt-bullock:
Behaviour seems weird here. I would expect hitting the button to force up the order ticket into view with the qty input active. Not to have to then manually scroll down and see error messages 

![Kapture 2020-02-11 at 12.30.36.gif](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/64f4fb1b-acf7-4aaa-9394-94e0552341a0)

joeykrug:
fix as part of launch imo

matt-bullock:
clicking the buttons aren't triggering anything  @phoebemirman

phoebemirman:
i'm seeing it scroll down. please double check

Chwy5:
Adding a bid worked fine, but if I click on Add Offer, it needs to enter a sell order....currently still has the order book on the buy side

matt-bullock:
It's scrolling down for me on mobile now. I was relying on desktop mobile emulator to test yesterday which doesn't work for some reason but I can't see this being an issue.

But as Chwy mentioned above , it isn't switching between buy and sell

Chwy5:
Fixed closing

PR #7129, opened 2020-04-06
don't show errors when scrolling orderbook into view

phoebemirman:
https://github.com/AugurProject/augur/issues/5034


bthaile (approved):
lgtm

PR #7130, opened 2020-04-06
show alerts for both disputing and reporting being redeemed

phoebemirman:
part of https://github.com/AugurProject/augur/issues/7095


bthaile (approved):
looks good. we could get REP sent to user's account if we need to. look at REP faucet and how that triggers an alert.

Issue #6997, opened 2020-03-29
Post Only Limit order Bug - api


Chwy5:
The post-only limit order option ensures the limit order will be added to the order book and not match with a pre-existing order. If your order would cause a match with a pre-existing order, your post-only limit order will be cancelled.

Post-Only Limit Order ensures that you will pay the maker fee and not the taker fee unless matched with a hidden order

PR #7135, opened 2020-04-06
add safe order creation method that will throw error if order crosses…

bthaile:
… the book or order price is outside of market price range

https://github.com/AugurProject/augur/issues/6997

create safe create order that throws Error


nuevoalex (approved):
👍

PR #7142, opened 2020-04-07
Tooltip disappearing on scroll, arrow up/down and touch screen

bconfortin:
#6120 
To have the tooltip follow the scroll would most likely be a laggy solution. Hiding it is a much cleaner and better performing choice.


bthaile (approved):
lgtm

pgebheim:
@Mergifyio rebase

Issue #7107, opened 2020-04-04
Claim Proceeds Modal needs tooltip 


Chwy5:
We need to add tooltip explaining where that number comes from, this tooltip is on claim market proceeds modal. 

tooltip:
"This number is the return of Frozen Funds for any position(s) held in this market + any PnL accrued in this market"

matt-bullock:
Tootip can go immediately after 'Proceeds after market fees'

matt-bullock:
looks good. closing

PR #7163, opened 2020-04-07
Added the tooltip for claiming proceeds

bconfortin:
#7107 

Preview
Captura de Tela 2020-04-07 às 18 35 59


bthaile (approved):
Lgtm, we’ll prob have to massage that tooltip language when we get more clear verbiage.

PR #7209, opened 2020-04-09
save selected universe in local storage

phoebemirman:


bthaile (approved):
I'd rather the user always starts at genesis universe and navigates to child, but at this point if they switch might as well start there when they return. 
In dev this can get to be a pain when the universe stored in localStorage doesn't exist b/c moving testnets.

PR #7216, opened 2020-04-09
audit funds contract and reduce pruning logic

### LINKED ISSUE #7162: Prune algo update
Current status
===

The prune algo queries the MarketFinalized logs for any that have a time stamp before the `PRUNE_HORIZON` which is currently 30 days previous to the current block's timestamp leaving just the most recent ShareTokenBalance logs.

The problem
===

We need some data for the UI to display that the user has funds to claim. This could be in the form of share tokens, disputing stake, or market creator / affiliate fees.

Solution
===
1. Make the prune delete all data of past-cutoff markets including share balance logs.
2. Provide a utility contract for taking an account and some list of markets which will it will scan and return data on what the user has to claim from them.
3. In the UI provide a basic UI to "recover old funds" or some such wording which will generate market ids and in chunks call this utility function to see what can be claimed.
4. UI presents basic info on claimable funds retrieved in (3) and button(s) to claim them.

nuevoalex:


packages/augur-core/tests/test_audit_funds.py
```diff
@@ -0,0 +1,192 @@
...
+from old_eth_utils import mk_contract_address
+
+def test_shares(kitchenSinkFixture, universe, cash, market, scalarMarket):
+    shareToken = kitchenSinkFixture.contracts['ShareToken']
+    auditFunds = kitchenSinkFixture.contracts['AuditFunds']
+
+    account = kitchenSinkFixture.accounts[0]
+
+    marketFactory = kitchenSinkFixture.contracts['MarketFactory']
+    marketAddress = mk_contract_address(marketFactory.address, 1).hex()
```
justinbarry: `marketAddress` doesn't appear to be used. Is this here to make a linter happy?

justinbarry (approved):
Absolute Magic. One comment because it is like signing a guest book. Just had to do it.

bthaile (approved):
lgtm

pgebheim:
@Mergifyio update

Issue #7083, opened 2020-04-02
User should be able to select transaction fee (again!)


pgebheim:
This should work just like it used to with no changes -- we just need it put back in.

Chwy5:
added - closing

PR #7228, opened 2020-04-10
add gasPrice selection/display back

andrewdamelio:
User should be able to select transaction fee (again!) #7083


bthaile (approved):
lgtm

Issue #7231, opened 2020-04-10
Confirmed button showing on claim stakes and fee modal before claiming the reporting stake: See images


matt-bullock:
Opened from notification:

Confirmed button already showing

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/a25c20a8-3230-4927-9f85-b9138d60aca4)

Clicking it makes it go away:

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/88cc6917-e552-471a-8ff1-1f4a37a3c2aa)

I had a winning position on the same market that I had winning REP. I claimed proceeds first
This is my bell history:

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/cfa1a5fd-cb1e-47cc-918b-0d2a2f845b75)

Note: maybe claiming on proceeds indirectly causes claiming stake and fees to show confirmed

bthaile:
pulling out marketIds is incorrect. Look at `getUserAccountData` method in Users.ts

![image.png](https://images.zenhubusercontent.com/5c00901a5cb43a45a18aeb7f/527afb5a-845a-49f6-b2a3-66a3fe73a4b1)


![image.png](https://images.zenhubusercontent.com/5c00901a5cb43a45a18aeb7f/16845fd2-6b4b-4605-a53d-f6c5e208ed2c)

phoebemirman:
it could also be from switching accounts

Chwy5:
not seeing this....closing

PR #7238, opened 2020-04-10
Clear pending

phoebemirman:
https://github.com/AugurProject/augur/issues/7231


packages/augur-ui/src/modules/pending-queue/reducers/pending-queue.ts
```diff
@@ -4,23 +4,27 @@ import {
...
+import { RESET_STATE } from "modules/app/actions/reset-state";
+import deepClone from "utils/deep-clone";
 
 const DEFAULT_STATE: PendingQueue = {};
 
 export default function(pendingQueue: PendingQueue = DEFAULT_STATE, { type, data }: BaseAction): PendingQueue {
   switch (type) {
     case ADD_PENDING_DATA: {
       const { pendingId, queueName, status, blockNumber, hash, info } = data;
+      let pending = deepClone<PendingQueue>(pendingQueue);
```
bthaile: why is a deep clone needed for every added pending queue item that comes in?

I'm not following why a deep clone is needed?

Isn't add clearing the state when user logs out or logs in good enough?

bthaile (commented):
.

bthaile (approved):
lgtm

Issue #7251, opened 2020-04-13
My Active Markets 24hr calculation needs to filter out resolved markets


Chwy5:
My Active Markets only displays users Open and In reporting Markets (pre-resolved), but the overall 24 hr calculation still includes markets that just got resolved into the calculation.  

Need to filter out the resolved markets from that calculation to make that number less confusing

Chwy5:
i think it is still calculating the resolved markets in

I was just showing 0.00 on 24hr and on the only market in there....I just went and did a trade on that market and this is the values I see

![4-1324.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/aefb21ef-652f-490b-b3b8-7e9434b5c611)

Chwy5:
closing

Issue #6919, opened 2020-03-25
add 'recently depleted liquidity' to the sort drop down in the my created markets section of portfolio


matt-bullock:
To allow user to bring all markets to the top with depleted liquidity in order of most recently depleted

Chwy5:
I see it added, but doesnt update when I change to it....I need to leave the page and return in order for it to update

Chwy5:
added - closing

PR #7252, opened 2020-04-13
Ignore finalized pre load history

bthaile:
https://github.com/AugurProject/augur/issues/6919
https://github.com/AugurProject/augur/issues/7251

1. filter markets in awaiting finalization and finalized when calculating user's 24 hour unrealized.
2. fix recently depleted sort in my markets section of portfolio.


bconfortin (approved):
Lgtm

Issue #7127, opened 2020-04-06
Initialized account via taking an order off the books but I'm still getting this confirmed message on every market I visit


matt-bullock:
It does go away once I dismiss it but it doesn't need to be there if user didn't follow the initialize account route via open order

![Kapture 2020-04-06 at 21.12.17.gif](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/855429ae-87d6-475e-9495-c06238c9f7a2)

matt-bullock:
Just tested this again. I'm still seeing the initialise message after already seeing the pop up when taking an order. You can see I have a position already

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/65791cfc-0d08-42c9-96c8-9d5a5d952060)

matt-bullock:
I'm no longer seeing the initialize account message. As soon as I transfer ETH to a fresh account, Augur takes $20 as reserve and doesn't notify me of this. I can create or taken an order without seeing the initialize account message @andrewdamelio

matt-bullock:
Looks good now. closing

PR #7258, opened 2020-04-13
hide notice if returning to trading page

andrewdamelio:
Initialized account via taking an order off the books but I'm still getting this confirmed message on every market I visit #7127

-- We will now remove the message (if it is shown) we when navigate to the trading page.


bthaile (approved):
lgtm

PR #7260, opened 2020-04-13
update walletStatus on TxSuccess

### LINKED ISSUE #7018: "Initialize Wallet" banner shows up on order form even if wallet was created via earlier create market
I created a market as the first thing the account did after funding the account. Market created ok, but initial liquidity kept failing. Tried to add open orders via the market page instead and saw that my account needed initialising.

andrewdamelio:
fixes: https://github.com/AugurProject/augur/issues/7018


bthaile (approved):
That was easy, lgtm

Issue #7197, opened 2020-04-09
Mobile Filters are not being saved


Chwy5:
On Mobile when I set my filters they should remain the same for the next time I come back on.  Currently I need to reset them to (where I like) every time I come onto my mobile acct.

Chwy5:
fixed - closing

PR #7263, opened 2020-04-13
Added filters to local storage

bconfortin:
#7197 

I programmed it so the filters from the query (address bar) take priority over saved filters from localStorage


bthaile:
look at `settings` it saved in localStorage already for user. 
image


touch base with Andy if you have any questions. I don't think we need a new object in localStorage

bthaile (changes requested):
see comment

bthaile (approved):
most excellent, look good to me.

PR #7273, opened 2020-04-14
Dump eth node call count every 60 seconds

### LINKED ISSUE #7229: Log number of eth calls by method every 60 seconds

justinbarry:
Closes #7229


pgebheim (changes requested):
I think we should have a config-level debug section that can turn stuff like this on and off. Other than that looks good!

Can you add something to the config to handle debugging info?

justinbarry:
ref #7277

pgebheim (approved):

pgebheim:
@Mergifyio update

Issue #6974, opened 2020-03-27
Sports Betting - My Accounts - My Favorites


JohnDanz:
https://www.figma.com/file/wogMuUDylgOHgTEYhZ3jgp/Augur-Bet---UI?node-id=3275%3A132693

![Screen Shot 2020-03-27 at 12.35.15 PM.png](https://images.zenhubusercontent.com/57155267bf93871ccd99a48b/b776f0c9-b852-4779-812e-774f3ad5e55e)

Update the My Favorites (Watchlist) component for betting ui and sports book to match this new style.

PR #7289, opened 2020-04-15
Sports Betting - My Accounts - My Favorites

rubyweb3:
https://github.com/AugurProject/augur/issues/6974

Screen Shot 2020-04-15 at 6 17 43 PM
Screen Shot 2020-04-15 at 6 17 34 PM


packages/augur-ui/src/modules/portfolio/components/favorites/favorites.styles.less
```diff
@@ -57,4 +57,29 @@
...
+        }
+      }
+    }
+
+    > div:last-of-type {
+      min-height: @size-414;
+    }
+  }
+
+  .Watchlist__EmptyDisplay {
```
rubyweb3: I find that the height of Favorites component will be different when it's empty and not empty. I have spent some time on the current implementation. What do you think about it?

JohnDanz: It looks fine to me however can we change the class name? we spent time to remove classes with `__` or `-` in their names. Maybe just call this `EmptyDisplay` or `WatchlistEmptyDisplay`

JohnDanz (changes requested):
minor request

packages/augur-ui/src/modules/portfolio/components/favorites/favorites.styles.less
rubyweb3: Thanks. Updated.

JohnDanz (approved):

Issue #7298, opened 2020-04-15
Transfer Modal Copy upgrade


Chwy5:
The new Transfer button which brings up the Modal still Says Withdraw funds in it

Need to change Withdraw to Transfer

Chwy5:
fixed - closing

PR #7303, opened 2020-04-15
update transfer from withdraw, so code base makes sense and update la…

bthaile:
…bels in modal

image

https://github.com/AugurProject/augur/issues/7298


andrewdamelio (approved):
Looks good

PR #7309, opened 2020-04-15
prune on warp sync boundaries

nuevoalex:


packages/augur-test/src/tests/warp/Pruning.test.ts
```diff
@@ -1,38 +1,50 @@
...
-  let mark: TestContractAPI;
+  let john: WarpTestContractApi;
+  let mary: WarpTestContractApi;
+  let mark: WarpTestContractApi;
   let seed: Seed;
 
   let market: ContractInterfaces.Market;
 
+  beforeAll(async () => {
+    ipfs = IPFS.create({
```
justinbarry: @nuevoalex Did this pass locally for you? IPFS is a fickle party to involve in the tests.

justinbarry (approved):
LGTM. Pending tests

packages/augur-test/src/tests/warp/Pruning.test.ts
nuevoalex: yup!

Issue #7304, opened 2020-04-15
warp sync market shows up randomly when searching in dispute page - remove from search


phoebemirman:
![Screen Shot 2020-04-15 at 12.38.31 PM.png](https://images.zenhubusercontent.com/574e216bbf93871ccd9be5fb/6cb817f4-d9bf-4ed9-a580-6b5003cd290d)

without a search:
![Screen Shot 2020-04-15 at 12.39.13 PM.png](https://images.zenhubusercontent.com/574e216bbf93871ccd9be5fb/f9fa6028-a886-4a06-ae52-d9d86c5fbe8b)

Chwy5:
fixed - closing

PR #7311, opened 2020-04-16
fix dispute search

phoebemirman:
https://github.com/AugurProject/augur/issues/7304


bthaile (approved):
cut and paste love it.

PR #7312, opened 2020-04-16
Move to Alchemy v2 endpoints

pgebheim:
Change all references to alchemy URLs to use their v2 endpoints. These
wont really be used in the UI but are still relevant for the infra that
is being run for demo.


justinbarry (approved):
LGTM

PR #7316, opened 2020-04-16
Add logger package

### LINKED ISSUE #7277: Consider: Implement Logger to help wrangle all the logs pre-launch
I'm not sure we want to do this now, but we at least need to wrangle the UI / SDK console.logs before launch.

Idea would be to use a logger library so we can set log levels in the config and then set up a sane set of levels.

justinbarry:
closes #7277


bthaile (approved):
lgtm

bthaile:
It would be beneficial if user can change logging level via web console using `AugurSDK` object.

justinbarry:
@bthaile Added a method called `setLoggerLevel` to the sdk.

bthaile:
`setLoggerLevel` is exactly what I was hoping for, 👍

Issue #7332, opened 2020-04-16
Trading Bug


Chwy5:
To create:

Buy an Outcome and spend nearly all of the users Available Funds. Then go and add an open order to sell out of the users whole position ( I actually get the error trying to sell any quantity greater than 40 shares out)....getting insufficient Dai error/  Should be using the users shares to sell out.

The order form isnt showing correct info as well....shows new position instead of Closing Out original position (see screenshot)

![4-16bug.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/6c946def-1804-4dee-b5f6-833bf807c750)

confirm section should say the user is closing out a position. 
It should not cost them any DAI so they should not get insufficient funds. 
Seems like user share balance isn't being passed to simulate trade. look in
`packages/augur-ui/src/modules/trades/actions/update-trade-cost-shares.ts`

Chwy5:
fixed - closing

PR #7334, opened 2020-04-16
fix shares

phoebemirman:
https://github.com/AugurProject/augur/issues/7332


bthaile:
Got a fix in. Closing

PR #7341, opened 2020-04-17
remove withdraw for affiliate fees.

### LINKED ISSUE #7338: Affiliate fees will be directly sent to affiliate during trading
In order to accomplish reasonable ux, fees will be sent to the affiliate during trading instead of accounting for them on chain and requiring claim calls.

This makes the feature cheaper for everyone, allows the UI to move forward with no ux changes, and doesn't have any attack surfaces not already present through scam markets.

Alex to do in RC2

nuevoalex:


bthaile (approved):
nice, auto withdraw for affiliate fees, if I'm reading this correctly. Deleting code is awesome!!!

PR #7359, opened 2020-04-17
P/L testcases

robert-e-davidson3:
Testcases from https://docs.google.com/spreadsheets/d/1ZOw3js_EKqyJiaemouVtFdksRKPjP_o681T6alii7Gk/edit?ts=5e8cdc13#gid=0

### DIFF
--- a/packages/augur-test/src/tests/state/getter/users/get-profit-loss-and-summary.test.ts
+++ b/packages/augur-test/src/tests/state/getter/users/get-profit-loss-and-summary.test.ts
@@ -5,24 +5,28 @@ import * as _ from 'lodash';
 import {
   _beforeAll,
   _beforeEach,
-  doTrade,
+  doTradeTakerView,
   LONG,
   SHORT,
-  YES,
-  MakerTakerTradeData,
-  ONE,
+  ONE, MakerTakerTrade, TradeData, PLResultData
 } from './common';
+import { formatBytes32String } from 'ethers/utils';
+import { Market } from "@augurproject/core/build/libraries/ContractInterfaces";
 
+const ZERO = new BigNumber(0);
+const HOUR = 60 * 60;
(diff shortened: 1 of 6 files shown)


packages/augur-tools/src/libs/TestEthersProvider.ts
```diff
@@ -25,9 +25,10 @@ export class TestEthersProvider extends EthersProvider {
     return makeProvider(seed, this.accounts);
   };
 
-  getConfig(): SDKConfiguration {
+  getConfig(overwrites?: RecursivePartial<SDKConfiguration>): SDKConfiguration {
```
pgebheim, line 28: Oh that's a nice API!

packages/augur-test/src/tests/state/getter/users/get-profit-loss-and-summary.test.ts
```diff
@@ -36,147 +40,502 @@ describe('State API :: Users :: ', () => {
...
+  await user.doInitialReport(market, outcomeList);
+}
+
+async function claimProceeds(market: Market, users: TestContractAPI[]) {
+  for (const user of users) {
+    await user.claimTradingProceeds(market);
+  }
+}
+
+function repeat<T>(thing: T, times: number): T[] {
```
pgebheim, line 535: You should put this in augur-utils :-)

pgebheim (approved):
One real comment to move the repeat<T> into augur-utils (I think we should start building this out)

packages/augur-test/src/tests/state/getter/users/get-profit-loss-and-summary.test.ts
robert-e-davidson3: good idea! https://github.com/AugurProject/augur/pull/7363

PR #7387, opened 2020-04-20
add some canned template markets to market creation canned markets, th…

bthaile:
…is is a start of the effort


packages/augur-tools/src/flash/create-canned-markets-and-orders.ts
```diff
@@ -139,6 +140,7 @@ export async function createCannedMarkets(person: ContractAPI, faucet=true): Pro
     const market = await createCannedMarket(person, can, faucet);
     markets.push(market);
   }
+  markets.concat(createTemplatedMarkets(person, faucet));
```
robert-e-davidson3: Array.concat does not alter the array. Maybe use push?
```suggestion
  markets.push(createTemplatedMarkets(person, faucet));
```

robert-e-davidson3 (changes requested):

packages/augur-tools/src/flash/create-canned-markets-and-orders.ts
bthaile: nice one

robert-e-davidson3 (approved):

Issue #902, opened 2019-01-26
SEO Updates: Augur UI


ghost:
## General: 

* [ ] ~~Test Google indexing after making these changes (+1 or 2 days later) by searching **"site:IPFS_ROOT_URL"** and see how markets and etc are being indexed, ensure descriptions, tags, titles and etc are reflecting these changes.~~ Test how google bot indexes by using the chrome devtools (Ctrl+P, network conditions; set user-agent to google bot), test against ipfs as well (v2-ipfs.augur.net points to the ipfs hash; try against cloudflare-ipfs.com)
* [x] Implement and deploy a sitemap that tells Google how to navigate all the URL paths the Augur webapp. 
* [x] Add meta tags that make the app mobile native when saved to mobile homescreen, and add the pop-over notification for mobile users to download the page to their homescreen for a native app experience. (@tk will take care of this and submit a PR).
* [x] Use / add the term "Prediction Markets" wherever you can, while still looking native, clean, and appropriate. Don't try to fit a round peg into a square hole, but use the term as frequently as possible as long as it makes sense. 
* [x] Review loading speed times for the in-app Javascript, figure out how to make it load faster (check with Google Pagespeed tester). 
* [x] See if its possible to add a paragraph of content (~500 words) to the homepage of the Augur UI. Maybe a pop-up prior to the disclaimer with a TL;DR of Augur and how it works, or a quick paragraph about how to get started. They dismiss this popup, get shown the disclaimer, then are free to use the app. Flexible on this, but ideally we have a paragraph of some content on the landing UI page. 
* [x] Figure out a method for not passing node parameters in the URL - but bundle them within the UI somehow, and have additional optional parameters for nodes. 
* [x] Use HelmetJS for any meta title, description, tags, etc updates -- probably the best method for doing this all. 
* [x] Reorganize and add missing headers (h1…h6) in all pages.

## Stable Loader
**The stable loader exists -- why would we want to pass in another IPFS hash? I think we really just need to validate that the stable loader page works**

* [x] Wrapper for the UI to check a reference of what the latest UI version is (ex: on Github somewhere) and have it load automatically (but support passing parameters to use a different IPFS hash), while still supporting URL parameters for nodes and links for the react stuff once it loads the actual UI by passing it all through. 

## Markets Page (markets' list):

* [x] Update page title (50–60 characters - " | Augur" adds 8 characters). Should be something like: _”Prediction markets for sports, politics, crypto and more | Augur”_
[...]

PR #7391, opened 2020-04-20
Specify max size of an asset so it splits

pgebheim:
This turns on code splitting which means we have a pile of resources now but I wanna see if this will unlock googlebots ability to crawl.

refs #902


justinbarry (approved):
LGTM

Issue #7361, opened 2020-04-17
Refresh dispute window stats when someone has purchased participation tokens


bthaile:
I noticed that when other person bought participation tokens my Participation section on Dispute page didn't update.

matt-bullock:
tested and working. closing

PR #7408, opened 2020-04-21
call to get dispute window stats if any participation tokens are purchased

bthaile:
call to get dispute window stats if any participation tokens are purchased

https://github.com/AugurProject/augur/issues/7361


andrewdamelio (approved):
LGTM!

PR #7413, opened 2020-04-21
created new format market shares method that take market type to know…

bthaile:
… what decimals to use in formatting

YesNo and Categorical markets don't need to display decimals for quantity. 
only scalar markets need to show decimal places for quantity

image

scalar

image

yesno/categorical

image

image


packages/augur-ui/src/modules/create-market/containers/initial-liquidity.ts
```diff
@@ -37,7 +36,7 @@ const mergeProps = (sP: any, dP: any, oP: any) => {
     {
       key: 'quantity',
       columnType: COLUMN_TYPES.VALUE,
-      value: formatShares(order.quantity, { decimals: 2, decimalsRounded: 2 }),
+      value: formatMarketShares(sP.market.markettype, order.quantity),
```
andrewdamelio: should this be marketType?

andrewdamelio (approved):

Issue #7409, opened 2020-04-21
Event expiration time bug


Chwy5:
This is happening for the muliple choice and scalar template for unemployment rate....

converted time doesnt match the preset time in the boxes though both are in UTC

![Screen Shot 2020-04-21 at 3.11.23 PM.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/20941b59-8ccf-447b-8ac9-e79a1c2657ea)

Chwy5:
fixed

PR #7415, opened 2020-04-21
fix moment calculation to get ned of next month of question month and…

bthaile:
… year

https://github.com/AugurProject/augur/issues/7409


andrewdamelio (approved):
Nice!

PR #7420, opened 2020-04-21
App status context updates

JohnDanz:
more app status flags added to the app-status context. I don't want to remove the redux versions of these values until we have finished the full transition to context for this category.


bthaile (approved):
lgtm

phoebemirman (approved):

Issue #7340, opened 2020-04-16
Performance - Adding Open Order gets stuck in processing state


Chwy5:
Have had this happen on multiple occasions....I place an open order and it gets stuck in processing state....I get this error...resfreshing removes the stuck processing but I need to retry the order

![4-16open.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/14f816a7-4993-459a-8d34-81fca74a462d)

Chwy5:
fixed - closing

PR #7433, opened 2020-04-22
include creating orders in the transaction timeout, include awaiting …

bthaile:
…signing as status to timeout as well.

https://github.com/AugurProject/augur/issues/7340


andrewdamelio (approved):
lgtm

PR #7448, opened 2020-04-23
using wrong property from state props in merge props

bthaile:
fix blow up in pre-liqudity from making quantity format changes.


bconfortin (approved):
Lgtm

Issue #7382, opened 2020-04-20
Sports Betting - My Bets - Responsive References


yanelbottini:
In the following link you can find references for all the different breakpoints for My Bets section:

![Screen Shot 2020-04-20 at 12.49.57.png](https://images.zenhubusercontent.com/5cc988ee77bc5b0505df9320/37147dcc-2d32-434f-a998-896dd8e4cc0b)

[Figma Link](https://www.figma.com/file/wogMuUDylgOHgTEYhZ3jgp/Augur-Bet-UI?node-id=4541%3A239131)

Issue #7377, opened 2020-04-20
Sports Betting - My Bets - View by Event Hover States


yanelbottini:
**Hover states** for the different elements:

![Screen Shot 2020-04-20 at 11.07.41.png](https://images.zenhubusercontent.com/5cc988ee77bc5b0505df9320/2da8f175-10b2-436c-8492-0dd30be390ec)

[Figma File](https://www.figma.com/file/42QoTLQcbObu7esoTPjWRv/Augur-Bet-Design-System?node-id=1140%3A0)

Please, remember to include hover state for the whole box (Shadow) as shown on the attached example.

Tooltip explaining what "High Risk" means should appear in the case o showing an Invalid Market:

![Screen Shot 2020-04-20 at 11.11.56.png](https://images.zenhubusercontent.com/5cc988ee77bc5b0505df9320/ed40f662-5deb-41bd-9037-c5d398c14046)

Issue #7376, opened 2020-04-20
Sports Betting - My Bets - View by Outcomes Hover States


yanelbottini:
This are the **Hover States** for View by Outcomes Table:

![Screen Shot 2020-04-20 at 10.58.26.png](https://images.zenhubusercontent.com/5cc988ee77bc5b0505df9320/032d9d3c-5e5b-4af2-85d9-94c82f6e7ee9)

[Figma Link](https://www.figma.com/file/42QoTLQcbObu7esoTPjWRv/Augur-Bet-Design-System?node-id=1140%3A0)

**Cashout Button style:**

![Screen Shot 2020-04-20 at 11.00.36.png](https://images.zenhubusercontent.com/5cc988ee77bc5b0505df9320/c3e24521-2db7-4d5c-8e8b-f8d61044977e)

[Figma Link](https://www.figma.com/file/42QoTLQcbObu7esoTPjWRv/Augur-Bet-Design-System?node-id=540%3A67)

PR #7449, opened 2020-04-23
My bets - add outcomes and style for mobile

phoebemirman:
closes https://github.com/AugurProject/augur/issues/7376
closes https://github.com/AugurProject/augur/issues/7377 
closes https://github.com/AugurProject/augur/issues/7382


packages/augur-ui/src/modules/portfolio/components/portfolio-view/my-bets.tsx
```diff
@@ -15,31 +19,62 @@ import {
...
+      setState({ ...state, rows: MOCK_GAMES_DATA });
+    } else if (
+      MY_BETS_VIEW_BY[viewBy].label === EVENT &&
+      SPORTS_MARKET_TYPES[selectedMarketCardType].label === FUTURES
+    ) {
+      setState({ ...state, rows: MOCK_FUTURES_DATA });
+    }
+  }, [selectedMarketCardType, viewBy]);
+
+  const showEvents = MY_BETS_VIEW_BY[viewBy].label === EVENT;
```
JohnDanz, line 100: should define this higher and reference it in the use effects so that we don't have to keep repeating it

packages/augur-ui/src/modules/portfolio/components/portfolio-view/my-bets.tsx
```diff
@@ -73,35 +110,69 @@ export const MyBets = () => {
...
+                defaultValue={MY_BETS_MARKET_STATUS[0].value}
+                onChange={selected =>
+                  setState({ ...state, marketStatus: selected })
+                }
+                minimalStyle
+              />
+            </span>
+          ) : (
+            <span>
+              Bet Date:{' '}
```
JohnDanz: can we use a template string to avoid having to add `{' '}`? I forget if that works I just hate those inserted spaces :)

packages/augur-ui/src/modules/portfolio/components/portfolio-view/my-bets.tsx
```diff
@@ -73,35 +110,69 @@ export const MyBets = () => {
...
         </div>
-        <PillSelection
-          options={SPORTS_MARKET_TYPES}
-          defaultSelection={selectedMarketCardType}
-          large
-          onChange={selected =>
-            setState({ ...state, selectedMarketCardType: selected })
-          }
-        />
+        {showEvents && (
```
JohnDanz, line 175: I would use one pill selector here for this and the bottom one, and instead change it to `options={showEvents ? SPORTS_MARKET_TYPES : MARKET_STATE_TYPES}`

packages/augur-ui/src/modules/portfolio/components/portfolio-view/my-bets.tsx
```diff
@@ -113,7 +184,15 @@ export const MyBets = () => {
             </span>
           </section>
         )}
-        {rows.length > 0 && rows.map(row => <Game row={row} type={SPORTS_MARKET_TYPES[selectedMarketCardType].label} />)}
+        {rows.length > 0 &&
```
JohnDanz: Don't need row.length > 0 if we are going to map, if you map on an empty array it won't do anything anyway so we can just do the showEvents and map

JohnDanz (changes requested):
couple little things

JohnDanz (approved):

PR #7452, opened 2020-04-23
SDK optimize contract calls

### LINKED ISSUE #6792: SDK Optimization: Minimize Contract Calls
We make a lot of read-only contract calls because it's easy in local dev or test environments but it doesn't scale well. We need to find and remove these calls wherever we can.

First we should try to get information from the SDK's database. Where that isn't possible, memoize.

- [ ] 3 net_version calls during sdk/ui initialization
- [x] Remove calls to contracts.isKnownUniverse_ and replace with DB call.

justinbarry:
Closes #6792


pgebheim:
@Mergifyio rebase

nuevoalex (approved):
👍

Issue #7444, opened 2020-04-23
Test trade completion issue


Chwy5:
Afte new user goes through test trade it takes them back to the same box (which would allow them to redo the test trade)

Should take them to the market trading list pg when test trade is complete

This box appears again:
![4-23testtrade.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/8c456540-6c6b-4718-80f7-265cd84bf251)

andrewdamelio:
@chwy fixed, -- this was actually a bug that was causing the Trading Tutorial to reset after step 11, instead of going to the end which shows the below modal.

![image](https://user-images.githubusercontent.com/1683736/80226336-3c782580-861a-11ea-80a3-333a26473b9d.png)

Chwy5:
Fixed - closing

PR #7462, opened 2020-04-24
fix tutorial bug

andrewdamelio:
Test trade completion issue #7444

fixed bug with Trading Tutorial, also adding smoke test to make sure the trading tutorial is able to be started/finished without issue.


pgebheim:
Refs #7444

bthaile (approved):
lgtm

PR #7480, opened 2020-04-27
Revert "Weird overflow + android address bar issue was hiding the button"

bconfortin:
Reverts AugurProject/augur#7392


bthaile (approved):
ok

Issue #7479, opened 2020-04-27
Torus: If user closes login modal, we display a modal without useful information.


pgebheim:
![image](https://user-images.githubusercontent.com/86010/80390462-4dff3e80-8861-11ea-9591-64a83976ff44.png)

Since this is an actual error case, we should either display something real here, or just go back to the normal page as if you never tried to log in. I'm pretty sure this is just a dev/debug message but it may need to be handled.

I think we should just assume this was a chosen close and not show any error.

@andrewdamelio can you remove?

andrewdamelio:
for all errors from the sdk or just the one about clicking decline?

pgebheim:
Specifically clicking decline / closing the window. Other SDK errors we should probably still show with some better copy above it.

something like:
```
There was an error while attempting to log in with {{ wallet }}. Please try again, and if it is still not working checkout the help center for logging in.

Error: {{ error_message }}
```

@Chwy5 thoughts?

Chwy5:
Sound good to me....I would leave it short

There was an error while attempting to log in with {{ wallet }}. Please try again.

PR #7504, opened 2020-04-28
update wallet error messages

andrewdamelio:
#7479

- if the users denies access to the sdk wallet don't display (user closed popup) error, just close the signin modal
- update error message template for all non deny access errors


bthaile (approved):
lgtm

Issue #7481, opened 2020-04-27
When filtering by all liquidity spreads, user should not see banner for potential invalid


pgebheim:
This message isn't true -- in this case users just see markets with bad spreads, but the invalid filter would still hide markets that appear invalid via the invalid orders filter.

![image](https://user-images.githubusercontent.com/86010/80392066-81db6380-8863-11ea-9eff-06ff37d988be.png)

pgebheim:
@andrewdamelio -- we need to confirm this (don't do until status is set to confirmed)

andrewdamelio:
@pgebheim To clarify, for this ticket should we just remove this notice for liquidity spread OR update the text to something else?

@Chwy5 Was there a reason we created this notice/text initially?

Chwy5:
Not that I recall.....I think we used it before we added the invalid spread filter

pgebheim:
@andrewdamelio We should not show the banner in this case.

PR #7516, opened 2020-04-28
remove liquidity spread notice

andrewdamelio:
When filtering by all liquidity spreads, user should not see banner for potential invalid #7481


bthaile (approved):
lgtm

Issue #7513, opened 2020-04-28
Mobile: Changes to menu and top bar (see chwys comment at bottom about logged out state)


matt-bullock:
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/fc2b9ba2-6d00-421c-8582-ce530b2ef440)

**Top bar:** [Design](https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=15345%3A271577)

- remove logo
- Remove funds
- Add account dropdown
___

**Account dropdown should look like this when active:** [Design](https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=15345%3A279990)

___

**Menu** [Design](https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=15345%3A270241)

- Add funds to top
- Update add funds button design
- Will need to add logo but hold off until we update logos everywhere
- Same as above for the app switcher

matt-bullock:
**Reduce the header bar to 48px height** [Design](https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=15345%3A271577)

Only for header bar,  not inside the hamburger menu as this still requires the height

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/1611bac2-ac2b-4764-882f-0cf2c56b4ac5)


**Logged out state** [Design](https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=11719%3A90)

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/3b7a4225-0625-402f-aacd-c04f18fd7adb)

matt-bullock:
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/0d72aba1-6ef1-4e2b-b5fe-f392eb8c8336)

[Design](https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=425%3A51)

Bell should sit no more than roughly 50% from the right of the screen. This means limiting the amount of characters displayed under account - whether this is the users email or wallet provider

Chwy5:
Issue with the logged out state- for both mobile and destop....should look like this...
[...]

PR #7518, opened 2020-04-28
update top bar and side nav for mobile

phoebemirman:
https://github.com/AugurProject/augur/issues/7513


bthaile:
desktop top bar got changed
image


needs to look the same as before
image

bthaile (changes requested):
mobile looks good, one issue on desktop

bthaile (approved):
lgtm

Issue #7510, opened 2020-04-28
Mobile: Market list page. Add a clear button next to the category header


matt-bullock:
If the user has filtered by a category then it should show the X (clear button) next to the category text. Tapping this will change the filter back to 'Popular Markets' and the text here will say 'Popular Markets'. No clear button needed when showing Popular Markets.

Give X button a good hit area of roughly 20x20px

[design](https://www.figma.com/file/GKhft5ZOwuH3ANrupGNCLHUU/Market-list-page?node-id=1331%3A3936)

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/c6dcc9de-cea6-48b8-ad7b-92913dad3c3a)

pgebheim:
I think we should push this into v2.1 -- nice to have but isn't stability/performance.

@matt-bullock @Chwy5

Chwy5:
We actually need it...I got into that state yesterday and there is no way out without getting out of the site or at least not easily found....there is no back button there

pgebheim:
What do you mean no way of getting out of the site? The user can click on "Categories and Filters" again, click "All Categories" and hit apply.

We need to be focusing on all the various stability fixes for this release, planning on UX improvements for just after v2 launch.

Chwy5:
you can get into it by not going into categories which i found out yesterday.  If You accidentally click on a tag from the market

pgebheim:
ahh... didn't realize that.

Still in terms of severity seems minor and could be pushed into after stability.

Chwy5:
yeah we can push

bthaile:
How about using this ticket to fix the back button when user clicks on category. Then user can get out and back to known screen.

confirmed that history object gets updated correctly and the back button does go to previous url correctly.  It doesn't trigger the UI to re-query to get previous market list. 

~~Adding a "X" wouldn't work, we'd still need to fix triggering previous URL loading.~~

after more investigation we could add "X", that would give us more control. We'd still need to solve the back button issue.  By solving the back button issue we don't need the "X".

Chwy5:
ok back button works....I want to add the x for post release...moving this ticket back to do later

bthaile:
@Chwy5  can you move this to RC3, post release?

Chwy5:
rc3 looks like it is pre launch stuff.....2.1 looks post

PR #7520, opened 2020-04-28
if query param for categories has changed and no categories but state…

bthaile:
… hasn't been updated reset categories.

https://github.com/AugurProject/augur/issues/7510

in market list, click on category in market card.
click back button and make sure market list refreshes with correct market list.
try clicking on first or second or third category
then back button.


andrewdamelio (approved):
Lgtm

PR #7521, opened 2020-04-28
App status context updates

JohnDanz:
Lots of updates, introduced exportable actions and getState for appStatus context. finished replacing redux appstatus and removed redux appstatus.


phoebemirman (changes requested):
getting error when going to market trading page in trading

phoebemirman (approved):

PR #7524, opened 2020-04-29
Prettify logger, sdk, tests, tools packages

justinbarry:


robert-e-davidson3 (approved):
approving for when this gets the conflicts fixed

justinbarry:
I'm moving so many things right now. This will be a source of pain for generations if we merged it.

Issue #3767, opened 2019-09-23
Publish all NPM modules to NPM with Release Versions


MicahZoltu:
**Is your feature request related to a problem? Please describe.**
I want to build something on the Augur SDK, but I don't want to or can't build from source.  I would like to just do `npm install @augurproject/sdk` and hit the ground running.

**Describe the solution you'd like**
```
cd packages/augur-sdk && npm publish --access public
```

**Describe alternatives you've considered**
None.  I **demand** NPM packages!

**Additional context**
I'm _really_ lazy.

pgebheim:
I feel yah micah!

We'll publish the whole monorepo as things come to launch.

pgebheim:
Re-opening 'cause Micah convinced me to start an epic for Launch Planning.

pgebheim:
We'll use `lerna publish` to publish all the parts of the mono-repo for this.

justinbarry:
In preparation we need to merge multiple modules together and put them all under the `@augurproject` org namespace.

Prospective package list:
===

@augurproject/artifacts
@augurproject/core
@augurproject/sdk
@augurproject/tools
@augurproject/types
@augurproject/utils
@augurproject/ethersjs-provider

These need to be moved/merged.
===
@augurproject/logger
blockstream-adapters

These can be merged together and then into the sdk module.
===

contract-dependencies-ethers
contract-dependencies-gsn
contract-dependencies

These should NOT be published
===
@augurproject/test 
@augurproject/ui

PR #7558, opened 2020-04-29
Prerelease cleanup

justinbarry:
This is part 1 of a two part series. Next step is actually publishing the packages.

ref #3767


packages/augur-test/src/tests/api/Trade.test.ts
```diff
@@ -39,15 +31,21 @@ test('Trade :: placeTrade', async () => {
     new BigNumber(0)
   );
 
+  console.log('1-checkpoint-1');
```
robert-e-davidson3, line 34: debug code?

packages/augur-test/src/tests/api/ZeroX.test.ts
```diff
@@ -76,9 +84,16 @@ describe('Augur API :: ZeroX :: ', () => {
         stringTo32ByteHex('B'),
       ]);
 
+      console.log('1-checkpoint-1');
```
robert-e-davidson3, line 87: more debug code?

packages/augur-tools/src/flash/ganache-scripts.ts
```diff
@@ -24,7 +24,7 @@ import { LogReplayerV1 } from './replay-logs-v1';
 import { MemDown } from 'memdown';
 import { sleep } from './util';
 
-export const defaultSeedPath = '/tmp/seed.json';
+export const defaultSeedPath = '/tmp/augur/seed.json';
```
robert-e-davidson3, line 27: does this work if `/tmp/augur` doesn't exist?

robert-e-davidson3 (approved):
couple of comments

packages/augur-tools/src/flash/ganache-scripts.ts
justinbarry: It does. I added a mkdir thing.

PR #7559, opened 2020-04-29
fixes for 7313 and 7512, plus some cleanup

### LINKED ISSUE #7313: Investigate: After withdrawal sdk state for useWallet and useRelay are not reverted
in `withDrawAllFunds` and `withdrawAllFundsEstimateGasCost` we calls `setUseWallet(false)` and `setUseRelay(false)` but never revert them back to their previous state.

This will likely cause a bug after someone withdraws.

Found by @robert-e-davidson3

### LINKED ISSUE #7512: When browser mesh not running, client attempts to use rpc even if its not configured
This happened on iOS.

image

robert-e-davidson3:


packages/augur-sdk/src/api/ZeroX.ts
```diff
@@ -521,14 +521,18 @@ export class ZeroX {
     try {
       if (this._mesh) {
         return this._mesh.addOrdersAsync(orders);
-      } else {
+      } else if (this._rpc) {
         return this._rpc.addOrdersAsync(orders);
+      } else {
+        throw Error('0x rpc and browser mesh are undefined so orders cannot be created')
```
robert-e-davidson3: This error message ends up being shown to the user so suggestions for better copy are appreciated.

packages/augur-sdk/src/state/create-api.ts
```diff
@@ -125,11 +125,11 @@ export async function createClient(
     connector,
     zeroX,
     enableFlexSearch
-  );{}
+  );
 
   // Delay loading of the browser mesh until we're finished syncing
   client.events.once(SubscriptionEventName.BulkSyncComplete, () => {
-    if (config.zeroX?.mesh?.enabled && createBrowserMesh) {
+    if (!config?.sdk?.enabled && config.zeroX?.mesh?.enabled && createBrowserMesh) {
```
robert-e-davidson3, line 132: We don't want to run the browser mesh if we're using the sdk server. This doesn't actually trigger as things stand because the BulkSyncComplete event doesn't get emitted when using the sdk server... but if that changes then this will prevent an error.

packages/augur-sdk/src/api/GSN.ts
```diff
@@ -64,13 +64,20 @@ export class GSN {
       const signerAddress = await this.augur.dependencies.signer.getAddress();
       const walletAddress = await this.calculateWalletAddress(signerAddress);
       const wallet = this.augur.contracts.augurWalletFromAddress(walletAddress);
+
+      const useWallet = this.augur.getUseWallet();
```
robert-e-davidson3, line 68: I didn't test this change as I'm not actually sure how to withdraw funds. @bthaile does this seem safe?

packages/augur-sdk/src/api/ZeroX.ts
justinbarry: Can we add something about *why* they are undefined?

nuevoalex (approved):
👍

packages/augur-sdk/src/api/ZeroX.ts
pgebheim:
```suggestion
        throw Error('Unable to fetch orders from ZeroX - The mesh has not yet been initialized or there is an error, and no RPC server was specified.')
```

pgebheim (changes requested):

pgebheim (approved):

Issue #7542, opened 2020-04-29
Finance template - market closed bug


Chwy5:
Not getting an error message when a market is closed for holidays (some do work others dont)

Noticed this for May 4th on Nikkei 225 index

To reproduce fill out template index template put in a date (which the market was closed) hit Next at the bottom...shouldnt let you proceed to next pg and should ahve an error message about trading on a holiday

also noticed this error message trigger for April 29th (but the holiday is April 30th)

Like For US markets (like NYSE) Monday May 25th (memorial day works) but I get the message for Christmas

bthaile:
how do you know that the exchange for `Nikkei` has a holiday?
What exchange is that?

Chwy5:
tokyo stock exchange - I looked it up

Like May 4th is Greenery Day for them
May 5th and 6th are 2 other holiday days for them too

Chwy5:
Still not working....Just tried Dec 25th for CAC-40 and Dax it lets me go forward
I put in May 1st (which is a holiday but shows me....

![4-30hol.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/5ca55c90-4341-4ece-a47d-e76e2cf7acea)

Still same issues from above as well

Chwy5:
Still has a bunch of holidays off

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/48c4e3fa-6b21-4290-b4aa-c2f3f9237bc8)

Chwy5:
Still not working found this in Dax (may 1 displaying wrong holiday) and June first doesnt bring up an error (june 1st is really their Whitmonday holiday)...we are displaying it on May 1

![5-1dax.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/13c28300-1124-47c6-8632-61ffeda3e47f)

Chwy5:
Closing this ticket and opening a new one with the specifics dates that need to be fixed/added

PR #7565, opened 2020-04-29
fix holiday comparison. use moment helper method for easy check.

bthaile:
https://github.com/AugurProject/augur/issues/7542


andrewdamelio (approved):
lgtm

Issue #7552, opened 2020-04-29
Shouldn't need to activate comments every time I visit a market


matt-bullock:
I'm having to activate comments even when I return to a market I activated them on

Chwy5:
Cant even load it to get into comments

pgebheim:
Comments are being removed from 2.0 -- will revisit all of these when we get back to adding them

Issue #7551, opened 2020-04-29
Shouldn't need to click 2 global chat buttons on mobile to activate it


matt-bullock:
Clicking the global chat button in the menu should trigger activation. No need to click the second one within the global chat

Chwy5:
same as ticket 7500

PR #7566, opened 2020-04-30
Added a few usability fixes for 3box chat and comments

bconfortin:
Should address those issues and one that wasn't mapped
#7551 #7552


bthaile (approved):
lgtm

Issue #7546, opened 2020-04-29
header alignment issue on mobile with lots of Created markets


andrewdamelio:
<img src="https://user-images.githubusercontent.com/1683736/80627322-ba26a180-8a1d-11ea-93b3-dda05a4ba91d.PNG" width="222" />

<img src="https://user-images.githubusercontent.com/1683736/80627344-c01c8280-8a1d-11ea-9e5b-3f75a1011534.PNG" width="222" />

Chwy5:
looks better

PR #7585, opened 2020-04-30
fix mobile header bug with long list

andrewdamelio:
header alignment issue on mobile with lots of Created markets #7546


bthaile (approved):
lgtm

PR #7589, opened 2020-05-01
[WIP] config.uniswap.exchangeRateBufferMultiplier + prettier

### LINKED ISSUE #7266: Make Uniswap slippage constant configurable with a config, use 0.5% for mainnet, 10% for testing
Currently this value is set to 10% for the auto-uniswaping of during contract contract calls to maintain a ETH balance in the signer.

@andrewdamelio -- what do you have it set to in the UI?

@epheph @MicahZoltu -- any thoughts here are on reasonable slippage from your experience?

robert-e-davidson3:
Could really easily add a few other values to the config too. Some constants from the gsn deps file:

```
const MIN_GAS_PRICE = new BigNumber(1e9); // Min: 1 Gwei
const DEFAULT_GAS_PRICE = new BigNumber(4e9); // Default: GasPrice: 4 Gwei

export const DESIRED_SIGNER_ETH_BALANCE = `0x${new BigNumber(
  0.04 * 10 ** 18
).toString(16)}`; // .04 ETH

const OVEREAD_RELAY_GAS = 500000;
const UNISWAP_MAX_GAS_COST = 150000;
const REFRESH_INTERVAL_MS = 15000; // 15 seconds
const GAS_PRICE_MULTIPLIER = 1.2;
const GAS_COST_MULTIPLIER = 1.2;
```

Can add the relay hub address too but it's a universal constant (create2) so that'd be superfluous.


nuevoalex (approved):
👍

PR #7590, opened 2020-05-01
deployment verify fixes & blocks-behind handling

nuevoalex:


bthaile (approved):
I'm not a fan of force reloading the UI, but it seems it would work to keep the user out of an error state.

PR #7596, opened 2020-05-01
Changed bg colors from fade to lighten

bconfortin:
Using the Less function fade for colors converts them to a RGBA color scheme instead of hex. That introduces opacity, which basically means that our buttons can be transparent.

That's harmful for us because, if we have a fixed button, like global chat, it'll be transparent on hover/active and whatnot, making the button see-through.

The function I used here is called "lighten" and it gives out the same color as fade, but without introducing opacity. The color is converted to a lighter version but in HEX value.

685291b7-5c15-4ae5-8b5a-45aff9ce8a9e


bthaile (approved):
lgtm

PR #7603, opened 2020-05-01
ui changes for uniswap config

### LINKED ISSUE #7266: Make Uniswap slippage constant configurable with a config, use 0.5% for mainnet, 10% for testing

robert-e-davidson3:


packages/augur-artifacts/src/configuration.ts
```diff
@@ -205,7 +205,8 @@ export const DEFAULT_SDK_CONFIGURATION: SDKConfiguration = {
     autoReport: false,
   },
   uniswap: {
-    exchangeRateBufferMultiplier: 1.05,
+    // mainnet will be <= 1.05 but for dev we can get away with a wide spread
+    exchangeRateBufferMultiplier: 1.75,
```
bthaile, line 209: can the value be moved to environment json config files since they will be different per network?

robert-e-davidson3: It is there. This is just the default. The mainnet.json file uses 1.05

bthaile (approved):
lgtm, I like @andrewdamelio to look at it also

andrewdamelio:
looks good!

packages/augur-artifacts/src/configuration.ts
MicahZoltu: It seems like the default should be mainnet numbers, and the config should override that.  That way if config fails to be applied for some reason, you don't fail over to "users get robbed blind".

Issue #7569, opened 2020-04-30
Mobile: Add logout button to account dropdown, remove from slideout menu + add dark background between overlay and page


matt-bullock:
[Design](https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=15402%3A1949)

Overlay background:

background: #0E0E0F;
opacity: 0.9;

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/4fe3ce63-cfce-4bc9-bc2f-1352b578832f)

Chwy5:
Not seeing the background darkened out

Chwy5:
fixed - closing

PR #7622, opened 2020-05-04
move login button

phoebemirman:
https://github.com/AugurProject/augur/issues/7569


bthaile (approved):
looks good. some single quotes `'` got replaced with `"` double quotes. need to check your formatter. approving

Issue #7574, opened 2020-04-30
Confirmed box not disappearing-bug


Chwy5:
After reporting on a market the box should processing then failed then confirmed ( and I do have a beel confirmation)....the confirmed box wont disappear....it is now 9 mins and I did confirm in the console that events are coming in and they are upating.

![4-30c.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/9db27000-d90d-4166-81ab-56981122c637)

clicking away and back it still is there....if i click on the x in the confrimed box it still shows in the notifiaction section with a report button

![4-30re.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/81079237-2103-42d6-842e-53c8ba685c03)

If i click report i can report again, even though i did

Chwy5:
seeing these

![4-30mm2.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/e20ee7e1-4c79-4c5e-b228-5dc5f5546525)

Chwy5:
![4-30dr.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/b46a49c9-2d9d-4c8f-8056-6e0d746553a8)

PR #7630, opened 2020-05-04
Update hash pending tx reporting disputing

bthaile:
https://github.com/AugurProject/augur/issues/7574

PR handles a few cases for pending queue.
1. There was a case where user sits on signing for awhile which processing button would show failed while transaction has been sent and is pending. Now, update blockNumber when transaction is signed. 
2. on occasion if user is fast to sign a transaction the blockNumber will be 0 in pending queue. ignore these transactions, they won't timeout. 

Also, set timeout blocks behind to 12


packages/augur-ui/src/modules/pending-queue/actions/pending-queue-management.ts
```diff
@@ -147,12 +231,13 @@ const processingPendingQueue = (
...
               pendingItem =>
                 (pendingItem.status === TXEventName.AwaitingSigning || pendingItem.status === TXEventName.Pending) &&
-                pendingItem.blockNumber <= thresholdBlockNumber
+                pendingItem.blockNumber && thresholdBlockNumber >= pendingItem.blockNumber
             )
         ),
       [] as PendingItem[]
     )
     .forEach(async queueItem => {
+      console.log(queueItem.blockNumber, 'threshold', thresholdBlockNumber);
```
andrewdamelio, line 240: did you want to keep this?

andrewdamelio (approved):
LGTM

PR #7632, opened 2020-05-04
add router queue. Fix market race condition

nuevoalex:


bthaile (approved):
lgtm, hopefully lock would solve the issue.

robert-e-davidson3 (approved):

Issue #7637, opened 2020-05-05
Toast message is too big-bug


Chwy5:
See screenshot (not supposed to be this long

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/26a465af-a018-4666-908c-672a2bd65feb)

Don't let the toast be wider than regular alerts

Chwy5:
not able to see any toast messages now when trading

matt-bullock:
this looks fixed. closing

Issue #7635, opened 2020-05-05
Mobile: space below 'betting UI coming soon' should be the same as space above


matt-bullock:
Currently it looks too big:

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/4b177be7-90ee-4aa7-bf3f-62b72bd98720)

matt-bullock:
looks good. closing

Issue #7634, opened 2020-05-05
Global chat button looks doubled up on mobile


matt-bullock:
Build:

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/05afa063-212e-4c9b-9f76-a80ddca194be)

Should look like:

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/6b102a3d-afa6-45b5-adf3-d81d53227dfe)

https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=12854%3A298319

matt-bullock:
fixed. closing

PR #7643, opened 2020-05-05
fix chat button borders and toast sizing and margins around betting ui msg

phoebemirman:
https://github.com/AugurProject/augur/issues/7634
https://github.com/AugurProject/augur/issues/7637
https://github.com/AugurProject/augur/issues/7635


bthaile (approved):
lgtm

PR #7645, opened 2020-05-05
process order events via bulk order events

bthaile:
created new bulk order event 
wired up UI to use it. 
use zerox status of `Synched` to call get user open orders if user is logged in.

### DIFF
--- a/packages/augur-sdk/src/state/db/ZeroXOrders.ts
+++ b/packages/augur-sdk/src/state/db/ZeroXOrders.ts
@@ -165,7 +165,7 @@ export class ZeroXOrders extends AbstractTable {
 
   async handleOrderEvent(orderEvents: OrderEvent[]): Promise<void> {
     if (orderEvents.length < 1) return;
-
+    const bulkOrderEvents = [];
     const filteredOrders = orderEvents.filter(this.validateOrder, this);
     let documents: StoredOrder[] = filteredOrders.map(this.processOrder, this);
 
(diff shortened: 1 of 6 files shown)


packages/augur-sdk/src/state/db/ZeroXOrders.ts
```diff
@@ -200,13 +201,16 @@ export class ZeroXOrders extends AbstractTable {
       const eventType = filledOrders[d.orderHash] ? OrderEventType.Fill : OrderEventType.Create;
       const event = {eventType, orderId: d.orderHash,...d};
       this.augur.events.emit('OrderEvent', event);
```
pgebheim, line 203: Any reason to keep OrderEvent?

packages/augur-sdk/src/state/db/ZeroXOrders.ts
```diff
@@ -216,7 +220,9 @@ export class ZeroXOrders extends AbstractTable {
       documents.forEach((doc) => delete this.pastOrders[doc.orderHash])
       await this.saveDocuments(documents);
       for (const d of documents) {
-        this.augur.events.emit('OrderEvent', {eventType: OrderEventType.Create, ...d});
+        const event = {eventType: OrderEventType.Create, ...d};
+        bulkOrderEvents.push(event);
+        this.augur.events.emit('OrderEvent', event);
```
pgebheim, line 225: Same here, do we even need this?

pgebheim (approved):
We should think about removing OrderEvent entirely unless it's being used

Issue #7652, opened 2020-05-06
Confirmed button and message Not going away again


Chwy5:
For the notification "Open Orders In a resolved Market" ...just stays there (this was fixed, regression

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/7aa3f9bd-1c5e-4151-a345-80d5a454b6b8)

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/f42602d2-5086-4968-9a2a-82e8f961c155)

Chwy5:
having same issue with the failed trade button too (little different than the message not going away)

Chwy5:
fixed - closing

PR #7662, opened 2020-05-06
missed adding bulk order event for fills after bulk zerox order import

bthaile:
Needed to use NewBlock logs to know Fill order is users and to to clear trading processing buttons

https://github.com/AugurProject/augur/issues/7652


andrewdamelio (approved):
lgtm

PR #7666, opened 2020-05-06
Outcomes table

phoebemirman:


packages/augur-ui/src/modules/market/components/market-view/betting-market-view.tsx
```diff
@@ -58,26 +85,49 @@ const BettingMarketView = ({ location }: BettingMarketViewProps) => {
           />
         </div>
       </div>
+      <OutcomeGroup
+        orderBook={orderBook}
```
JohnDanz: don't pass orderBook -- OutcomeGroup can just `useMarketsStore()` to get `orderbooks`, just use `marketId` so you can get the right book. also don't pass theme, same reason.

packages/augur-ui/src/modules/orders/actions/load-market-orderbook.ts
```diff
@@ -5,24 +5,7 @@ import { Action } from 'redux';
...
-    orderBook,
-  },
-});
-
-export const clearOrderBook = () => ({
-  type: CLEAR_ORDER_BOOK,
-});
+import { useMarketsStore, Markets } from 'modules/markets/store/markets';
 
 export const loadMarketOrderBook = (
```
JohnDanz, line 9: Ideally we wanna get rid of this entire action file -- maybe move this one function to a helper?

JohnDanz (changes requested):

JohnDanz (approved):

Issue #7539, opened 2020-04-29
Wrong network message issue


Chwy5:
When User is on the wrong Network - we trigger a message letting them know....the issue is the message only flashes for a split second, I couldnt even read it.

I am logged into MM and have MM on Mainnet network...then I try to pull up  V2, the "wrong Network" message flashes too quickly and disappears

bthaile:
Not able to repro. @Chwy5 have you seen this recently.

Chwy5:
yup...just tested and still did it...to recreate...dont have kovan v2 up....set MM to mainnet, then open a new window and bring up v2...it flashes super quick (cant read it) and it goes away

bthaile:
Steps to repro:
1. make sure you are logged out of augur and dev tools isn't not open.
2. set meta mask to mainnet
3. open new blank tab in chrome
4. go to v2.augur.net
5. you will see the network mismatch modal pop up then go away and market list just spins.

bthaile:
Issue is init-augur pops up the modal before the url is re-written to add `#!/markets` when that happens 
```    
window.onpopstate = () => {
      closeModal();
    };
```
is called which closes the modal. That code is used to close modals if user clicks back button.

need to either not change history object or need to re-write location before calling init-augur `connect`

Chwy5:
fixed - closing

PR #7677, opened 2020-05-07
convert modal view to pure component and use hooks to set up detectin…

bthaile:
…g browser back and forward button press

https://github.com/AugurProject/augur/issues/7539

to test:

have meta mask pointing to mainnet
load up UI, notice that the network mismatch modal doesn't flash

second test
after UI is loaded and network mismatch modal is up
remove `/#!/markets` from URL and hit <enter> to direct browser to new location
notice that network mismatch modal stays shown


andrewdamelio (approved):
nice fix!

PR #7693, opened 2020-05-07
check to see if account has token approval before trying to swap

andrewdamelio:
- added `checkSetApprovalAmount` to check to make sure users have token allowance before making a swap


bthaile:
Will the check prompt the user to sign a transaction if they need to have approval?
If so is the user warned?

bthaile (approved):
lgtm, one comment but wouldn't hold up merging

PR #7695, opened 2020-05-07
have one field for all the market properties that need to be full tex…

### LINKED ISSUE #7593: Market Search not working properly
When I enter NFL or Draft, the market I was looking for does not appear

bthaile:
…t searchable

1. noticed that flexsearch was searched even thought there was not `search` or `categories` passed into getMarkets. added a conditional
2. removed `*` appeneded to keyword search in UI.
3. re-organized fields in flexsearch, only a few fields needed to be indexed individually. The bulk of the market properties could be added to one field. 

https://github.com/AugurProject/augur/issues/7593


pgebheim (approved):
:+1:

Issue #7675, opened 2020-05-07
Trade history getting cut off at bottom for desktop


matt-bullock:
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/65dc7917-d08b-48a4-add1-6dbbad8be276)

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/7a6c80b6-3a18-4f66-8874-67d53128ad9d)

Chwy5:
the trade history scrolls...there were trades under that

just tested

closing

matt-bullock:
This has caused a new problem - it looks like it's getting cropped off at the bottom of the page. The bottom of trade history should line up with the bottom of order ticket and orders

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/234a5368-e7ea-402c-b6e3-e3c565cd4024)

andrewdamelio:
@matt-bullock could you please provide browser type, market id and screen resolution for the above screenshot

matt-bullock:
@andrewdamelio https://v2.augur.net/#!/market?id=0x057BDc6949450a897769F8c1597c84cf28Ff6CD1

Chrome on mac. 1034px browser height

matt-bullock:
@andrewdamelio still looks off on chrome

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/9ec2a225-9311-407d-8933-449377fce2b8)

matt-bullock:
Still seeing it

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/4d56b69c-41dc-4b51-849d-6058b544dc96)

andrewdamelio:
@matt-bullock i made the global chat button have some opacity also as we discussed 
<img width="439" src="https://user-images.githubusercontent.com/1683736/84151342-38209400-aa31-11ea-8a96-8aec72696243.png">

matt-bullock:
Looks much better. closing

PR #7710, opened 2020-05-08
fix trade history

andrewdamelio:
Trade history getting cut off at bottom for desktop #7675

### DIFF
--- a/packages/augur-ui/src/modules/market/components/market-trade-history/market-trade-history.styles.less
+++ b/packages/augur-ui/src/modules/market/components/market-trade-history/market-trade-history.styles.less
@@ -13,7 +13,7 @@
   > div {
     display: block;
     overflow: overlay;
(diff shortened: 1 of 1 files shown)


packages/augur-ui/src/modules/market/components/market-trade-history/market-trade-history.styles.less
```diff
@@ -11,11 +11,9 @@
   }
 
   > div {
-    display: flex;
-    flex-direction: column;
-    flex-grow: 1;
+    display: block;
     overflow: overlay;
-    max-height: 50vh;
+    max-height: calc(50vh + 53px);
```
bthaile: can you use one of these?
@size-50: 3.125rem; // 50px
@size-52: 3.25rem; // 52px

bthaile (approved):
lgtm

PR #7721, opened 2020-05-08
Revamped the Hero to work like the new landing page

bconfortin:
The hero now accepts a background image and 2 images (primary and secondary), just like the new landing page.


bthaile (approved):
nice, lgtm

Issue #7706, opened 2020-05-08
Test Trading broke on firefox


Chwy5:
Using FF on Windows:
see screenshot

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/ef985b87-3e11-4968-847d-f74ba431e7c5)

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/155896c3-2eef-4e26-a92c-913a2c983122)

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/bf9e97f7-e43b-45ee-b6ba-c01d0b24e441)

matt-bullock:
Looks ok on firefox on mac. @Chwy5 can you test on windows.

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/66addeaf-9dba-47ec-a273-180583a7c8a0)

The styling looks off on the error message (X looks tiny and copy doesn't need to be all caps). Should follow this ![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/1030ce26-a3e6-45b3-b5ae-e9d807ab69cf)

https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=8218%3A266108

Chwy5:
fixed - closing

Chwy5:
Mobile "Open" issue...Works for me on the emulator using FF but @matt just downloaded FF on his phone and still sees the issue

Chwy5:
looks fine closing

PR #7732, opened 2020-05-11
Fixed tutorial popovers in Firefox

bconfortin:
#7706


bthaile:
merged this PR into 7707, to test both at same time

PR #7735, opened 2020-05-11
Add market not found modal

bthaile:
The PR handles 2 scenarios. 
1. malformed ethereum address. it doesn't parse using `getAddress` from 'ethers/utils/address'; This method call throws exception
2. valid ethereum address, just isn't a marketId. 

https://github.com/AugurProject/augur/issues/6691
scenario 2, valid ethereum address isn't a market id. 👇  animated gif


andrewdamelio (approved):
Lgtm

PR #7743, opened 2020-05-12
Separating 3box in its own component and using lazy load

bconfortin:
Refactored the way 3box works so we can have better maintenance and not have it injecting iframes in Augur's body when 3box is not being used.


bthaile (approved):
lgtm

Issue #7707, opened 2020-05-08
Firefox market display bug on mobile


Chwy5:
On Firefox on mobile go to a market that has many resolution details. Click on the arrow to expand the resolution details

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/bfd5ffd9-d200-4cb7-8f4e-af2cf4d9cba0)

matt-bullock:
still looks broken to me

Chwy5:
now fixed- closing

bconfortin:
For some reason this was still broken on Firefox for iPhone but not Firefox for Android. Turns out it was just cached, even after I cleared it and reinstalled Firefox on the iPhone. Seems fixed now, will let you guys test.

matt-bullock:
Unable to load markets @chwy can you test

Chwy5:
i dont have an iphone works on the emulator

PR #7751, opened 2020-05-13
Firefox market page bugs

bconfortin:
Adjusted the DismissableNotice just like the Figma Design (#7706)
Removed static height from the market header in order to fix Firefox layout (#7707)


bthaile (approved):
lgtm

Issue #7739, opened 2020-05-12
Sportsbook Accounts Touchups


JohnDanz:
- [x] The "Your Funds" section is too tall, it looks to be about ~40 px to tall.

![Screen Shot 2020-05-12 at 11.43.02 AM.png](https://images.zenhubusercontent.com/57155267bf93871ccd99a48b/0d60e7b2-71ad-46a9-8f30-3084ce9ebcee)

- [x] I would suggest we use grid gap instead of margins on the various pieces to create the space between them.

- [x] The pill selectors are missing the dark borders for the selected time in Augur Status section and the Overview section

![Screen Shot 2020-05-12 at 11.45.46 AM.png](https://images.zenhubusercontent.com/57155267bf93871ccd99a48b/156f4418-189e-4bfb-972b-e3875029ec7a)

![Screen Shot 2020-05-12 at 11.46.47 AM.png](https://images.zenhubusercontent.com/57155267bf93871ccd99a48b/dc69f10a-4f64-4626-8426-5098b535d059)

- [x] The Overview chart shouldn't be leaving this whitespace at the bottom, we want the grey background/chart to go all the way to the bottom of this section. (This section is most likely too tall)

![Screen Shot 2020-05-12 at 11.47.25 AM.png](https://images.zenhubusercontent.com/57155267bf93871ccd99a48b/8e6772fc-ff20-4547-b218-f4e43141714e)

- [x] We are missing a count for My Active Markets section and we need to give more space between that count and the search bar on that section as well as any other section with a search bar (see mocks):

![Screen Shot 2020-05-12 at 11.48.35 AM.png](https://images.zenhubusercontent.com/57155267bf93871ccd99a48b/64270e82-6146-44ce-b392-5ec4d4cdea2f)

- [x] The Augur Status section also appears to be too tall, much like overview, we shouldn't have so much dead whitespace below the bottom stats.

![Screen Shot 2020-05-12 at 11.49.32 AM.png](https://images.zenhubusercontent.com/57155267bf93871ccd99a48b/13fae847-6f96-4341-84cf-80f7cae1140f)

- [x] We are missing the bottom bar on My Created Markets with a link to Create Market (this link should take us to Create Market section but also switch the theme to Trading  < if you are unsure on this please let me know and i can give you an example of how to do it>)

[...]

PR #7753, opened 2020-05-13
Fix layout and search bar Sportsbook accounts touchups

rubyweb3:
https://github.com/AugurProject/augur/issues/7739

Screen Shot 2020-05-13 at 7 17 48 PM

- The "Your Funds" section is too tall, it looks to be about ~40 px to tall.
- use grid gap instead of margins on the various pieces to create the space between them.
- The Overview chart shouldn't be leaving this whitespace at the bottom, we want the grey background/chart to go all the way to the bottom of this section. (This section is most likely too tall)
- We are missing a count for My Active Markets section and we need to give more space between that count and the search bar on that section as well as any other section with a search bar (see mocks):
Screen Shot 2020-05-13 at 6 15 43 PM
- The Augur Status section also appears to be too tall, much like overview, we shouldn't have so much dead whitespace below the bottom stats.


JohnDanz:
Screen Shot 2020-05-13 at 11 18 05 AM

we need to have this module grow a bit more and not have a scrollbar.

Title of the page also looks a bit too spaced out compared to the mocks.

JohnDanz (changes requested):
couple notes at the bottom, also still missing some things from the original issue but I assume that is on purpose for now

Edit:

the updated null states look good, need to add the counters to the right of the titles for the market panes however

rubyweb3:
> couple notes at the bottom, also still missing some things from the original issue but I assume that is on purpose for now
> 
> Edit:
> 
> the updated null states look good, need to add the counters to the right of the titles for the market panes however

Updated. Thanks

JohnDanz (approved):

PR #7759, opened 2020-05-13
Initial cut of zerox to WW

justinbarry:


packages/augur-ui/src/services/browser-mesh.ts
```diff
@@ -105,42 +106,33 @@ export async function createBrowserMesh(
...
-    const newMeshTimerLabel = 'new Mesh() startup duration: ';
-
-    zeroX.client.events.emit(SubscriptionEventName.ZeroXStatusStarting, {});
-    console.time(zeroXTimerLabel);
     await loadMeshStreamingWithURLAsync('zerox.wasm');
 
     const meshConfig = createBrowserMeshConfig(
       config.ethereum.http,
-      web3Provider,
+      undefined,
```
pgebheim, line 113: Ah yes, this is the killer as to why this needs to be in the main thread. In order for this to be feasible we would need to comlink proxy all the eth calls back for this too.

pgebheim (changes requested):
I'm not sure about this yet -- it needs:
   - Turning off the worker easily and just starting mesh from the main thread
   - We really need to proxy the eth connections back to the main thread even for a short term workaround.

Other than that the strategy looks alright -- wondering if we should split this into a couple of commits to get the CI changes etc all in and then think about whether or not to move forward with the worker.

justinbarry:
Moved CI stuff to #7773

packages/augur-ui/src/services/browser-mesh.ts
justinbarry: I will see if I can do that with comlink.

pgebheim (approved):

packages/augur-ui/src/services/MeshTransferHandler.ts
```diff
@@ -0,0 +1,32 @@
+import * as Comlink from 'comlink';
+import { BigNumber } from '@0x/utils';
+import { mapValues } from 'lodash';
+
+const serialize = item => {
```
robert-e-davidson3, line 5: Should be able to do this as part of JSON.stringify.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

justinbarry: I can't if I want to add extra information for deserialization.

justinbarry: If we decide this those also support WS here then we should update zero validation login to handle "string-ed" big numbers.

PR #7760, opened 2020-05-13
change tooltip in account dropdown, add eth reserve replenish notice…

bthaile:
… in order form

Screenshots are 
1) before replenishment
2) notice to tell user they will replenish eth reserve
3) after replenishment

Screen Shot 2020-05-13 at 4 18 07 PM

Screen Shot 2020-05-13 at 4 19 09 PM

ETH reserve was added to
Screen Shot 2020-05-13 at 4 20 30 PM

Changed tooltip in account dropdown
image

Also fixed issue where numFills was used for number of transactions to take all orders. transaction can take 3 orders in one transaction.


andrewdamelio (approved):
👍

Issue #7770, opened 2020-05-14
tennis template rule fix 


Chwy5:
For the Singles And doubles template for multiple choice templates (not yes/no) the rule should be:

If a Tournament or Event is cancelled or postponed and will not be completed before the market's Event Expiration time, the market should resolve as 'No winner/Event Cancelled'.

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/b8616fbb-6921-40b2-9970-5ee84d9fc640)

Chwy5:
The wrong rule got changed...now all the tennis multiple choice ones have an incorrect one....ping me and we can go through it together

Chwy5:
fixed - closing

Issue #7727, opened 2020-05-11
Open Order Preset Cancellation limit


Chwy5:
When a user creates an open order default expiration should be preset to when the Event Resolution for that market begins

If a market has reached the In Reporting State any New orders placed should have a preset order cancellation limit at 1 week

Market creation and order form
 - add advanced to pre-liquidity
 - default order expiration to market event expiation 
 - use local time with local timezone (EDT), ...

Chwy5:
everything working correctly, but still not seeing local time zone in order seeing utc version instead

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/3c82f3ff-bbff-436a-aa7b-47c3d2b3df36)

bthaile:
I forgot to use the timezone abbr instead of utc format. Push this ticket back to ui dev incoming and I’ll make that change

Chwy5:
Seeing some funkiness with this....

If I fill in the order ticket and add an open order to the orderbook and then do nothing for a while.  IF i come back later and enter the order book again (it still shows the time under order expiration, when I did my first trade

If I clear it out and reenter the order ticket it didnt show me any date...I could still hit the place buy order button and my order went through with the correct expiration time

bthaile:
If the market is still open then the expiration will stay the same as market event expiration. 

if market is in-reporting the order form will show order exp of 1 week. If you leave and come back the time will update to be 1 week in future.

Chwy5:
In market creation...the pre-liq order doesnt match the event expiration time

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/bf8a12b2-8297-43f5-aabe-4dba7f7fe38f)

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/93ee4590-e1e1-4aba-8b1a-85ba533160af)

Chwy5:
Still off....

If I create a market for later today(for expiration) the pre-liq order shows 7 days.

I also tried 3 days in the future and the time is off (see screenshots

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/1c9d65e1-b471-4484-bf5b-ab4b9baf2a68)

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/7e07a17d-4c2b-4bff-aade-f1d16924aec4)

Chwy5:
fixed -closing

PR #7772, opened 2020-05-14
Negative frozen funds timezone def order exp

bthaile:
https://github.com/AugurProject/augur/issues/7770
https://github.com/AugurProject/augur/issues/7727

also, don't let frozen funds be negative. It's possible but only show 0.

show user's timezone abbr, not UTC offset.
image

Template tweaks to be consistent and update res rule for tennis templates.


andrewdamelio (approved):
lgtm

PR #7783, opened 2020-05-15
changed wrong res rules, added in missing outcomes for two templates

bthaile:
https://github.com/AugurProject/augur/issues/7770


andrewdamelio (approved):
lgtm

PR #7801, opened 2020-05-18
need to set expirationDate and need to not update based on external c…

bthaile:
…hanges

https://github.com/AugurProject/augur/issues/7727


bconfortin (approved):
Lgtm

PR #7804, opened 2020-05-18
added test to test partial close of user position

bthaile:
added test for user close partial position.

### DIFF
--- a/packages/augur-test/src/tests/state/getter/users/get-profit-loss-and-summary.test.ts
+++ b/packages/augur-test/src/tests/state/getter/users/get-profit-loss-and-summary.test.ts
@@ -821,44 +821,46 @@ describe('State API :: Users :: PL :: ', () => {
     });
   });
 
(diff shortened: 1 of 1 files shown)


packages/augur-test/src/tests/state/getter/users/get-profit-loss-and-summary.test.ts
```diff
@@ -820,6 +820,47 @@ describe('State API :: Users :: PL :: ', () => {
       },
     });
   });
+
+  test('Scenario 7: partial closing initial position', async () => {
```
robert-e-davidson3: The "Scenario #" convention is from a specific google doc. I'd put this test in its own `describe` block.

bthaile: ok

robert-e-davidson3 (approved):

Issue #7765, opened 2020-05-14
firefox mobile android: all app screens getting cut off before the bottom of browser


matt-bullock:
See bottom of screen. This shouldn't be happening. Can you also test for this on iOS

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/4370e23f-16a8-431c-a1b0-16bbd701256b)

bconfortin:
That is going to be a tricky one. I'll have to investigate further, but It doesn't happen on the iPhone, which makes me think it's the extra padding we leave at the bottom because of the address bars on the top.

bconfortin:
Leaving this in "UI Dev Outgoing" for now. My Android phone should arrive in the mail any minute now, will test the whole app on iPhone and Android to see if I broke something, then I'll move this to product/design review.

matt-bullock:
@bconfortin looks like you've fixed part of it but it's caused another issue - you can see the help button gets caught on the main screen and it still shows the gap until I scroll. And the help button isn't showing on the menu where it should be

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/05ef3294-7f25-4705-88da-9a33b921956d)

matt-bullock:
Still seeing this problem. I'm going to close and create a new ticket as the original issue was addressed

PR #7806, opened 2020-05-19
Removed 90px of difference since it's a fixed top bar

bconfortin:
Mobile: Removed 90px of difference since it's a fixed top bar
#7765


justinbarry (approved):
LGTM

PR #7807, opened 2020-05-19
Switched to percentage because of double scrolling bug

bconfortin:
#7765
I fixed the screen being cut, but there is something weird when scrolling using `100vh`.

Double scrolling bug:
It seems that using `100vh` instead of `100%` makes the height calculation slower than percentage, making the user have to scroll all the way to the bottom or top twice on a mobile device. Like... pull once and pull again, otherwise the top or bottom of the screen will be just slightly cut.

Animated GIF-downsized_large


bthaile (approved):
Stamped

PR #7816, opened 2020-05-19
Gh 7259 remove warp sync processingreporting on mobile

### LINKED ISSUE #7259: Remove warp sync processing/reporting on mobile
- Keep loading from warp checkpoint
- Do not generate checkpoints on mobile

This should be implemented by turning off a config in `init-augur` and the rest of the changes should check that config.

justinbarry:
Closes #7259.

Note: This will cause the client to do a complete refresh every 7 days from warp sync. Unsure how desirable this is


bthaile (approved):
lgtm. we need to see if it works

Issue #7822, opened 2020-05-20
stock/ETF scalar market templates fix


Chwy5:
Preset the denomination to whatever the currency is selected from the exchange pick- where it says (trade in USD)

Chwy5:
fixed -closing

Issue #7819, opened 2020-05-20
Sports template bug- not allowing 0 to be valid input


Chwy5:
This appears to be effecting all sports

User needs to be able to put in 0...bc spreads and o/u can be 0.5 (the .5 is part of the template question not in the input)...see screenshots below

For templates vs the spread and Over/Under not allowed to input a 0...triggers an error message

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/a509b6b5-44c4-4422-9c95-52393841f3ce)

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/8aafd01d-85f7-4e56-8751-818a365ec9e3)

Chwy5:
fixed -closing

PR #7824, opened 2020-05-20
Fix template pre filled values

bthaile:
1. fixed typo on market creation review page
2. fixed allowing 0 as a whole number value
3. added ability to set denomination for financial scalar template based on exchange picked

https://github.com/AugurProject/augur/issues/7819
https://github.com/AugurProject/augur/issues/7822


andrewdamelio (approved):
lgtm!

Issue #7726, opened 2020-05-11
Future Template Ideas


Chwy5:
** MMA (this is the sport to list)
    UFC (is the organization like Basketball and NBA/WNBA/NCAA)- Organization should be shown before each market question
    Bellator

Multiple Choice Markets
#1)
[Fighter A] vs [Fighter B] `(Moneyline)`: Who will win?
Estimated Event Start time

Outcome 1) Fighter A
Outcome 2) Fighter B
Outcome 3) Draw

#2)
[Fighter A] vs [Fighter B] O/U: [#}.5 Rounds
Estimated Event Start time

Outcome 1) Over #.5 Rounds
Outcome 2) Under #.5 Rounds

#3)
[Fighter A] vs [Fighter B]: Method of victory?
Estimated Event Start time

Outcome 1) [Fighter A] by KO/TKO
Outcome 2) [Fighter A] by Submission
Outcome 3) [Fighter A] by Points
Outcome 4) [Fighter B] by KO/TKO
Outcome 5) [Fighter B] by Submission
Outcome 6) [Fighter B] by Points
Outcome 7) Draw

#4) 
[Fighter A] vs [Fighter B]: How will the fight end?
Estimated Event Start time

Outcome 1) KO/TKO
Outcome 2) Submission
Outcome 3) Points

#5)
[Fighter A] vs [Fighter B]: What round will the fight end?
Estimated Event Start time

Outcome 1)  Round 1
Outcome 2)  Round 2
Outcome 3)  Round 3

**Boxing

Multiple Choice Markets
#1)
[Fighter A] vs [Fighter B] `(Moneyline)`: Who will win?
Estimated Event Start time

Outcome 1) Fighter A
Outcome 2) Fighter B
Outcome 3) Draw

#2)
[Fighter A] vs [Fighter B] O/U: [#}.5 Rounds
Estimated Event Start time

Outcome 1) Over #.5 Rounds
Outcome 2) Under #.5 Rounds

#3) 
[Fighter A] vs [Fighter B]: Method of victory?
Estimated Event Start time

Outcome 1) [Fighter A] by KO
Outcome 2) [Fighter A] by TKO
Outcome 3) [Fighter A] by Points
Outcome 4) [Fighter B] by KO
Outcome 5) [Fighter B] by TKO
Outcome 6) [Fighter B] by Points
Outcome 7) Draw

#4) 
[Fighter A] vs [Fighter B]: How will the fight end?
Estimated Event Start time

[...]

PR #7834, opened 2020-05-21
Template group hash structure

bthaile:
https://github.com/AugurProject/augur/issues/7726

template changes, added MMA, boxing and car racing templates and marketed for sportsbook

added grouping for NFL, basketball and baseball. 

image


pgebheim (approved):
:+1:

PR #7836, opened 2020-05-21
Final context rollups

JohnDanz:
Finished context rollup with the except of initialize3Box which we might drop from state, I gotta talk to bruno


packages/augur-ui/src/modules/create-market/components/form.tsx
```diff
@@ -179,450 +187,493 @@ interface Validations {
...
+  } = AppStatus.get();
+  const { addUpdateDraft, updateNewMarket } = AppStatus.actions;
+
+  if (newMarket.description === EMPTY_STATE.description) {
+    onError('description', draftError);
+    return;
   }
 
-  componentWillUnmount() {
-    if (!this.state.blockShown) this.unblock();
```
phoebemirman, line 204: missing this call

phoebemirman (changes requested):
seeing it crash on 2nd page of market creation form, also crashes on initial load because is logged is declared twice in log-handlers

phoebemirman (approved):

PR #7857, opened 2020-05-26
set time and unit based on market endTime. there was a bug if market …

bthaile:
… expires within a day, order expiration would say 7 days. Needed to set order expiration based on market endTime using time units, days, hours, minutes. 

Screen Shot 2020-05-26 at 11 25 28 AM

Screen Shot 2020-05-26 at 11 27 43 AM


bconfortin (approved):
Seems to be working as intended. Lgtm :D

Issue #7577, opened 2020-04-30
Old bell alerts showing up after being cleared.


Chwy5:
I had notification of Open orders in a resolved market....I went to modal and hit cancel all....they are showing up in the bell message as order placed....

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/2690ac1c-40d5-4eb5-b3ce-eca3663172ec)

Chwy5:
Had this occur again....

I had the cancel order notification- when I open it the modal it had 1 order shown
I hit cancel all button (which failed), but got the bell alert of 2 "orders placed" 
-(first issue, we need to check, we are showing all the open orders)
-(2nd thing) Why are they showing up as a bell alert that says order placed? (We also dont do a bell message for cancelled orders)
I went and found the market and confirmed that the orders did not get placed
Not seeing any errors in the console

![5-1orderplaced-cancelled.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/b38d5ddc-4976-4453-9a5c-5055ca3e0d5e)

![5-1orderplaced-cancelled2.jpg](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/49a41fad-fbdb-4a19-9e3f-d0daf599d782)

bthaile:
I changed the title of this one. I think what is going on is old cleared bell alerts are showing back up. moving this to UI dev incoming.

bconfortin:
This is actually happening whenever you have orders placed and refresh the page. Even if you cleared it beforehand, the bell alerts will reappear.

Chwy5:
looks fixed

PR #7865, opened 2020-05-26
Moved logs the order created bell alert

bconfortin:
#7577


bthaile:
created 4 orders as fast as the UI would allow. seems to work fine.
image



made a few changes. I think there would be a bug if all open orders didn't get a bell alert.

bthaile (approved):
lgtm

PR #7881, opened 2020-05-28
update Uniswap contracts

nuevoalex:


bthaile (approved):
seems to be mostly name changes. lgtm

PR #7886, opened 2020-05-28
Fix transfer modal max

bthaile:
few changes in this mix-bag PR.
1. user full precision of eth dai and rep balances. don't use format functions
2. change ETH reserve to Fee reserve
3. fixed transfer.
 a. user has to have ETH in signing wallet to transfer tokens from signing wallet to trading account
 b. user can use ETH in signing wallet or DAI in trading account to transfer out of trading account

4. small template tweaks
5. fixed id of tooltip, total funds was getting order form tooltip


bconfortin (approved):
lgtm

Issue #7840, opened 2020-05-22
Firefox mobile: header and close button not showing on categories and filters menu


matt-bullock:
![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/9481ddd8-d59c-4456-ad18-d2ed08d2e262)

Should see this

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/6d4e2626-7e06-41ab-ae40-69165b674e0b)

bconfortin:
@matt-bullock Not able to reproduce. Used chrome and firefox in both Android and iOS. Also tried a small screen on chrome and firefox in my Macbook.

Can you please test that again?

Android Firefox
![Screenshot_20200527-112359_Firefox Beta.jpg](https://images.zenhubusercontent.com/5d7fcf516b3dcf00019b60cd/3c8198f3-0387-48af-a23e-0b368bcf48dd)

Android Chrome
![Screenshot_20200527-112446_Chrome.jpg](https://images.zenhubusercontent.com/5d7fcf516b3dcf00019b60cd/95b901fb-5341-466c-930b-cf59976f449e)

iOS Firefox
![IMG_5052.PNG](https://images.zenhubusercontent.com/5d7fcf516b3dcf00019b60cd/c81dba3e-91da-438f-a604-dab8537682b2)

iOS Chrome
![IMG_5051.PNG](https://images.zenhubusercontent.com/5d7fcf516b3dcf00019b60cd/e9d12a46-4cc1-4a29-9cff-96ac9eead8be)

bconfortin:
@matt-bullock Should be working now. I had to remove the fixed header from the mobile menu because it is a bug specific to the browser and how it handles transition (sliding menu) and position (fixed header). The Firefox Beta app on Android doesn't have this bug so maybe we can make the header fixed again in the future.

matt-bullock:
I think that's fine to not have it fixed. I didn't notice the difference. Nice work getting that fixed

PR #7891, opened 2020-05-29
Firefox will bug when there's position fixed and transition happening

bconfortin:
#7840 

This was a tricky one. Explanation below:

We have a transition when the Filter and Categories' menu is being open on mobile and we also have that inner header ("Categories & Filters") set to `position: fixed`, which causes the animation to lose itself and kinda crash. Sometimes it works a bit choppy,  sometimes it simply vanishes, which is when the user has to use the "back" button on Android to get unstuck or refresh the page.

I've tested that on desktop, in an old mozilla browser (version 68, which seems to be the same-ish version as mobile's version) and the bug is there as well.

The fix I've found is to use a `position: initial` header, which makes the header not follow the user's scroll but will provide a much better user experience, given that the user won't get stuck every time they open the Filter and Categories' menu.

### **Important to note!**
This fix can be only temporary. I've downloaded Firefox Beta on my phone and they seem to have fixed that animation bug. I imagine we can reverse this changes in a few months, when almost everyone has the newer version of the Browser.

### DIFF
--- a/packages/augur-ui/src/modules/app/components/app.styles.less
+++ b/packages/augur-ui/src/modules/app/components/app.styles.less
@@ -210,16 +210,20 @@ h6 {
   background: @color-module-background;
   border-right: @size-1 solid @color-dark-grey;
 
-  @media @breakpoint-mobile {
-    padding-top: 0;
-  }
-
   > div:first-of-type {
-    left: 0px;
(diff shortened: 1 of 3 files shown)


bthaile (approved):
Lgtm

PR #7898, opened 2020-05-29
Retry loading of wasm 5 times before giving up

### LINKED ISSUE #7749: Handle Error during mesh startup and try again
Since we've switched to instantiateStreaming, any errors during wasm parsing (like the safari stack depth issue) are no longer retried.

We need to extend some error handling around the loading of the mesh and retry if it fails.

justinbarry:
Closes #7749


pgebheim (approved):

pgebheim:
Looks good

PR #7907, opened 2020-05-30
Add to withdraw all funds

bthaile:
Added transfer DAI from withdraw all funds if user doesn't have ETH in signing wallet. 
Converted transfer-form to pure component.
Added better gas estimates and conversion to dai for display
Added processing button to transfer form

There are still some edge cases to figure out when signing wallet doesn't have enough ETH to transfer user's DAI from trading account.


packages/augur-ui/src/modules/auth/components/connect-dropdown/connect-dropdown.tsx
```diff
@@ -176,7 +176,7 @@ const ConnectDropdown = (props: ConnectDropdownProps) => {
       logo: EthIcon,
       value: ethReserveInDai,
       subValue: reserveEthAmount.formattedValue,
-      disabled: GsnEnabled ? (balances.signerBalances?.eth || 0) === 0 : false,
+      disabled: GsnEnabled ? balances.signerBalances.eth !== "0" : false,
```
bconfortin, line 179: This might cause a TypeError: Cannot read property 'eth' of undefined. I suggest re-adding the "?." / optional chaining.

packages/augur-ui/src/modules/modal/cash-out-form.tsx
```diff
@@ -28,34 +37,65 @@ interface CashOutFormProps {
   reserveInDaiFormatted: FormattedNumber;
   totalDaiFormatted: FormattedNumber;
   tradingAccountEthFormatted: FormattedNumber;
+  totalDai: string;
 }
 
-export const CashOutForm = ( props: CashOutFormProps) => {
+export const CashOutForm = (props: CashOutFormProps) => {
```
bconfortin: Since you touched that file, I would suggest removing the destructuring from the first line of this function and just adding to the parameters, like so:

```jsx
export const CashOutForm = ({
    closeAction,
    withdrawAllFunds,
    withdrawAllFundsEstimateGas,
    account,
    gasPrice,
    totalOpenOrderFundsFormatted,
    availableFundsFormatted,
    reserveInDaiFormatted,
    totalDaiFormatted,
    tradingAccountEthFormatted,
  }: CashOutFormProps) => {
```

packages/augur-ui/src/modules/auth/components/connect-dropdown/connect-dropdown.tsx
bthaile: it won't the reducer is defaulted to have a value. the `?` isn't needed.

packages/augur-ui/src/modules/modal/cash-out-form.tsx
bthaile: keeping the format. following how betting ui is doing it. it'll make merging easier for them

bthaile: good idea. I'll do that

bconfortin (approved):
Lgtm

Issue #7930, opened 2020-06-01
toast message bug with scalars


Chwy5:
Market range was 120-200

I entered an order to buy 1 share at 150...shows as 270 in the toast

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/ec10c850-23dd-4b1d-b59b-fff5b4e22673)

Chwy5:
fixed -closing

PR #7939, opened 2020-06-02
Fix scalar short price alert text

bthaile:
Need to convert price differently based on open order or trade.

image

https://github.com/AugurProject/augur/issues/7930


packages/augur-ui/src/modules/common/labels.tsx
```diff
@@ -1089,7 +1089,7 @@ export const PendingLabel = (props: PendingLabelProps) => (
         </ReactTooltip>
       </>
     )}
-    {props.status && props.status === TXEventName.Failure && (
+    {status && status === TXEventName.Failure && (
```
bconfortin: You don't need this since you've already destructured the props above:
`status && `

bconfortin: You actually didn't need it before either

bthaile: status can be null or undefined

bthaile: but I see your point. I'll clean it up more

bthaile: pushed, updated PR

bconfortin (approved):
Lgtm

Issue #7949, opened 2020-06-02
claiming REP initial report which finalized the market and frozen funds isn't correct.


Chwy5:
I had to claim rep forinitial report for multiple markets (1 which was in awaiting finalization), after claiming my FF are off by $10, doesnt appear to have given the bond back....

I think the market is finalized but FF isnt updating

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/7cd15847-9dd2-4d13-b9f2-b00c016d8032)

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/7e42ff68-99bf-4fb6-b187-e28dea7e4165)

Chwy5:
fixed closing

PR #7960, opened 2020-06-03
fix scenario where market owner doesn't have position and market is f…

bthaile:
…finalized and market creators frozen funds doesn't update

https://github.com/AugurProject/augur/issues/7949


bconfortin (approved):
Lgtm

PR #7965, opened 2020-06-03
fix typo and move up conditional to save on execution

bthaile:
fixed typo, `lenth` should have been `length` and move up conditional to save on executions.


bconfortin (approved):
Lgtm

PR #7969, opened 2020-06-04
only show automatic top off checkbox if user has min eth for fee rese…

bthaile:
…rve in signing wallet. if they have more then don't show

User has lots of eth in signing wallet, so don't show automatic top off checkbox
image

user has less than 20% over min eth in signing wallet
image


andrewdamelio (approved):
LGTM

Issue #7970, opened 2020-06-04
Update Twitter template


Chwy5:
Twitter: Will @[Social Media Handle] have [Number] [Amount Unit] or more twitter followers on [Day of Year], according to www.trackalytics.com?

change trackalytics to www.socialblade.com

We need to make event expiration date and time 24 hrs after the date selected in the market question

Change the rules to say:

This market is to resolve based off of the number shown for the specified date on www.Socialblade.com.

To find the results: In the header of the page. Select "Twitter" from the drop down menu. The enter the twitter handle in the search bar.  Scroll down the page until you reach the section that says "Twitter Stas Summary/ User Statistics for (named person).  Find the date named in the market question and get the larger number under the "Followers" section (number will not have a +/-).  Settlement should be based off of this number.

Chwy5:
added twitter + instagram

PR #7972, opened 2020-06-05
7970 - Changing suggested market resolution text to socialblade.com

jfarid27:
# Description

Changes suggested twitter market resolution text to socialblade.com.

# Issue
#7970

### DIFF
--- a/packages/augur-tools/src/templates-source.ts
+++ b/packages/augur-tools/src/templates-source.ts
@@ -8591,8 +8591,9 @@ export const TEMPLATES = {
         templates: [
           {
             marketType: YES_NO,
-            question: `Twitter: Will @[0] have [1] [2] or more twitter followers on [3], according to www.trackalytics.com?`,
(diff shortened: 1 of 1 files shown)


jfarid27:
@bthaile Do I have to check in any build files? I'm not exactly sure if this is the only required step here.

bthaile:
Thanks for the suggestion, we are already in the process of changing the resolution source for the twitter template and looking into adding more templates. 

Also for future suggestions, the long text would be a resolution rule not in the market question. 

```
'To find the results: In the header of the page. Select "Twitter" from the drop down menu. The enter the twitter handle in the search bar. Scroll down the page until you reach the section that says "Twitter Stas Summary/ User Statistics for (named person). Find the date named in the market question and get the larger number under the "Followers" section (number will not have a +/-). Settlement should be based off of this number.',
```
would be another `text` in the REQUIRED `resolutionRules` closing

Yes there is a flash script to generate templates
`yarn flash generate-templates`

PR #7988, opened 2020-06-09
Setting the environment config in webpack as an envvar instead of usi…

robert-e-davidson3:
…ng require.context


robert-e-davidson3:
Going with Paul's approach.

Issue #7990, opened 2020-06-09
Twitter/insta template fix


Chwy5:
For Yes/No for number of followers

1) The number column should allow for 1 decimal place (ie 119.6)

2) The number range should be from 0.1-999.9

3) Remove hundred from dropdown

4) Add an empty space dropdown

Chwy5:
added-closing

PR #7995, opened 2020-06-09
Add number range template inputs

bthaile:
https://github.com/AugurProject/augur/issues/7990
https://github.com/AugurProject/augur/issues/7998

image

image

image

image


bconfortin (approved):
Lgtm

PR #7997, opened 2020-06-09
Updates to trading form

JohnDanz:
Removed redux from trading form.


phoebemirman (changes requested):
initial liquidity in market creation isn't working and also clicking on my own open order behavior isn't correct

phoebemirman (approved):

Issue #7991, opened 2020-06-09
Add new `Augur Markets:` section to Invalid rules


bthaile:
I reverted the code below, it wasn't correct. 

Needs to be one section with no divisions

![image.png](https://images.zenhubusercontent.com/5c00901a5cb43a45a18aeb7f/7f635de4-4eb7-41ec-a6da-4218352dd175)

![image.png](https://images.zenhubusercontent.com/5c00901a5cb43a45a18aeb7f/1ab0ebbc-f10c-4eb8-8bfb-e3bba8cbedae)

```
export const AugurMarkets = [
  'Should resolve using general knowledge if the market does not have a resolution source in market question.',
  'Cover events that occur between market start time and end time in the market question. If start time is not specificed in the market quetion, market creation date/time is used. If no end time is specified in market question, the event expiration is to be used. If the event occurs outside of these bounds, the market should resolve as invalid'
]
```

text has already been added to 
`packages/augur-ui/src/modules/create-market/constants.ts`

new section goes everywhere the invalid rules are 
 - market creation
 - trading page Invalid tooltip
 - reporting form

Chwy5:
This rule needs to be removed from the top section (a market is invalid if:) section - we moved it down below

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/db5f64df-241c-45b0-b465-990f887b5fac)

Chwy5:
fixed-closing

PR #7999, opened 2020-06-09
Added a new section for invalid rules

bconfortin:
#7991


bthaile:
1. tooltip gets too wide depending on size of browser window. 

image

image

2. ticket said *"Needs to be one section with no divisions"* 
in other words, don't have two blocks needed to combine them. 
image

3. also
`'Augur Markets'` needs `:` to be consistent format with `A market is invalid if:`

4. Market Creation Review page is missing new `Augur Markets` section

image

bthaile (changes requested):
need some changes, seem comment.

bthaile (approved):
lgtm

PR #8001, opened 2020-06-09
Disable fee reserve topoff transfer

bthaile:
withdraw all funds for user who has activated trading account and only ever had 40 or less DAI.


andrewdamelio (approved):
LGTM!

PR #8002, opened 2020-06-09
fx btns actions to fix create market review

phoebemirman:


JohnDanz (approved):
seems to fix this issue but not really sure why... either way we should look at this area and improve load/transition times between steps in create market. the first load of review is a noticeable hang, however going back to another page then returning to review is instant. probably has a hangup in it's render function that is waiting on something when we need a loading state to instantly show something to the user.

PR #8004, opened 2020-06-09
various fixes and gas estimate tweaks

nuevoalex:
- reduce gas estimate multipliers
- give a more accurate estimate for trades and reduce gas buffer
- when warp sync file cant be found don't delete the DB


bthaile (approved):
lgtm

Issue #6629, opened 2020-03-06
Apply button is buggy - when scrolling up it will disappear for a moment (seeing this on android chrome) 


matt-bullock:
I'm testing chrome on android and getting this issue

It only appears after I scroll a bit and after that I can scroll the opposite direction and it'll disappear 

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/356de118-2c94-4815-9f6f-6d349ba19690)

matt-bullock:
@bconfortin I'm still seeing some weird behaviour on chrome android. If I scroll to the bottom and then scroll back up the apply button will disappear for a second. Here's a screenshot:

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/ea628786-c97e-4776-83bd-5b0ddeda250a)

Chwy5:
I am moving this to RC2 - more of a performance issue....to replicate....while in the categories filters area, change one of the filters, then scroll up or down very quickly and stop, then go the other direction.....the apply button goes away but does come back

matt-bullock:
Still seeing this. Please test on an android chrome device

matt-bullock:
On my android phone I'm still seeing weird behaviour where I'm able to scroll the markets behind and can see them scrolling in the gap below. Only happens for a brief second but in that time I've accidentally scrolled to the very bottom of the market page behind. It's not a deal breaker but it does look buggy.

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/ec9aaf63-dcd2-4f48-aa96-810f22f4bd93)

matt-bullock:
Not seeing this issue anymore. Closing

PR #8009, opened 2020-06-10
Possible fix for weird glitches involving chrome's address bar

bconfortin:
#6629

Bug happening on Android phones using Chrome. When the address bar was being hidden, some content would get cut or turn transparent at the bottom of the screen.

I think I found a fix. Tested on my Galaxy S9+ and it's not getting cut anymore.

Most common area where this was happening: scrolling down when inside "Categories & Filters" menu would show a gray bar at the bottom. If you had clicked something and the "Apply" button was showing, that button was getting cut. See screenshot bellow for the error. That has been fixed. Have to test on different devices.

Cut Apply Button


bthaile (approved):
lgtm

Issue #8014, opened 2020-06-11
Invalid to Invalid Market


Chwy5:
Change Invalid to Invalid Market 

1)Market cards
2)Market trading pg
     - Order ticket dropdown
     - Outcome Table
     - Order book
3) Reporting/disputing Cards
4)  Reporting/disputing modal

PR #8022, opened 2020-06-11
standardize invalid market label and create a comparison constant to …

bthaile:
https://github.com/AugurProject/augur/issues/8014

replace 'Invalid' text to use  constant. standardized to Invalid Market. There are a few places where outcome description is compared now using constant

image

image

image

image

image


bconfortin (approved):
Lgtm

PR #8028, opened 2020-06-11
SDK Server subscriptions work

robert-e-davidson3:


packages/augur-ui/src/modules/events/actions/listen-to-updates.ts
```diff
@@ -62,6 +64,13 @@ const EVENTS = {
   [SubscriptionEventName.WarpSyncHashUpdated]: wrapLogHandler(
     handleWarpSyncHashUpdatedLog
   ),
+  [SubscriptionEventName.OrderEvent]: wrapLogHandler(
```
bthaile: UI listens for bulk zerox order events not individual order events there is too much chatter and causes too many getter calls. This event wire up in start up events collection.

`[SubscriptionEventName.BulkOrderEvent]: wrapLogHandler(handleBulkOrdersLog),`

Need to follow the same pattern as bulk order events.

bthaile (changes requested):
Can these SDK server zerox order events be a bulk event. UI doesn't subscribe to individual zerox order events.

packages/augur-ui/src/modules/events/actions/listen-to-updates.ts
robert-e-davidson3: Fixed!

pgebheim (approved):

packages/augur-ui/src/modules/events/actions/log-handlers.ts
```diff
@@ -405,6 +405,14 @@ export const handleBulkOrdersLog = (data: {
   }
 };
 
+export const handleDBUpdatedZeroXOrders = (log: StoredOrder) =>
```
bthaile: if UI is processing bulk orders this isn't needed.

bthaile (changes requested):
remove individual order event processing in UI. It should only be processing bulk zeroX order events

packages/augur-ui/src/modules/events/actions/log-handlers.ts
pgebheim: Agreed, there should be zero changes to which events are generated or listened to.

As an optimization we can allow each SDK consumer to subscribe to the relevant events in the future.

robert-e-davidson3: Good catch. I missed this when I switched to bulk orders. Deleted.

bthaile (approved):
lgtm

Issue #8021, opened 2020-06-11
Firefox mobile: Filters getting hidden or covered by apply button (unable to scroll to see them)


matt-bullock:
This is on a google pixel 3 phone, so quite a common size. Show filters button is hard to hit as it's getting hidden:

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/54095fa4-e09c-424a-a37a-59d4f8902e99)

If the apply button is active then it gets totally hidden:

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/b9f2e153-e602-44f6-9f13-3c89a7231951)

If I do manage to open the filters then the invalid radio buttons get hidden:

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/79e714ff-bdf8-4ec4-9839-f29943570d22)

matt-bullock:
This works well now. Closing

PR #8029, opened 2020-06-11
Added a fix for both chrome and firefox

bconfortin:
Tested it on my phone and it seems to work well with firefox and chrome.

#8021


bthaile (approved):
ok, that is weird but lets run with it

PR #8034, opened 2020-06-12
changed 2% to 5% because gas is more expensive these days

joeykrug:


bthaile (approved):
lgtm, created ticket to have this value float. <https://github.com/AugurProject/augur/issues/8061>

PR #8037, opened 2020-06-12
Betting ui onboarding fixes

andrewdamelio:
- fixed modals to show when appStatus.modal gets updated dynamically
- fixed style regressions with onboarding modals
- fixed onboarding modals to trigger on sign up
- some container cleanup

TODO
- seeing issues with activation account button when fired, looking into that


packages/augur-ui/src/modules/app/store/app-status-hooks.ts
```diff
@@ -710,7 +710,10 @@ export const useAppStatus = (defaultState = DEFAULT_APP_STATUS) => {
         dispatch({ type: SET_CATEGORY_STATS, categoryStats }),
       updateFilterSortOptions: filterSortOptions =>
         dispatch({ type: UPDATE_FILTER_SORT_OPTIONS, filterSortOptions }),
-      setModal: modal => dispatch({ type: SET_MODAL, modal }),
+      setModal: modal => {
+        dispatch({ type: CLOSE_MODAL });
```
JohnDanz: we shouldn't need to close modal before setting modal, setModal overwrites whatever's already set for modal with whatever you pass it.

packages/augur-ui/src/modules/modal/components/modal-view.tsx
```diff
@@ -184,6 +187,8 @@ const ModalView = ({
     }
   };
 
+  const appStatus = useAppStatusStore();
```
JohnDanz: I would break this out as `const { modal } = useAppStatusStore();` since we will never or should never be passing a modal to modalView, so the modal should always come from the app status state.

JohnDanz (changes requested):
two comments, otherwise looks good

JohnDanz (approved):

PR #8045, opened 2020-06-15
0x-10

pgebheim:
You need to put the 10.0 mesh bundle into the monorepo root named
`zerox.wasm` for this branch to load.


pgebheim:
Merged with another PR

PR #8057, opened 2020-06-16
Migrate rep signing wallet

bthaile:
give user ability to migrate REP if in signing wallet or in trading account.


andrewdamelio (approved):
LGTM

PR #8076, opened 2020-06-17
add sort-by fees generated

andrewdamelio:
Works along side https://github.com/AugurProject/augur/pull/8090
- Add Sort by fees generated, we wanna sort by the amount of fees generated
- Make this sort default in the UI


packages/augur-sdk/src/state/getter/Markets.ts
```diff
@@ -744,6 +745,18 @@ export class Markets {
...
+    marketData = marketData.map(market => {
+      const m = orderFilledLogsByMarket[market.market];
+      const numberOfTrades = m?.length || 0;
+      const feesGenerated = (new BigNumber(Number(market.feePercent))).times(numberOfTrades);
+
+      return {
+        ...market,
+        feesGenerated: feesGenerated.toNumber(),
+      }
+    });
```
pgebheim: We can make this simpler and JUST sort by the number of trades :-D

```suggestion

      return {
        ...market,
        numberOfTrades,
      }
    });
```

Also, can you move this code into the MarketDB reducer? There's boiler plate in there which recalculates stuff on the MarketDB on a trade. That would result in the original query having a `numberOfTrades` property just like it has `lastTradeTimestamp` now -- and then the getter doesn't need to fetch all this data.

pgebheim (changes requested):

packages/augur-sdk/src/state/getter/Markets.ts
andrewdamelio: 👍 can do!

nuevoalex (approved):
👍

pgebheim (approved):

PR #8083, opened 2020-06-18
fixed incorrect character

bthaile:
fixed incorrect character
fixed typos that msagansk pointed out in UI channel


bconfortin (approved):
Lgtm

PR #8094, opened 2020-06-19
Deterministic wallets

nuevoalex:
Update GSNV2 wallet registry and refactor to use a generic factory so we can have deterministic wallets


packages/augur-core/src/contracts/gsn/v2/BatchForwarder.sol
```diff
@@ -2,15 +2,21 @@
...
 
 /**
  * batch forwarder support calling a method sendBatch in the forwarder itself.
  * NOTE: the "target" of the request should be the BatcherForwarder itself
  */
-contract TrustedBatchForwarder is TrustedForwarder, BaseRelayRecipient {
+contract BatchForwarder is Forwarder, BaseRelayRecipient {
+
+    function versionForwarder() external view returns (string memory){
+        return "2.0.0-alpha.1+opengsn.batched.iforwarder";
```
bthaile, line 16: are we going to be stuck with this `version` text? What about when version 2.0.0 comes out for real?

packages/augur-core/src/contracts/gsn/v2/Forwarder.sol
```diff
@@ -2,61 +2,69 @@
...
 
     // solhint-disable-next-line max-line-length
-    bytes32 public constant RELAYDATA_TYPEHASH = keccak256("RelayData(address senderAddress,uint256 senderNonce,address relayWorker,address paymaster)");
+    bytes32 public constant RELAYDATA_TYPEHASH = keccak256("RelayData(address senderAddress,uint256 senderNonce,address relayWorker,address paymaster,address forwarder)");
+
+
+    string public versionSM = "augur-forwarder";
+
+    function versionForwarder() external view returns (string memory) {
+        return "2.0.0-alpha.1+opengsn.forwarder.iforwarder";
```
bthaile, line 28: seems to be in multiple places, i assume they won't have to be updated in the future

bthaile (approved):
lgtm, so does this give GSN V1 and V2 wallet support? no changes in SDK?

packages/augur-core/src/contracts/gsn/v2/BatchForwarder.sol
nuevoalex: yea its just to satisfy an interface

nuevoalex:
yea no SDK chnages needed since we dont even have the GSNV2 SDK work in yet

PR #8095, opened 2020-06-19
Performance Improvements Mega-branch

pgebheim:
This behemoth of a branch re-arranges the entire codebase to allow splitting. This integrates with Alex's augur-sdk-lite in order to provide a place to put stuff that doesn't depend on the whole world. The end result is significantly smaller entrypoint bundles and lazy loading.

Note: This still does not only use the SDK-lite for hotloading but we gotta get this thing in and other eyes on it.

NB: There are some potential issues caused by us using `import type` now -- which can mean that enums are imported they will be undefined at runtime. All enums have been moved to `@augurproject/sdk-lite` so any issues of this manner usually just require moving the import. We've gone through lots of things but its possible something is missed due to the lack of typechecking in the UI codebase.


nuevoalex (approved):
Mostly just looked at the hot loading changes. LGTM

bthaile (approved):
I made some fixes to references in UI to "fix" some blow ups. @pgebheim to move utility methods around so that the UI doesn't reference `/sdk`

Issue #8077, opened 2020-06-17
Soccer Championship market year validation issue


Chwy5:
![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/491a02fe-eb92-4a0a-9168-ef734f541852)

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/b2f08060-81d5-4624-8439-58480c5caf2f)

Chwy5:
fixed -closing

PR #8100, opened 2020-06-19
Add year valiation checks templates

bthaile:
https://github.com/AugurProject/augur/issues/8077

template changes to support year validation.
year in question can not be before market creation or after end time.
ending year range in question can not be before market creation or after end time.


bconfortin (approved):
Lgtm

PR #8109, opened 2020-06-22
Update get markets sportsbook

bthaile:
changes to getMarkets getter for sportsbook
1. categories based on liquidity pools not individual markets
2. added liquidityRank which is liquidityInfo 10% value


bthaile:
John moved code to betting-ui based PR. closing

Issue #8108, opened 2020-06-22
Add Template Validation Tests


Chwy5:
add validation to make sure the template is in the categories that we put the template in. 
ie.   sports will have to be in sports.
baseball will have to be in sports->baseball ...

Chwy5:
Needs to be tested through api

PR #8112, opened 2020-06-22
add template validations to verify markets are in the correct categor…

bthaile:
…ies based on template structure

1. fixed validation for opening gross sales in entertainment templates
2. added category validation checks.

https://github.com/AugurProject/augur/issues/8108


bconfortin (approved):
Lgtm

PR #8136, opened 2020-06-23
remove-dispatch

JohnDanz:
removed all traces of dispatch being used/called in the UI (redux dispatch) still a few containers laying around but not many.


bthaile (approved):
lgtm

Issue #8129, opened 2020-06-23
Only get certain balance on login


bthaile:
In an effort to reduce eth calls per block. 

Only check users signing wallet for dai, rep and legacy rep on login. 

Only check Legacy rep balance on trading account on login. 

This will reduce balance calls in half on each block.

Chwy5:
works-closing

PR #8138, opened 2020-06-23
only get rep,dai,eth for trading account and eth on signing wallet ea…

bthaile:
…ch block. others only on login or when a transfer has confirmed from UI

https://github.com/AugurProject/augur/issues/8129

user will have to logout and back in if they transfer rep, dai, leg rep to signing wallet or 
or leg rep to trading account.


bconfortin (approved):
Lgtm

Issue #8133, opened 2020-06-23
Filter by market type


Chwy5:
We should add a filter by Market type to go under the filters section (After liquidity spread before Invalid Markets filter

Market Types:

1) All (preset)
2) Yes/No
3) Categorical
4) Scalar

Chwy5:
added - closing

PR #8146, opened 2020-06-23
give user ability to filter based on market type

bthaile:
https://github.com/AugurProject/augur/issues/8133

image

1. save user selection to settings in localStorage
2. add market type as query param as `type`
3. added db filter in sdk


bconfortin (approved):
Lgtm. Can't believe we didn't have those filters yet. Looks good, tested on mobile, desktop and using the query parameters.

PR #8147, opened 2020-06-24
Add wallet connect

jfarid27:
# Description

Adds [WalletConnect](https://walletconnect.org/) as a login type, allowing [these Wallets](https://walletconnect.org/wallets) to directly connect to Augur.

# Testing
1. On a phone, go to `test.walletconnect.org`.
2. Select `Kovan` or your current network.
3. Select `Connect to WalletConnect`.
4. When the QR code appears, select `Scan` on the WalletConnect test account.

# TODO
- [ ] Get WalletConnect Icon
- [ ] Figure out `getSigners` missing function on provider. This fails when creating the account object.
- [ ] Correctly cleanup. After logging out, you cannot log into a different login type.

# DOD
- [ ] Assert users can log in using WalletConnect.
- [ ] Assert users can log out using WalletConnect.
- [ ] Assert users can log back into WalletConnect with a new account (Assert cleanup).
- [ ] Assert users can log into a separate Wallet type after logging out of WalletConnect (Assert cleanup).


pgebheim:
This is a cool patch (and thank you for the contribution). Currently the WC interface requires that you bring your own ETH node, which none of the other parts of Augur require (we piggyy back wallet connections). While not technically impossible its a hurdle to think about (clients aren't gonna wanna specify that, lets be honest).

Gonna back-burner a full WC for now but I think @andrewdamelio can take a look at the `signer` issue and get this unblocked -- would be good to have the working bones for it ready when we can take a look at how to integrate with another provider for the actual eth connection.

jfarid27:
> Currently the WC interface requires that you bring your own ETH node, which none of the other parts of Augur require (we piggyy back wallet connections).

Does this mean the required InfuraId? This actually seems to work just throwing in the WalletConnect documentation ID. I could talk to them and see if that really is required.

pgebheim:
The way this works would require a way to talk to an Eth node -- since the WC provider more or less just signs transactions as far as i know. Using a publicly published Infura ID will ultimately just be rate limited or blocked.

One option here is to combine providers to piggy-back on the eth connection of another, but this is a bit off because the wallet could just block the app for using resources without funneling any users through it.

I actually think this is one of the big downsides of WC as it really assumes centralized infrastructure in multiple ways.

jfarid27:
> it really assumes centralized infrastructure in multiple ways.

Got it. so @pgebheim I also added the provider itself in the `signer` property of the accountObject. This does end up "working" and the new account shows up, but I'm in this weird state where I can't clear the previous wallet connection provider state after logging out. It's another thing to be aware of but I'm just not familiar with the Augur application state.

pgebheim:
Closing for now -- can revisit if we decide to add this.

PR #8156, opened 2020-06-24
Add frozen funds breakdown

bthaile:
1. new getter for frozen funds breakdown
2. change style of unprofitable trade notice in order form
3. fixed res rules for sports scalar markets

```
await AugurSDK.getUserFrozenFundsBreakdown({
  account: state.loginAccount.address,
  universe: state.universe.id
})
```
to call the getter


packages/augur-sdk/src/state/getter/Users.ts
```diff
@@ -1016,74 +1126,13 @@ export class Users {
     db: DB,
     params: t.TypeOf<typeof Users.getTotalOnChainFrozenFundsParams>
   ): Promise<UserTotalOnChainFrozenFunds> {
-    if (!params.universe && !params.account) {
+    if (!params.universe || !params.account) {
```
bconfortin, line 1129: Error message should say `requires a 'universe' and 'account'`, not or

bconfortin (approved):
Lgtm

packages/augur-sdk/src/state/getter/Users.ts
bthaile: good one. updated.

PR #8166, opened 2020-06-25
cache additional block info requests

nuevoalex:


packages/ethersjs-provider/src/EthersProvider.ts
```diff
@@ -95,10 +98,12 @@ export class EthersProvider extends ethers.providers.BaseProvider
             let result: Promise<any>;
             const itemString = JSON.stringify(item);
             if ((item.message === "call" ||
-                 item.message === "getBalance"
+                 item.message === "getBalance" ||
```
pgebheim, line 101: :hear_no_evil:

pgebheim (approved):

PR #8170, opened 2020-06-25
Clean up subscription handling

pgebheim:
The Subscription class was intended to be used for a server to multiplex events to multiple clients, so I've restored it to its original purposes for the WS server. Everywhere else the Augur#events emitter is used which is just an EventEmitter.

This also fixes a stupid bug where the constants for Mesh/RPC order events were using the same string and thus all events were being duplicated.


nuevoalex (approved):
👍

PR #8171, opened 2020-06-25
use Fortmatic for email login

andrewdamelio:
- remove Portis
- add Fortmatic as email options
- other minor fixes/updates


yarn.lock
```diff
@@ -3477,7 +3477,7 @@
     xhr "^2.2.0"
     xtend "^4.0.1"
 
-"@portis/web3@2.0.0-beta.55", "@portis/web3@^2.0.0-beta.54":
+"@portis/web3@^2.0.0-beta.54":
```
bthaile, line 3480: why is this getting pulled into yarn.lock. I would expect all portis stuff to be removed.

andrewdamelio: looks like its being used by a upstream dependency, `3id-connect`

bthaile (approved):
lgtm, it would have been nice to fully remove all portis deps. Andy saw that 3id has a dep on portis web3 package

Issue #8142, opened 2020-06-23
Users frozen funds breakdown modal


matt-bullock:
Add an underline to the frozen funds amount figure to indicate it's clickable

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/e6e72c78-cd8c-4628-8001-960dce1b4e67)

This will open the frozen funds modal 

Design for modal (both desktop and mobile):

https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=15832%3A16312

matt-bullock:
On mobile. The modal goes bigger than my phone screen. Needs to add some spacing between top and bottom.

The top and header and footer components look twice the size they should be.

~~Cut market titles off after 2 lines and add ...~~

updated: Cut market titles off after `1` line and add ...

pgebheim:
![image](https://user-images.githubusercontent.com/86010/86024826-53544300-b9e2-11ea-9d74-fa9b6a0aecac.png)

PR #8176, opened 2020-06-25
Add frozen funds breakdown

bthaile:
https://github.com/AugurProject/augur/issues/8142
1. shows user's frozen funds breakdown
2. market titles load as they are loaded into state
3. there is a lag when the modal opens, it takes a bit to get the data from sdk.

image


bconfortin (approved):
Lgtm

PR #8179, opened 2020-06-25
changed up conditional to get markets when warpsync and reporting sta…

bthaile:
…te filters are passed in to getMarkets


nuevoalex (approved):
👍

Issue #8161, opened 2020-06-24
Resolution date and time not adjusting for market creators timezone when MC is ahead of New York time


matt-bullock:
Prevents MC from being able to create the market due to the error message

Issues so far are with Finance and sports templates

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/b278e390-ca90-448b-b5be-0eeeb070011a)

On the screenshot below the conversion to UTC time should be 9pm, not 10pm

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/0e61329c-edf3-46b5-a144-dc5ba9d098f8)

matt-bullock:
Works now. closing

matt-bullock:
all looks correct

PR #8184, opened 2020-06-25
removed utc() on getting start date so it matches the date the user p…

bthaile:
…icks in template market creation

https://github.com/AugurProject/augur/issues/8161


bconfortin (approved):
Lgtm

Issue #8128, opened 2020-06-23
transaction modal can't search


bthaile:
trying to filter on action results in errors. 

![image.png](https://images.zenhubusercontent.com/5c00901a5cb43a45a18aeb7f/2d85e03e-7c03-4ea9-96bf-dbbeef918624)

to repro
have trades, claim proceeds, claim rep and fees, created markets and try to look them up in transaction modal using the filter

Chwy5:
works - closing

PR #8185, opened 2020-06-25
Removed one of the actions

bconfortin:
#8128 

I talked to Alex and there is no call in the DB to get the Claimed Market Creator Fees and that is rather complicated to add. At this point of the game, we decided to remove that from the Transactions modal and add it post v2.


bthaile (approved):
lgtm

Issue #8152, opened 2020-06-24
Initial liquidity `Sign in to approve orders` message bug


Chwy5:
I am getting a message to sign to approve orders even though i dont have any orders
I have a notification in acct summary pg for it as well

1) create market
2) Add pre-liquidity orders
3) Go back and cancel those orders
4) complete market creation
5) once market is created the message appears even though I have no orders (and does not disappear)

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/a0b9a594-9142-47be-a161-38f79315938b)

Chwy5:
fixed -closing

PR #8188, opened 2020-06-25
fix issue where removing all pre-liquidity orders still causes 'sign …

bthaile:
…orders ...' link is shown after market is created when there is no orders to sign

https://github.com/AugurProject/augur/issues/8152


bconfortin (approved):
Lgtm

Issue #8177, opened 2020-06-25
Market Creation Form Hangs on Keypresses


nuevoalex:
When typing normally in the market creation textareas the UI will hang. This is because there is an onChanged handler for the form which calls "updateNewMarket" which takes ~200msec per call.

We should either make this not so expensive or call "updateNewMarket" much less frequently

Chwy5:
It is also effecting the dropdowns in the market creation....big delay

Chwy5:
seems much faster now

PR #8206, opened 2020-06-26
Little better input market creation

bthaile:
https://github.com/AugurProject/augur/issues/8177

with throttle


packages/augur-ui/src/modules/create-market/components/form.tsx
```diff
@@ -724,6 +725,8 @@ export default class Form extends React.Component<FormProps, FormState> {
       });
     }
     this.onError(name, '');
+    console.log(updatedMarket);
```
bconfortin, line 728: Is that console.log() intended?

bconfortin (approved):
Awesome. Lgtm

PR #8210, opened 2020-06-26
make frozen funds modal smaller on mobile so it is the same as other …

bthaile:
…info modals. only show one line of market title

https://github.com/AugurProject/augur/issues/8142

update to use more standard modal components.
image


bthaile:
header and footer are too big for mobile. going to fix.

bconfortin (approved):
Lgtm

PR #8213, opened 2020-06-26
Text Input gets debounce

bthaile:
1. fix sig of our debounce method
2. remove use of lodash debounce, apply debounce to TextInput so all text inputs benefit


bconfortin (approved):
Good one! Lgtm :D

Issue #8201, opened 2020-06-26
order ticket scrolling within it's own frame on mobile


matt-bullock:
Makes the page feel clunky as it forces you to scroll twice + cuts off dropdowns 

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/13fbef83-279e-474f-8fcc-21b152a37ec4)

matt-bullock:
I'm still seeing this

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/f8454dfc-ce4f-402d-857e-25779720fe3b)

matt-bullock:
looks fixed

PR #8234, opened 2020-06-29
Made the dropdown dropup when the tip is not showing

bconfortin:
#8201 

Captura de Tela 2020-06-29 às 17 02 52
Captura de Tela 2020-06-29 às 17 02 24

### DIFF
--- a/packages/augur-ui/src/modules/trading/components/form.tsx
+++ b/packages/augur-ui/src/modules/trading/components/form.tsx
@@ -104,6 +104,7 @@ interface FromProps {
   gasPrice: number;
   getGasConfirmEstimate: Function;
   endTime: number;
+  showTip: boolean;
 }
 
 interface TestResults {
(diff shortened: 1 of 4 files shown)


packages/augur-ui/src/modules/trading/components/form.tsx
```diff
@@ -1183,6 +1185,7 @@ class Form extends Component<FromProps, FormState> {
                     )}
                     <SquareDropdown
                       defaultValue={s.expirationDateOption}
+                      openTop={!(showTip && !initialLiquidity)}
```
bthaile: I don't see how `showTip` and dropdown correspond. It's confusing. 

showTip is to show a tip message at the bottom of the order form not a height. Use a different approach to determine if the dropdown will not be visible when opened.

bthaile (changes requested):
Need better approach to determine if selector will expand out of component.

bthaile (approved):
works, we've had discussions about replacing isMobile with media queries in the future.

PR #8247, opened 2020-06-30
Fix local utc mixup

bthaile:
UI was mixing up local time and utc time when processing date times in market creation.
remove `DATESTART` and use `DATEYEAR`
 * fixed NFL template that was using year range should have been using YEARS
 * fixed grammar in on-boarding modal.
 * validate on market question date string rather than timestamp (could have been a validation exploit)
 * added helper to convert utc timestamp to formatted date object, existing was using local timestamp when it was actually utc timestamp from contracts.

https://github.com/AugurProject/augur/issues/8161

test Financial templates with date in market question. verify that event expiration is 1 to 2 hours after market close on same day in market question.


packages/augur-ui/src/modules/create-market/get-template.ts
```diff
@@ -395,14 +393,18 @@ export function getEventExpirationForExchange(
   if (!closing) return null;
   const dateYearSource = inputs.find(i => i.id === closing.inputDateYearId);
   const timeOffset = closing.userInputObject as TimeOffset;
-  if (dateYearSource && dateYearSource.setEndTime && timeOffset) {
-    const closingDateTime = getTemplateExchangeClosingWithBuffer(
-      dateYearSource.setEndTime,
+  if (dateYearSource && dateYearSource.userInput && timeOffset) {
```
bconfortin, line 396: I believe `dateYearSource && dateYearSource.userInput` could be simplified to `dateYearSource?.userInput`

bconfortin (approved):
Lgtm

PR #8251, opened 2020-06-30
update sanitize URL

### LINKED ISSUE #8222: Sharing links should sanitize URLs
URL in the market description means that twitter/facebook will expand that link instead of the sharing link.

Sanitize these with some simple regexp 

```
title.replace(/(\w+)\.([a-zA-Z]{2,9})/, "$1 . $2")
```

andrewdamelio:
https://github.com/AugurProject/augur/issues/8222

Screen Shot 2020-06-30 at 2 16 48 PM

@pgebheim let me know if you think i should add any other TLD's


pgebheim:
@andrewdamelio After some fiddling I came up with this. It's possible of false positives because urls are weird but I think will work more abstractly without pulling in `tldts` or something.
```javascript
title.replace(/(([a-zA-Z\-_]+)\.)+?\/?/g, "$2(dot)").replace(/\(dot\)(\s|$)/, ".$1")
```

Examples:
image

```javascript
[
"This is some test text via www.tradingview.com",
"This could be a title with cdc.gov in it. ",
"This has a title with predictions.global in it and no space at the end.",
"Here is a good market. With the resolution via predictions.markets",
"https://www.tradingview.com/aasd",
"cdc.gov/blah",
"https://google.com/asdasd",
"And this one includes a http://weird_url.com/index.html And I wonder if it will work"
].map((u) => {
  return u.replace(/(([a-zA-Z\-_]+)\.)+?\/?/g, "$2(dot)").replace(/\(dot\)(\s|$)/, ".$1")
});
```

pgebheim (changes requested):
See feedback :point_up:

pgebheim (approved):

andrewdamelio:
@pgebheim Thanks Paul for updating this yesterday

PR #8257, opened 2020-07-01
getter cache

nuevoalex:
Cache the results of getter requests and evict either:
 - After a successful transaction (for account specific getters)
 - After a relevant log is processed
 - After some maximum time to live for all cached results


bthaile (approved):
Looks amazing. Im super excited

Issue #8068, opened 2020-06-17
Pre-liquidity orders don't cause bell alert


Chwy5:
I created a new market and added pre-liquidity (passed the spread filter)....My market gets created (and I get a toast message to confirm this) I sign my pre-liquidity orders.....they do show up in open orders, but I do NOT get a toast/bell message for them and my newly created market still shows liquidity depleted.

liquidity depleted goes away after refreshing......but i still get no bell messages for those orders being created

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/3a40136b-20ab-4569-9553-8c12cef82ec8)

bthaile:
The liquidity label is based on a property that is calculated in the sdk, it gets re-calculated (it’s an Resource intensive calculation) every minute or so. Therefore the label will go away a minute or so after orders have been created.

matt-bullock:
Seeing it in the bell now

PR #8258, opened 2020-07-01
Added bell alerts for unsigned orders being created

bconfortin:
#8068 

On `market-place-trade.ts`, there is a conditional before dispatching the alert, which is the following:

```javascript
if (tradeInProgress.numFills === 0) {
  // do alert things
}
```

We don't have the `tradeInProgress` object inside `liquidity-management.ts`. Does that make any difference?

Image of all the orders to the left, before signing them. When I signed all of them, the bell alerts to the right appeared.
Captura de Tela 2020-06-30 às 22 08 16


bthaile (approved):
works, eventually we'll get all the bell alert logic in one place.

PR #8259, opened 2020-07-01
Always hotload market info.

pgebheim:
This ensures that if the market info in the database is tampered with
externally the user is presented with the actual market details when
viewing the detail page.

Not 100% certain this is the right approach -- we could perhaps only do this on the first time the market details are loaded?


justinbarry:
@pgebheim Lets talk about this tomorrow. I think it was correct before-ish. We probably need to add a bit about hot loading on marketid change.

pgebheim:
Correct @justinbarry I'm gonna close this and we'll do it right.

Issue #8182, opened 2020-06-25
Need `transfer my REP` modal just like we have for `transfer my DAI`


bthaile:
need to detect user has v2 REP in signing wallet on login. 
add `transfer my REP` modal to be brought up just like `transfer my dai` modal under account dropdown

matt-bullock:
this works

PR #8263, opened 2020-07-01
transfer my dai make more generic to be used for rep and dai when in …

bthaile:
…signing wallet

https://github.com/AugurProject/augur/issues/8182


andrewdamelio (approved):
Lgtm!

Issue #8274, opened 2020-07-02
Participation Tokens should display Rep in Transaction pg


Chwy5:
![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/3bd6caee-29d3-49c0-baf8-b0751acd19fa)

matt-bullock:
showing as REP

PR #8277, opened 2020-07-02
change coin from dai to rep for participation tokens in transaction m…

bthaile:
…odal
https://github.com/AugurProject/augur/issues/8274

image


bconfortin (approved):
Lgtm

PR #8292, opened 2020-07-03
dont return BNs from getters

nuevoalex:


bthaile (approved):
Lgtm, some good clean up. Like the added interface

PR #8298, opened 2020-07-03
fix warning with new icon, default dropdown to first value

bthaile:
market creation set 3rd category to first option in list.

image


andrewdamelio (approved):
LGTM

PR #8308, opened 2020-07-04
if there isn't a warp sync file, hotloading markets doesn't return an…

bthaile:
…y markets and limit gets set to 0 which causes getMarkets to not return any markets


nuevoalex (approved):
👾

Issue #8296, opened 2020-07-03
Refreshing pg, Resets my sorts


Chwy5:
If I set my sort by on the market list pg and then need to refresh my pg, bc i am too many blacks behind, my sort refreshes....should stay at what i set it to

Chwy5:
fixed - closing

PR #8311, opened 2020-07-05
Add sort limit offset to settings

bthaile:
https://github.com/AugurProject/augur/issues/8296

save user settings when certain filter sort options are changed. reduce number of dispatches to save filter sort options

new settings to save
 * market reporting state
 * market sort

new filter sort options
 * market list reporting state
 * market list sort
 * limit (market cards per page)
 * offset (page number)

new query parameters processed
 * limit
 * offset
 * market list reporting state
 * market list sort


andrewdamelio (approved):
Nice update!

PR #8312, opened 2020-07-06
reporting only UI, add modal, disable market creation button and trad…

bthaile:
…ing button, add notice in trading form

Not sure how to build a reporting only UI. I just tested by hard coding `config.ui.reportingOnly: true` in init-augur `initAugur` method before `dispatch(updateEnv(config));` call

Screen Shot 2020-07-06 at 09 09 15

Screen Shot 2020-07-06 at 09 09 35

Screen Shot 2020-07-06 at 09 09 50

If user lands on scalar market, the reporting only modal shows then regular scalar modal.


andrewdamelio (approved):
lgtm

PR #8349, opened 2020-07-07
Catch liquidity pools

JohnDanz:
Adds liquidity pools and updated the market cards to handle new event emitter. this is still not a 100% complete feature as we still need the initial getter in order to get the latest liquidityPools state as the event emitter only emits changes/updates. This is a good enough to go on and does progress the sportsbook so going to put it in for now, and add that final getter call when it's been made.


packages/augur-ui/src/modules/modal/signin.tsx
```diff
@@ -119,15 +119,15 @@ export const SignIn = ({ isLogin }) => {
       // type: ACCOUNT_TYPES.PORTIS,
       icon: EmailLogin,
       text: `${LOGIN_OR_SIGNUP} with Email`,
-      subText: `Powered by ${ACCOUNT_TYPES.FORMATIC}`,
+      subText: `Powered by ${ACCOUNT_TYPES.FORTMATIC}`,
```
andrewdamelio, line 122: ;)

andrewdamelio (approved):
easy to parse, big fan of betting-ui
LGTM

PR #8370, opened 2020-07-08
track best offer again

JohnDanz:
tracking the best offer for staging env, thought we used dev for dev.augur.net but we use staging.


bthaile (approved):
lgtm

PR #8371, opened 2020-07-08
needed to pass in token name to transfer my token component

bthaile:
missing tokenName property 
it's needed to get token balance for display

Screen Shot 2020-07-08 at 09 40 10

after fix
Screen Shot 2020-07-08 at 09 41 56


andrewdamelio (approved):

bthaile:
@Mergifyio backport dev

PR #8404, opened 2020-07-09
Fix custom end time formatting

bthaile:
parsing timestamp and adding hours/minutes doesn't work well for moment when using methods like `add` and .hours(). better to break down timestamp into formatted string and re-parse as needed. 

I made a mistake in previous PR displaying date in local time, needed to use regular existing helper method to formatted date. removed the UTC format date helper, not needed.


andrewdamelio (approved):
Lgtm

bthaile:
@Mergifyio backport dev

PR #8405, opened 2020-07-09
dev updates for AddFunds and account summary

andrewdamelio:

### DIFF
--- a/packages/augur-ui/src/modules/auth/helpers/login-account.ts
+++ b/packages/augur-ui/src/modules/auth/helpers/login-account.ts
@@ -115,9 +115,9 @@ export const getCoreStats = (isLogged, loginAccount) => {
 };
 
 export const getEthReserve = () => {
-  const { loginAccount: { balances } } = AppStatus.get();
-  const ethNonSafeBN = createBigNumber(balances.ethNonSafe);
-  let desiredSignerEthBalance = createBigNumber(formatAttoEth(Number(DESIRED_SIGNER_ETH_BALANCE)).value);
+  const { env, loginAccount: { balances } } = AppStatus.get();
+  const ethNonSafeBN = createBigNumber(balances.signerBalances.eth || 0);
+  let desiredSignerEthBalance = createBigNumber(formatAttoEth(Number(env.gsn.desiredSignerBalanceInETH * 10**18)).value);
   if (ethNonSafeBN.lt(desiredSignerEthBalance)) desiredSignerEthBalance = ethNonSafeBN;
   const reserveEthAmount: FormattedNumber = formatEther(desiredSignerEthBalance, {
     zeroStyled: false,
--- a/packages/augur-ui/src/modules/auth/helpers/login-account.ts
+++ b/packages/augur-ui/src/modules/auth/helpers/login-account.ts
@@ -115,9 +115,19 @@ export const getCoreStats = (isLogged, loginAccount) => {
 };
 
 export const getEthReserve = () => {
-  const { env, loginAccount: { balances } } = AppStatus.get();
-  const ethNonSafeBN = createBigNumber(balances.signerBalances.eth || 0);
-  let desiredSignerEthBalance = createBigNumber(formatAttoEth(Number(env.gsn.desiredSignerBalanceInETH * 10**18)).value);
+  const {
+    env: {
+      gsn: { desiredSignerBalanceInETH: gsnETHBalance },
+    },
+    loginAccount: {
+      balances: {
+        signerBalances: { eth: signerEth },
+      },
+    },
+  } = AppStatus.get();
+
+  const ethNonSafeBN = createBigNumber(signerEth || 0);
+  let desiredSignerEthBalance = createBigNumber(formatAttoEth(Number(gsnETHBalance * 10**18)).value);
   if (ethNonSafeBN.lt(desiredSignerEthBalance)) desiredSignerEthBalance = ethNonSafeBN;
   const reserveEthAmount: FormattedNumber = formatEther(desiredSignerEthBalance, {
     zeroStyled: false,
--- a/packages/augur-ui/src/modules/modal/add-funds.tsx
+++ b/packages/augur-ui/src/modules/modal/add-funds.tsx
@@ -31,12 +31,6 @@ import Styles from 'modules/modal/modal.styles.less';
 
 interface AddFundsProps {
   autoSelect: boolean;
-  tokenToAdd: string;
-  ETH_RATE: BigNumber;
-  REP_RATE: BigNumber;
-  config: SDKConfiguration;
-  initialAddFundsFlow?: string;
-  initialSwapToken?: string;
 }
 
 const addFundsFortmatic = async (amount, crypto, address) => {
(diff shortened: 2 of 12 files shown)


packages/augur-ui/src/modules/auth/helpers/login-account.ts
```diff
@@ -115,9 +115,9 @@ export const getCoreStats = (isLogged, loginAccount) => {
 };
 
 export const getEthReserve = () => {
-  const { loginAccount: { balances } } = AppStatus.get();
-  const ethNonSafeBN = createBigNumber(balances.ethNonSafe);
-  let desiredSignerEthBalance = createBigNumber(formatAttoEth(Number(DESIRED_SIGNER_ETH_BALANCE)).value);
+  const { env, loginAccount: { balances } } = AppStatus.get();
```
JohnDanz: destruct this so it's `{ env: { gsn: { desiredSignerBalanceInETH: gsnETHBalance } }, loginAccount...`

then later:
`let desiredSignerEthBalance = createBigNumber(formatAttoEth(Number(gsnETHBalance * 10**18)).value);` -- been trying to favor destructuring instead of `foo.bar.biz`

packages/augur-ui/src/modules/modal/add-funds.tsx
```diff
@@ -58,42 +51,41 @@ const addFundsTorus = async (amount, address) => {
...
   autoSelect = false,
-  tokenToAdd = DAI,
 }: AddFundsProps) => {
+  const BUY_MIN = 20;
+  const BUY_MAX = 250;
   const { loginAccount, modal, actions: {closeModal} } = useAppStatusStore();
-  const isRelayDown = false;
-  const { useSigner, initialSwapToken, initialAddFundsFlow } = modal;
-  const showTransfer = modal.showTransfer || false;
+  const { env: config, ethToDaiRate, repToDaiRate } = useAppStatusStore();
```
JohnDanz: these should all be in the same const definition, not sure why we have 2 calls to `useAppStatusStore()`

packages/augur-ui/src/modules/modal/add-funds.tsx
```diff
@@ -58,42 +51,41 @@ const addFundsTorus = async (amount, address) => {
...
+
+  if (!loginAccount) {
+        return null;
+  }
 
   const usingOnRampSupportedWallet = accountMeta &&
     accountMeta.accountType === ACCOUNT_TYPES.TORUS ||
     accountMeta.accountType === ACCOUNT_TYPES.FORTMATIC;
 
+  const [amountToBuy, setAmountToBuy] = useState(createBigNumber(0));
```
JohnDanz: useStates/hooks should always be as high up as possible. the `if (!loginAccount)` above will cause issues if this exits early and this hook is unreachable. Always put hooks as high as they can go and never have a way for the component to return before it would run each hook.

packages/augur-ui/src/modules/modal/add-funds.tsx
```diff
@@ -58,42 +51,41 @@ const addFundsTorus = async (amount, address) => {
...
   const ETH_RATE = createBigNumber(1).dividedBy(
     ethToDaiRate?.value || createBigNumber(1)
   );
   const REP_RATE = createBigNumber(1).dividedBy(
     repToDaiRate?.value || createBigNumber(1)
   );
-  
+
+  const { useSigner = false,  initialAddFundsFlow = null, initialSwapToken= null, tokenToAdd = DAI } = modal;
   const address = loginAccount.address;
```
JohnDanz: just destructure these

packages/augur-ui/src/modules/modal/add-funds.tsx
```diff
@@ -58,42 +51,41 @@ const addFundsTorus = async (amount, address) => {
...
     }
     closeModal();
   };
-    
 
-  const [amountToBuy, setAmountToBuy] = useState(createBigNumber(0));
-  const [isAmountValid, setIsAmountValid] = useState(false);
-  const [poolSelected, setPoolSelected] = useState(false);
-  const { env: config, ethToDaiRate, repToDaiRate } = useAppStatusStore();
   const ETH_RATE = createBigNumber(1).dividedBy(
```
JohnDanz: not sure why ETH_RATE/REP_RATE are defined here. just use the swap component below and don't pass props if we can just refer to the hook inside the leaf component (swap in this case)

packages/augur-ui/src/modules/modal/add-funds.tsx
```diff
@@ -213,27 +200,26 @@ export const AddFunds = ({
...
-                  toToken={tokenToAdd}
-                  fromToken={initialSwapToken ? initialSwapToken : null}
-                  ETH_RATE={ETH_RATE}
-                  REP_RATE={REP_RATE}
-                  config={config}
-                  useSigner={useSigner}
-                />
-              )}
+
+              <Swap
```
JohnDanz: Anything that is passed to this component but isn't used on otherwise should just be refactored into this component directly. we don't want to pass props unless they are actually needed to be passed like a marketID or something.

packages/augur-ui/src/modules/modal/add-funds.tsx
```diff
@@ -213,27 +200,26 @@ export const AddFunds = ({
...
+                ETH_RATE={ETH_RATE}
+                REP_RATE={REP_RATE}
+                config={config}
+                useSigner={useSigner}
+              />
             </>
           )}
 
           {selectedOption === ADD_FUNDS_CREDIT_CARD && (
             <CreditCard
```
JohnDanz, line 174: same comment as the swap above. AddFundsTorus and Fortmatic for example could just be defined inside of `creditCard` instead of here.

packages/augur-ui/src/modules/modal/components/modal-view.tsx
```diff
@@ -56,10 +56,10 @@ function selectModal(type, props, closeModal, modal) {
       return (
         <>
           {/* MOBILE */}
-          <ModalAddFunds />
+          <ModalAddFunds autoSelect={false} />
```
JohnDanz: autoselect should just be defaulted to false, then we don't need to specifiy.

packages/augur-ui/src/modules/modal/components/modal-view.tsx
```diff
@@ -56,10 +56,10 @@ function selectModal(type, props, closeModal, modal) {
...
 
           {/* DESKTOP */}
-          <ModalAddFunds {...props} autoSelect />
+          <ModalAddFunds autoSelect={true} />
```
JohnDanz: if a prop is a boolean, you don't need to explicitly pass `someProp={true}` you can simply just put in `someProp` and that will count as true.

packages/augur-ui/src/modules/swap/components/swap-row.styles.less
```diff
@@ -51,11 +51,11 @@
         }
 
         svg:last-of-type {
-          fill: var(--color-outcome-seven);
           width: @size-12;
 
           > g {
             stroke-width: 2;
+            fill: var(--color-outcome-seven);
```
JohnDanz: we really should look at this, as outcome seven is completely different colors based on the theme. if this is always supposed to be a specific color, we should just make an exception -- see `--color-facebook-1` kinda thing. we have hardcoded third party colors we need due to branding so we have an exception where it's the same no matter the theme

JohnDanz (changes requested):
couple of smaller changes requested.

JohnDanz (approved):

Issue #8399, opened 2020-07-09
Topbar total funds tooltip confusing


matt-bullock:
Showing DAI value for in ETH in signing wallet when user withdraws all and hasn't activated their account.
we are showing DAI value for ETH in signing wallet in the tooltip. 
but since the user hasn't activated their trading account we aren't showing fee reserve

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/a768ec26-7271-4fed-97c4-660512907120)

matt-bullock:
looks correct. closing

PR #8412, opened 2020-07-10
Fix tooltip non activated trading accounts

bthaile:
https://github.com/AugurProject/augur/issues/8399

No tooltip if user hasn't activated their trading account.
image

tooltip the same when account is activated.


bthaile:
@Mergifyio backport dev

matt-bullock (approved):

PR #8415, opened 2020-07-10
Fix filled orders hotloaded markets

bthaile:
hot loaded markets are in marketInfos state which causes filled orders to not find outcome and blow up. if can not find outcome don't load fill order, regular marketInfos will be loaded once sdk is synced"


bthaile:
@Mergifyio backport dev

matt-bullock (approved):

Issue #8111, opened 2020-06-22
wrong network bug


Chwy5:
IF on mainnet, i get a message that i am on the wrong network, if i switch and refresh it blows up, get a white screen says something went wrong

pgebheim:
That wrong network dialog seems to be coming from when the SDK fails to load. Is this still in progress @bconfortin or was this not reproducible?

Chwy5:
tested this...started on main, got message switched networks and hit refresh....wouldnt load for me

I needed to exit out and try again with the correct network set

Chwy5:
still an issue....If I am on the network I get the pop-up message telling me to switch (which I do).  I then hit the refresh button and the pg does not refresh but eventually blows up and see this....

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/c9bf5541-8c8f-4468-932c-0bead3ded0e6)

bthaile:
Not able to repro.
test steps. 
1. load UI when MM is pointed to mainnet
2. get network mismatch
3. change network to kovan

at one point I could repro a major lag, but not anymore. We'll keep trying.
4. refresh
UI loads normally.

pgebheim:
This happens for me too -- the JS on the page looks like it hangs indefinitely when MM changes the provider.

![image](https://user-images.githubusercontent.com/86010/87080734-08ee7580-c1dd-11ea-89e6-9431f24cf67a.png)

Then on refresh:
![image](https://user-images.githubusercontent.com/86010/87080821-2de2e880-c1dd-11ea-9758-0f284fef9ae1.png)

bthaile:
UI just spins. tested local against mainnet. prob should handle this case just in case.

![image](https://user-images.githubusercontent.com/3970376/87164935-0a787600-c28f-11ea-86b6-4fc33cf7f1e1.png)

Chwy5:
fixed - closing

PR #8428, opened 2020-07-10
Fix hanging refresh network mismatch

bthaile:
https://github.com/AugurProject/augur/issues/8111

Not super excited about this UX. On occassion the user has MM installed and init-augur can get chainId, some times chainId is undefined. To always get the chainId `ethereum.enable` is called which prompts the user to signin and allow Augur app. Then the chainId is available. This should only happen when MM is installed. 

Changed error modal displayed if SDK can't be created. the error modal was the network connect modal, which was confusing because there was nothing wrong with ethereum endpoint.

Use this MM documentation <https://docs.metamask.io/guide/ethereum-provider.html#using-the-provider>


bthaile:
@Mergifyio backport dev

andrewdamelio (approved):

PR #8434, opened 2020-07-13
Network mismatch modal is coming up for local chain when it doesn't n…

bthaile:
tell user they need to switch MM to local network. There was a bug for local, init-augur was showing network mismatch even though user had MM set to local and local node was running. 

Corrected:
image


andrewdamelio (approved):

bthaile:
@Mergifyio backport dev

Issue #8435, opened 2020-07-13
Remove Tie from NFL


Chwy5:
1) Remove Tie from NFL Outcome Tie/No Contest - Only use No Contest

2) Change Rule: Remove Tie from Tie/No contest in the 2nd and 3rd rule

3) Add Rule:

If the game ends in a tie, the market should resolve as "No Contest"

Chwy5:
fixed - closing

PR #8436, opened 2020-07-13
update NFL template to remove Tie and add resolution rule.

bthaile:
make NFL template consistent with spread and over/under

https://github.com/AugurProject/augur/issues/8435


Chwy5 (approved):

bthaile:
@Mergifyio backport dev

Issue #8343, opened 2020-07-07
Design QA: Global Chat


yanelbottini:
**Chat button**
Style and position is different from design

![global-chat-button.png](https://images.zenhubusercontent.com/5cc988ee77bc5b0505df9320/30266594-bb87-40de-b65a-b775634127ff)

**Figma reference**
https://www.figma.com/file/wogMuUDylgOHgTEYhZ3jgp/Augur-Bet-UI?node-id=3494%3A13

yanelbottini:
![Screen Shot 2020-07-20 at 08.10.36.png](https://images.zenhubusercontent.com/5cc988ee77bc5b0505df9320/7dd788f8-5317-4c92-8a3d-c41d6df50dc0)

- Button is still bigger than original design (Should be 36px height)
- Font size is different: should be 10px

PR #8443, opened 2020-07-13
Adjusted colors for Global Chat button

bconfortin:
#8343

### DIFF
--- a/packages/augur-ui/src/assets/styles/_colors.less
+++ b/packages/augur-ui/src/assets/styles/_colors.less
@@ -148,9 +148,7 @@
   --color-primary-text-40: fade(#0E0E21, 40%);
   --color-primary-text-50: fade(#0E0E21, 50%);
   --color-primary-text-80: fade(#0E0E21, 80%);
-  --color-primary-text-85: fade(#0E0E21, 85%);
   --color-primary-text-90: fade(#0E0E21, 90%);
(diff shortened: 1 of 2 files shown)


packages/augur-ui/src/assets/styles/_colors.less
```diff
@@ -148,7 +148,9 @@
   --color-primary-text-40: fade(#0E0E21, 40%);
   --color-primary-text-50: fade(#0E0E21, 50%);
   --color-primary-text-80: fade(#0E0E21, 80%);
+  --color-primary-text-85: fade(#0E0E21, 85%);
```
JohnDanz: you gotta define these for trading (aka just `:root`) as well otherwise this will cause an error when it can't find this variable name defined.

packages/augur-ui/src/modules/global-chat/components/global-chat.styles.less
```diff
@@ -98,6 +98,52 @@
...
+      }
+
+      > svg {
+        display: none;
+      }
+    }
+  }
+}
+
 .ThreeBoxChat {
```
JohnDanz: if we removed three box, lets remove the styling for it as well.

JohnDanz (changes requested):

JohnDanz (approved):

Issue #8440, opened 2020-07-13
V2 Bug blow up


Chwy5:
![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/da4ff286-3188-415e-b53c-395ae6daa034)

![image.png](https://images.zenhubusercontent.com/5c2f99aa80139b3816b48707/9fd5dd07-4503-4944-84b5-fdba183a24a9)

Chwy5:
not seeing this - working better

PR #8446, opened 2020-07-14
Fix hotloading market trading page

bthaile:
added more conditions to hotloading market. the conditional was not checking if canhotload is true. 
https://github.com/AugurProject/augur/issues/8442
https://github.com/AugurProject/augur/issues/8440


bthaile:
@Mergifyio backport dev

andrewdamelio (approved):

PR #8453, opened 2020-07-14
add pending to my bets, update data shown, fix bugs with pending queue and reporting from state refactor

phoebemirman:


packages/augur-ui/src/modules/common/buttons.tsx
```diff
@@ -670,7 +671,11 @@ export const CashoutButton = ({
   const { 
       accountPositions: positions,
       loginAccount: { address: account },
+      pendingQueue,
+      actions: { addPendingData }
   } = useAppStatusStore();
+  const queueId = bet.marketId+'_'+bet.orderId;
```
JohnDanz: use template string

JohnDanz (changes requested):
just a minor adjustment

JohnDanz (approved):

Issue #7954, opened 2020-06-03
Landing Page Banners


yanelbottini:
We have updated the landing page for both Betting Exchange and SportsBook to include multiple banners:

*Please take into consideration that banners are still in progress, but page layout won't change*

On desktop breakpoint, we have a main banner (That corresponds to the site the user is looking at and invites to Sign Up), and 2 side banners that link through our other apps. 

On Mobile, Tablet and Tiny Desktop, the 3 banners lives together on a carrousel: 

[Figma reference](https://www.figma.com/file/wogMuUDylgOHgTEYhZ3jgp/Augur-Bet-UI?node-id=2423%3A0)

matt-bullock:
@bconfortin The side banners looks good. The top banner is missing the gradient overlay and the dots graphic. Let me know if you need help with exporting this image. Font weight looks off.

Should also get the mobile carousel set up with all the images. We'll get you the side images at the right sizes to work in the carousel.

yanelbottini:
Hi @bconfortin, thanks for the updates. Here is the link where you can grab the final banner design for the TOP banner:

**https://www.figma.com/file/wogMuUDylgOHgTEYhZ3jgp/Augur-Bet-UI?node-id=7396%3A216185**

We need to buy the side banner photo. I'll let you know when we do it. 
For mobile, we plan to have a carousel set up with all the images. We'll get you the side images at the right sizes to work in the carousel.

yanelbottini:
@matt-bullock 
I've moved this ticket to design review cause I've already prepared the different banner variations for the different breakpoints. You can find them here:
https://www.figma.com/file/wogMuUDylgOHgTEYhZ3jgp/Augur-Bet-UI?node-id=7687%3A276585

We need to buy the iStockPhoto for Betting Exchange banner. I would like to have the photo to prepare the final assets and send this ticket back to Bruno. 

https://www.istockphoto.com/es/foto/aislado-hombre-de-jugador-de-f%C3%BAtbol-americano-gm868560768-145172673

WDYT?

yanelbottini:
@bconfortin 
I've updated the banners with the missing picture (The one with the football player) and I've also prepared the different variations according to the different breakpoints. 

You can check the banners here:
https://www.figma.com/file/wogMuUDylgOHgTEYhZ3jgp/Augur-Bet-UI?node-id=7789%3A278705

Chwy5:
Sent back...still needs to be corrected

PR #8485, opened 2020-07-15
Added new Hero Banner slider and side banners

bconfortin:
#7954


packages/augur-ui/src/modules/markets-list/components/slider.styles.less
```diff
@@ -0,0 +1,93 @@
...
+      }
+
+      &.RightAligned {
+        align-items: flex-end;
+
+        > h2 {
+          text-align: right;
+        }
+      }
+    }
```
bconfortin: These lines are repeated in `landing-hero.styles.less`, since we have a static banner and a slider/carousel.
Do you suggest anything here? Like adding these lines to a shared stylesheet?

Or do you want me to adjust the slider to convert from slider to static using something like `<Media query="TABLET_MAX">etc</Media>`?

packages/augur-ui/src/modules/trading/betslip.tsx
```diff
@@ -78,6 +80,14 @@ export const Betslip = () => {
...
         </section>
       </section>
+      <section className={Styles.SideImages}>
+        <a href="" target="_blank" rel="noopener noreferrer">
+          <img src={BannerSportsbook} alt="The image shows an American Football player running. The banner says '+1000 markets' and there is a button entitled 'Explore Sportsbook'."/>
+        </a>
+        <a href="" target="_blank" rel="noopener noreferrer">
+          <img src={BannerTrading} alt="The image shows Augur's Trading user interface and it says 'Trade Now!'."/>
+        </a>
+      </section>
```
JohnDanz, line 90: I don't think this will be the spot we put these. aka they don't belong to the `betslip` just because they are under it. However for now, this is fine, but it will definitely move.

packages/augur-ui/src/modules/markets-list/components/landing-hero.tsx
```diff
@@ -1,38 +1,89 @@
...
+    text: 'Sign up to start betting',
+    link: 'https://google.com/'
+  }
+}];
 
 export const LandingHero = () => {
   const {
     actions: { setModal },
   } = useAppStatusStore();
+  const { theme } = useAppStatusStore();
```
JohnDanz: put this above, we don't call `useAppStatusStore()` more than once.

packages/augur-ui/src/modules/markets-list/components/slider.styles.less
JohnDanz: I would prefer a non-media solution, like defining a class for that shared style on that div, and then simply importing the same stylesheet for both locations like we do in other areas, like create-market flow uses `styles` and `formStyles` for example on some components to be able to use classes from both.

JohnDanz (changes requested):

packages/augur-ui/src/modules/markets-list/components/landing-hero.tsx
bconfortin: Can't believe I didn't see that! I gotta get my eyes checked.

JohnDanz (approved):

PR #8500, opened 2020-07-16
Brings Migrate UI in line with new naming convention.

MicahZoltu:
Figured a PR was just as easy an issue suggesting the change.  While I disagree with the choice in naming of `REP` (instead of `REPv1`), I feel even stronger that we should be consistent with the chosen naming scheme throughout the Augur UI, and the ecosystem at large.

I suspect there are many other places in the UI that refer to `REP` which should instead refer to `REPv2`, but I'm not sure where they all live.  The Token used in Augur v2 is **NOT** `REP`, it is `REPv2`.  If we had decided to name `REPv1` then this would be less of an issue.  😉  As it is, if someone sees they need "REP" in order to dispute an incorrect outcome, they will go buy `REP` off an exchange and be sad when it isn't what they need to dispute.

### DIFF
--- a/packages/augur-ui/src/modules/modal/migrate-rep.tsx
+++ b/packages/augur-ui/src/modules/modal/migrate-rep.tsx
@@ -111,13 +111,13 @@ export const MigrateRep = () => {
       />
 
       <main>
-        {!inSigningWallet && <h1>You have V1 REP in your trading account</h1>}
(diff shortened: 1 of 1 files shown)


bthaile (approved):
lgtm

bthaile:
@Mergifyio backport v2

MicahZoltu:
Closing in favor of #8501

PR #8503, opened 2020-07-16
Cross orders cond order age

bthaile:
if size is the same, keep oldest order when comparing for cross orderbook.


bthaile:
@Mergifyio backport dev

Chwy5 (approved):

Issue #8525, opened 2020-07-17
Invalid Outcome Change


Chwy5:
For Market Trading pg continue to show Invalid Market as a % but take number from bid and not last trade price, 

No Bid = 0%

Chwy5:
added - closing

PR #8530, opened 2020-07-18
Track market invalid bids

bthaile:
https://github.com/AugurProject/augur/issues/8525

SDK portion of market invalid bids. 
created subscription to spit out Invalid bids to populate UI Invalid outcome in orderbook. sending all Bids so there is no management with regular orderbook data when user nav to market trading page.

`trackMarketInvalidBids` config, we might want to toggle it off for sportsbook.


packages/augur-sdk/src/api/MarketInvalidBids.ts
```diff
@@ -0,0 +1,44 @@
...
+    );
+  }
+
+  getMarketInvalidBids(orders: { logs: ParsedOrderEventLog[] }) {
+    const onlyInvalidBids = orders.logs.filter(
+      (o) => String(o.orderType) === OrderTypeHex.Bid && o.outcome === `0x00`
+    );
+    const markets = _.map(onlyInvalidBids, 'market');
+    const marketIds = Array.from(new Set(markets));
+    Promise.all(
```
pgebheim, line 27: I think promise.all actually had a limit to the number of promises it handles..

@justinbarry was this some platform specific thing?

pgebheim (approved):
Approving but we need to check the limits of Promise.all

bthaile:
@Mergifyio update

pgebheim:
@bthaile this needs to go into dev as well yes?

PR #8532, opened 2020-07-19
By default show a dash for invalid outcome percentage

bthaile:
Show `-` when there are no bids, this covers two scenarios:
1. orders have not been loaded
2. no bids exist for Invalid outcome

Screen Shot 2020-07-19 at 06 56 46


bthaile:
Not going to show dash if no bids. just show 0.00

Issue #8341, opened 2020-07-07
Design QA: Menu for users that are not logged in


yanelbottini:
**General**
- Please correct the order of the elements on the right-corner: Help / Odds Settings / Login / Sign Up.
- Double check font-size for buttons (Login / Signup): seem smaller than original design.
- Double check font-size for Trading / Betting Exchange / Sportsbook: seem bigger than original design. 
- “Betting UI - Coming Soon” should not appear on Sportsbook.  

![Screen Shot 2020-07-07 at 09.58.27.png](https://images.zenhubusercontent.com/5cc988ee77bc5b0505df9320/98a390ea-9153-4f8b-9331-78b55236e588)

**Help / Settings dropdown menu** 
- Please double check size of “Popular help resources” title: should be same height / style then “Odds settings” and “Time format”
- Please add tick to the selected option on Odds settings / Time format dropdown menu. 

![help-settings.png](https://images.zenhubusercontent.com/5cc988ee77bc5b0505df9320/3df6be95-5e4c-4964-a54c-9a45d7b750b7)

**Figma reference**
https://www.figma.com/file/42QoTLQcbObu7esoTPjWRv/Augur-Bet-Design-System?node-id=283%3A145

yanelbottini:
We still need to match Login / Signup font-size:

![Screen Shot 2020-07-20 at 08.05.13.png](https://images.zenhubusercontent.com/5cc988ee77bc5b0505df9320/e827f790-89cb-406a-809f-7cbeb27c8b73)
*dev on top; design on bottom*

yanelbottini:
<img width="220" alt="Screen Shot 2020-08-03 at 07 02 32" src="https://user-images.githubusercontent.com/49915900/89171528-b4c67f00-d557-11ea-8711-b93756198011.png">

Login font is smaller than Sign Up.

yanelbottini:
Done!

PR #8557, opened 2020-07-20
Betslip actions - hook up modal, have retry and correct data, and check mark

phoebemirman:
also this https://github.com/AugurProject/augur/issues/8341


packages/augur-ui/src/modules/app/components/app.tsx
```diff
@@ -14,7 +14,7 @@ import AlertsContainer from 'modules/alerts/components/alerts-view';
 import ToastsContainer from 'modules/alerts/components/toasts-view';
 
 import { Betslip } from 'modules/trading/betslip';
-import { BetslipProvider } from 'modules/trading/store/betslip';
+import { BetslipProvider, useBetslipStore } from 'modules/trading/store/betslip';
```
JohnDanz: remove the `useBetslipStore` here since we don't use it below.

packages/augur-ui/src/modules/market-cards/common.tsx
```diff
@@ -502,7 +509,8 @@ export const SportsOutcome = ({
     topLabel = determineTopLabel(market.sportsBook, outcomeId, outcomeLabel);
     label = OddToUse.full;
     disabled = false;
-    action = () =>
+    action = () => {
+      toggleHeader(BETSLIP_SELECTED.BETSLIP);
```
JohnDanz: I think we should just update the `addBet` reducer to always set the header to BETSLIP since I can't think of a situation where someone would be trying to `addBet` but we wouldn't want to show them the slip, so instead of triggering 2 state updates each time we can just modify `addBet`

packages/augur-ui/src/modules/market-cards/common.tsx
```diff
@@ -471,10 +472,16 @@ export const SportsOutcome = ({
...
 }: SportsOutcomeProps) => {
   const { liquidityPools } = useMarketsStore();
-  const { addBet } = Betslip.actions;
+  const { addBet, toggleHeader } = Betslip.actions;
   const poolId = market?.sportsBook?.liquidityPool;
-  const bestAsk =
-    poolId && liquidityPools[poolId] && liquidityPools[poolId][outcomeId];
+  let bestAsk = null;
+  if (poolId) {
+    Object.values(liquidityPools).map(pool => {
```
JohnDanz: we should avoid mapping here, since this could be a very large object and map will touch all of it when we could just reference directly. could switch to something like:

```
let bestAsk = null;
if (poolId && liquidityPools[poolId] && liquidityPools[poolId][outcomeId]) {
  bestAsk = liquidityPools[poolId][outcomeId];
}
```
this way we don't have to run through the object

packages/augur-ui/src/modules/modal/message-modals.tsx
```diff
@@ -604,3 +605,51 @@ export const ModalInitializeAccounts = () => {
...
+
+export const ModalCashoutBet = () => {
+  const {
+    modal,
+    actions: { closeModal },
+  } = useAppStatusStore();
+
+  return (
+    <Message
+      title={'Are you sure you want to cash out?'}
```
JohnDanz: `title="Are you sure you want to cash out?"` I've been changing these when I find them, I think we should just pass a plain old string if it's a hardcoded string.

packages/augur-ui/src/modules/modal/message-modals.tsx
```diff
@@ -604,3 +605,51 @@ export const ModalInitializeAccounts = () => {
     />
   );
 };
+
+export const ModalCashoutBet = () => {
+  const {
+    modal,
```
JohnDanz, line 611: lets destruct the `stake, cashOut, profit, and cb`

packages/augur-ui/src/modules/trading/betslip.tsx
```diff
@@ -19,8 +19,11 @@ import Styles from 'modules/trading/betslip.styles';
 import { PrimaryButton } from 'modules/common/buttons';
 
 export const Betslip = () => {
-  const [minimized, setMinimized] = useState(true);
-  const { theme } = useAppStatusStore();
+  const {
+    theme,
+    betslipMinimized,
+    actions: { setBetslipMinimized },
+  } = useAppStatusStore();
```
JohnDanz, line 26: on theme switch, if we are switching to `SPORTS` or `BETTING` we should make sure to reset `betslipMinimized` to minimized. This way when they switch it's always expected to start Minimized on tablet/mobile.

packages/augur-ui/src/utils/betslip-helpers.ts
```diff
@@ -6,6 +6,7 @@ import { Markets } from 'modules/markets/store/markets';
 import { placeTrade } from 'modules/contracts/actions/contractCalls';
 import { Betslip } from 'modules/trading/store/betslip';
 import { createBigNumber } from './create-big-number';
+import { formatDate } from './format-date';
```
JohnDanz: this also looks like it got added by vscode by accident, please remove.

JohnDanz (changes requested):
couple of smaller things, otherwise looks good to go

JohnDanz (approved):

Issue #8473, opened 2020-07-15
Ipad issue: unable to open my fills or positions on market page after submitting an order - tabs are unresponsive


matt-bullock:

matt-bullock:
I'm still seeing this. Have you been able to test on a tablet device?

matt-bullock:
works now

PR #8558, opened 2020-07-20
Bruno/8473

bconfortin:
#8473 

If there was no orders in the table and the user placed an order that fails, the table footer would overlap the table header and get in front of the tabs.

I've added a check to see if at least 1 order is not pending, meaning it's actually populating the table.

@mergifyio backport dev


bthaile (approved):
lgtm

PR #8580, opened 2020-07-21
Add check for existing betslip bet, fix wagers

phoebemirman:
also add error check and message for editing betslip wager and comparing it to best ask
also add updating best ask when having clicked a bid


packages/augur-ui/src/assets/styles/_colors.less
```diff
@@ -185,6 +185,7 @@
   --color-open-40: fade(#18C88D, 40%);
   --color-reporting: #F59300;
   --color-reporting-15: fade(#F59300, 15%);
+  --color-reporting-30: fade(#F59300, 30%);
```
JohnDanz, line 189: gotta add a version for the root as well, if it's undefined in trading it may cause issues. trading also uses a slightly different color for reporting so we should make sure it has a 30% fade as well for completeness sake.

packages/augur-ui/src/modules/trading/store/betslip-hooks.ts
```diff
@@ -248,7 +265,10 @@ export function BetslipReducer(state, action) {
     }
     case MODIFY_BET: {
       const { marketId, orderId, order } = action;
-      const toWin = getNewToWin(getOddsObject(order.normalizedPrice)[ODDS_TYPE.AMERICAN], order.wager);
+      const toWin = getNewToWin(
+        getOddsObject(order.normalizedPrice)[ODDS_TYPE.AMERICAN].value,
```
JohnDanz: We shouldn't assume American Odds and we also shouldn't ever be doing calculations in odds, that's only for the output and should be handled within the component, not saved to state. if they change odds, that won't update the betslip state and our display should always match their odds. It looks like we had this in place before but now that we know we need to adjust by odds for display we should update this. might get sticky if we start using odds in the background instead of sticking to what we do in trading

packages/augur-ui/src/utils/get-odds.ts
```diff
@@ -79,12 +79,20 @@ export const convertToNormalizedPrice = ({
   return normalizedPrice;
 };
 
+export const convertToWin = (normalizedPrice, quantity, toDecimals = 2) => {
```
JohnDanz: this whole section should be re-thought to not be using odds at all in any way to do calculations. this was a mistake early on made by me in the way I was setting these up. we should only use normalized prices and then output that normalized price to getOddsObject based on the current theme. we should avoid using odds objects to do math.

JohnDanz (changes requested):

JohnDanz (approved):

Issue #8468, opened 2020-07-15
Ipad: resolution details aren't showing on initial load


matt-bullock:
Only start showing after clicking the arrow 

![image.png](https://images.zenhubusercontent.com/5c07af47035895308911c4c5/24c64828-68ce-4ffe-a52c-bfe41d021fb5)

matt-bullock:
Needs the '...' before being cut off. This isn't loading on initial load

I'd prefer a read more button at the end of the text instead of the ... and down arrow as this would be clearer. Hopefully implement for v2.1

PR #8581, opened 2020-07-21
Bruno/8468

bconfortin:
#8468 

Issue was only happening on an iPad device, not the iPad resolution

@Mergifyio backport dev

Screenshot at Jul 21 17-32-46


bthaile (approved):
good one, lgtm

Issue #8564, opened 2020-07-21
Gas Detection issue when first coming to site


Chwy5:

Chwy5:
fixed -closing

PR #8585, opened 2020-07-21
Fix estimate dispute fallback

bthaile:
https://github.com/AugurProject/augur/issues/8564

use no-show bond as estimate value to get gas estimate
@Mergifyio backport dev


nuevoalex (approved):
This will still fail if the remaining amount in the bond would be less than the market creation bond. Instead this value should be the remaining bond to fill amount or if the user does not have that much REP 1 attoREP.

PR #8588, opened 2020-07-22
fix metamask local

robert-e-davidson3:
@Mergifyio backport dev


packages/augur-ui/src/modules/app/actions/init-augur.ts
```diff
@@ -105,7 +105,7 @@ const isNetworkMismatch = async (config, dispatch): boolean => {
   const web3NetworkId = String(createBigNumber(chainId));
   const privateNetwork = isPrivateNetwork(config.networkId);
   const isMisMatched  = privateNetwork ?
-    web3NetworkId !== "NaN" : // MM returns NaN for local networks
+    !(isPrivateNetwork(web3NetworkId) || web3NetworkId === 'NaN') : // MM can return NaN for local networks
```
robert-e-davidson3, line 108: Metamask seems to have fixed the problem where they'd return NaN for local networks. But I kept the NaN check in case we get people with old metamask versions. I'm not sure how realistic that is though.

bthaile (approved):
Lgtm

PR #8590, opened 2020-07-22
Fix Icons

JohnDanz:
corrected issue with attributes on svgs that are not parseable by JSX because they included a -.


bthaile (approved):
lgtm

Issue #8589, opened 2020-07-22
Ipad Issue: Not seeing my open orders in the open orders section of trading page


matt-bullock:
Refreshing isn't fixing it

![image](https://user-images.githubusercontent.com/39496225/88172202-e1f26380-cc20-11ea-8503-2d7bf73635ad.png)

matt-bullock:
fixed. closing

PR #8595, opened 2020-07-23
Switched from 0.5fr to 50vh and removed a useless filter

bconfortin:
#8589 

Safari was not showing orders because they implement Flexbox in a weird way.
That fix also fixes the footer of the table getting in front of the header, so I reverted that openOrders.filter because it could potentially cause perfomance issues.


bconfortin:
@mergifyio backport dev

bthaile (approved):
lgtm

PR #8635, opened 2020-07-26
fixed 2 issues, first reference to util function had wrong reference.…

bthaile:
… seconds fixed percentage calculation that only effected scalars

@Mergifyio backport dev


andrewdamelio (approved):
Lgtm

PR #8636, opened 2020-07-26
if scalar has more tickSize decimals than yesno/cat then ellipse it

bthaile:
ellipse prices that have more tickSize decimals

image

Other issue
image

fixed
image


andrewdamelio (approved):

bthaile:
@Mergifyio backport dev

PR #8646, opened 2020-07-27
run augurSdk.connect() before listening for events

robert-e-davidson3:


robert-e-davidson3:
@Mergifyio backport dev

robert-e-davidson3:
I tested this. It works.

pgebheim (approved):

pgebheim (changes requested):
This is wrong -- we want the ui to listen to the events before doing the connect.

That is what triggers the various startup events.

robert-e-davidson3:
> This is wrong -- we want the ui to listen to the events before doing the connect.
> 
> That is what triggers the various startup events.

to clarify: this works in the websocket connector mode but breaks the normal UI mode

robert-e-davidson3:
Actually that does not seem to be a problem but since this is a high risk change, we're going to sit on it until after launch.

pgebheim (approved):

PR #8720, opened 2020-07-29
Calc activating fee reserve

### LINKED ISSUE #8708: ETH Reserve being used when DAI balance in Trading Wallet doesn't have sufficient DAI to fund it
The client attempts to create an ETH reserve of .16 ETH (worth ~$50) present) if the trading wallet balance is above 40 DAI.

The Transaction itself costs between 15-30 DAI.

This means if the trading wallet contains an amount between roughly 40-80 it is likely to fail because the trading wallet will not have sufficient DAI in order to complete the transaction.

FIX: The client should globally disable trying to create an ETH reserve when the trading wallet balance is less than .16ETH worth of DAI + some excess buffer amount that represents a maximum expected TX fee (for now lets assume 6mil gas). This would disable the ETH reserve for trading wallet balances less than ~$100 currently.

Algo:

```
const desired_eth_reserve = 0.16 - signing_wallet_eth_balance;
const desired_eth_reserve_in_dai = to_dai(eth_reserve);
const total_estimated_max_ = 6000000 * gasPrice; // Estimate of a very large tx with current gas price
const total_dai_worst_case = total_estimated_max + desired_eth_reserve_in_dai ;
const use_eth_reserve = trading_wallet_dai_balance > total_dai_worst_case;
```

bthaile:
calc activating fee reserve based on user dai balance after a worst case tx. if they are above min dai balance keep fee reserve activated.

https://github.com/AugurProject/augur/issues/8708

@Mergifyio backport dev


packages/augur-ui/src/modules/auth/actions/update-assets.ts
```diff
@@ -72,8 +74,29 @@ export const updateAssets = (): ThunkAction<any, any, any, any> => async (
...
+    const desiredEthBalance = feeReserveAmount.minus(signerEthBalance);
+    let activateFeeReserve = true;
+    if (desiredEthBalance.gt(ZERO)) {
+      // gasPrice
+      const gasPriceGwei = gasPriceInfo.userDefinedGasPrice || gasPriceInfo.average
+      const gasPriceWei = createBigNumber(gasPriceGwei).times(createBigNumber(GWEI_CONVERSION));
+
+      // convert worst cost tx in dai
+      const dai2EthConvert = createBigNumber(dai2Eth.value);
+      const desiredEthReserveInDai = feeReserveAmount.times(dai2EthConvert);
```
nuevoalex: `feeReserveAmount` here should be `desiredEthBalance `

bthaile: made change

nuevoalex (approved):

PR #8730, opened 2020-07-29
disable account activation

bthaile:
disable account activation button so user doesn't accidentally click it again. 
dismiss transfer modal so user doesn't get confused when they transfer from onboarding


packages/augur-ui/src/modules/common/buttons.tsx
```diff
@@ -208,7 +208,7 @@ const ProcessingButtonComponent = (props: DefaultButtonProps) => {
           icon={icon}
           text={buttonText}
           action={buttonAction}
-          disabled={isDisabled}
```
justinbarry, line 211: @bthaile Above we have `let isDisabled = props.disabled;`

justinbarry (approved):
One comment. Otherwise lgtm

PR #8752, opened 2020-07-30
check users eth balance based on gas cost of transaction

bthaile:
check user's eth balance when buying participation tokens


pgebheim:
Cherry-picked this into release

PR #8756, opened 2020-07-30
uniswap fixes

### LINKED ISSUE #8753: Convert REP -> DAI is Inverted
image

@andrewdamelio -- Uniswap pairs are ordered lexicographical by the token address -- so with the mainnet REPv2 token its in the opposite order as the testnet deploys.

We need to swap this rate. Also, should update the copy to say REPv2 and use the new icon.

andrewdamelio:
https://github.com/AugurProject/augur/issues/8753


bthaile (approved):
lgtm

PR #8788, opened 2020-07-31
add client side redirect from http to https when not in dev mode

bthaile:
add client side redirect from http to https when not in dev mode


packages/augur-ui/src/index.ejs
```diff
@@ -107,6 +107,12 @@
 
   <!-- Google Translate -->
   <script type="text/javascript">
+    if ((window.location.protocol === "http:" &&
+      (window.location.host !== "localhost" && window.location.host !== "127.0.0.1") &&
```
justinbarry:
```suggestion
      (window.location.hostname !== "localhost" && window.location.hostname !== "127.0.0.1") &&
```

See: https://stackoverflow.com/questions/6725890/location-host-vs-location-hostname-and-cross-browser-compatibility

justinbarry (approved):
LGTM

packages/augur-ui/src/index.ejs
MicahZoltu: Ironically, while @justinbarry was correct in this feedback, if @bthaile's original code was used then local IPFS browsing wouldn't have broken (since it uses port 8080).  😄  (see #8808)

PR #8805, opened 2020-08-01
Turn off timing out failing queue

bthaile:
turn off timing out processing buttons/orders/queued tx status.


pgebheim:
Cherry-picked into v2

PR #8806, opened 2020-08-01
Adding template for reporting TVL using DeFi pulse.

jfarid27:
## Description

Adds a template for TVL using DefiPulse as a resolution source.

## TODO

- [ ] Confirm if template is unambiguous.
- [ ] Confirm if other resolution sources should exist.
- [ ] Confirm working state.


packages/augur-tools/src/templates-source.ts
```diff
@@ -9772,6 +9773,39 @@ export const TEMPLATES = {
...
+                type: TemplateInputType.DATEYEAR,
+                daysAfterDateStart: 1,
+                placeholder: `Day of Year`,
+              },
+            ],
+            resolutionRules: {
+              [REQUIRED]: [
+                {
+                  text: `This market shall resolve based on the Total Value Locked (USD) in DeFi reported by defipulse.com at the specified time. If data is unavailable at defipulse.com, the market shall resolve according to the DeFi pulse API.`,
+                  text: `To report, choose "30D" in the Total Value Locked (USD) for ALL DeFi. Hover over points until you reach the specified date and report the TVL (USD) in Billions (B) with 3 decimal points. If more are available, truncate the third decimal point following rounding rules: if the fourth decimal is 5 or greater, round up. Otherwise remain the same.`,
```
MicahZoltu: These instructions can't be used for reporting since they cannot be followed 30 days or more after the market enters reporting.  Consider making the API the primary source (assuming it goes back unboundedly) and having the UI mechanism as an "easy alternative if available but not authoritative".

packages/augur-tools/src/templates-source.ts
```diff
@@ -9772,6 +9773,39 @@ export const TEMPLATES = {
...
+                },
+                {
+                  text: `If using the DefiPulse public API found at https://public.defipulse.com/api/GetHistory?period=1m&api-key=<API_KEY> and report the value shown for the required date.`,
```
MicahZoltu: Are API keys free?  If you have to pay or KYC to get them then I don't think this can reasonably used as a reporting source.  May be worth mentioning how to get an API_KEY (e.g., a helpful but non-authoritative link).

packages/augur-tools/src/templates-source.ts
```diff
@@ -9772,6 +9773,39 @@ export const TEMPLATES = {
...
+                  text: `The date will be in Unix timestamp in UTC for 00:00 (12:00 AM). You must convert the timestamp to a date and report the matching date.`,
+                },
+                {
+                  text: `If DefiPulse is not available before the end of the reporting period for the given date, the market should resolve as "Invalid".`,
```
MicahZoltu: This is unenforceable and should be removed.  Markets resolve as invalid if the data is not available at reporting start time.  You cannot override that via a rule in the market.

packages/augur-tools/src/templates-source.ts
MicahZoltu: This API call looks like it only gets recent history.  For reporting we need a source that can be accessed (with reliable results) many months after reporting starts.

packages/augur-sdk-lite/src/templates/constants.ts
```diff
@@ -57,6 +57,7 @@ export const INDEXES = 'Indexes';
 export const BITCOIN = 'Bitcoin';
 export const ETHEREUM = 'Ethereum';
 export const LITECOIN = 'Litecoin';
+export const DEFI = 'DeFi';
```
MicahZoltu, line 60: I don't think this is deserving of its own category, recommend putting it in CRYPTO.  If we end up with so many markets like this that they deserve their own category we can create a new category at that time and move them all.

MicahZoltu: (FWIW, I think a lot of the categories shouldn't be categories)

jfarid27: This actually is in the "CRYPTO" category in the templates at the moment. It looks like CRYPTO has children already. Not sure how to stick the template in there when the second level is groupings anyway. Can you link an example?

packages/augur-tools/src/templates-source.ts
jfarid27: Got it. Will update to remove this.

packages/augur-tools/src/templates-source.ts
jfarid27: So the API allows up to a year and probably could be more. I imagine within a year they'll beef this up too. I added some information about what could possibly be called as long as it's grouped by date, and included a link to the documentation.

jfarid27: Also, API keys are free, but I've also removed that. This endpoint is public so you don't actually need an API key.

packages/augur-tools/src/templates-source.ts
jfarid27: Done.

packages/augur-tools/src/templates-source.ts
```diff
@@ -9772,6 +9773,43 @@ export const TEMPLATES = {
           },
         ],
       },
+      [DEFI]: {
+        templates: [
+          {
+            marketType: SCALAR,
+            question: `What will the Total Value Locked "USD" in Defi be on [0] (UTC), according to https://data.defipulse.com/ ?`,
+            example: `What will the Total Value Locked "USD" in Defi be on 03/27/2022 (UTC), according to https://data.defipulse.com/ ?`,
+            denomination: 'Price',
```
MicahZoltu: For scalar markets, shouldn't this be "Total Value Locked"?

jfarid27: Done!

bthaile:
closing PR until process is identified.

PR #8813, opened 2020-08-02
set template filter to all, add custom market label to market cards

### LINKED ISSUE #8812: Market Card and Market Filter Change
1) Add this to custom market cards right above title 

image

2) Make Markets Filter set to All

bthaile:
https://github.com/AugurProject/augur/issues/8812

1. set default template filter to 'all'
2. put custom market label on market cards.


bthaile:
@Mergifyio backport dev

PR #8815, opened 2020-08-02
Change text color and copy of Invalid Outcome

pgebheim:


bthaile (approved):
lgtm

bthaile (changes requested):
not sure we want to change the color of Invalid outcome

bthaile (approved):
lgtm

Issue #8833, opened 2020-08-03
Display affiliate fee


matt-bullock:
https://www.figma.com/file/fLWVwmanAwetVZbujQquEi/Market-Page?node-id=3037%3A0

![image](https://user-images.githubusercontent.com/39496225/89206397-8d080300-d5b9-11ea-94b9-81ebbb9aa984.png)

![image](https://user-images.githubusercontent.com/39496225/89206717-f9830200-d5b9-11ea-9133-f2ed4909c5c2.png)

Chwy5:
added and fixed closing

PR #8858, opened 2020-08-04
Fix reporting form affiliate fee blowup

bthaile:
https://github.com/AugurProject/augur/issues/8833

missed core properties in reporting form.


pgebheim (approved):

packages/augur-ui/src/modules/market/components/core-properties/core-properties.tsx
```diff
@@ -182,6 +161,27 @@ const CoreProperties: React.FC<CorePropertiesProps> = ({
               }
             </>
           )}
+          {loadAffiliateFee && <PropertyLabel
```
robert-e-davidson3, line 164: Won't this insert the falsy `loadAffiliateFee` into the children of the parent dom node? Like, wouldn't that become an errant `0` or `undefined` or something?

bthaile: I'll take any false
undefined | 0 | false

as long as Function returns true.

Issue #8825, opened 2020-08-03
Firefox displaying network mismatch even though MM is connected to Mainnet


Chwy5:
@MicahZoltu  reports Firefox Issue:

Latest UI in Firefox keeps telling me there is a network mismatch yet MetaMask is connected to mainnet.
MetaMask is connected to the wrong Ethereum network. Please set the MetaMask network to: undefined.
In the console I see:
Uncaught (in promise) Error: API must be initialized before use.
I also see:
web3 updated, network changed to 1
The network mismatch modal has no way to close it, I can only refresh.
No markets every load in Firefox w/ MetaMask either when loading from insecure context (localhost subdomain).
I don't see any failed requests in the network tab.
I see a Torus error in console.
I see some miscellaneous warnings in console but nothing that looks critical.
I see one error in console which doesn't sound critical but maybe is causing the initialization to generally fail.
The Notification permission may only be requested in a secure context.
I do see this warning which maybe is related:
asm.js type error: expecting argument type declaration for 'e' of the form 'arg = arg|0' or 'arg = +arg' or 'arg = fround(arg)'

MicahZoltu:
Two other users in #general have reported that the app doesn't work in Firefox.  It is unknown if they are using localhost IPFS or hosted IPFS.

MicahZoltu:
Firefox connected via CloudFlare gateway:
* When I connect, complains about wrong MM network despite being connected to mainnet.
* Error in console: `Uncaught TypeError: o is undefined`
* Error in console: `Uncaught TypeError: getStore().commit is not a function`
* Error in console: `Uncaught (in promise) Error: API must be initialized before use.`

Chwy5:
Not seeing this in testing now

PR #8864, opened 2020-08-04
Dont keep processing init augur web3 wrong network

bthaile:
one Issue with init-augur when MM is on the wrong network. the client is tried to be created and won't work. need to just return callback(null) since Network Mismatch modal is already up. 

with this fix, correct modal stays up.

image

https://github.com/AugurProject/augur/issues/8825


bthaile:
@Mergifyio backport dev

pgebheim (approved):
I do like this!

Issue #8657, opened 2020-07-27
iPad issue: market preview getting cut off at top (hides close preview button) and buttons are visually broken


matt-bullock:
In the image below I'm having to drag down to reveal the button

![image](https://user-images.githubusercontent.com/39496225/88549505-a7be0300-d020-11ea-923c-38881e995608.png)

PR #8878, opened 2020-08-04
Changed from flex to grid because of weirdness happening in iPad

bconfortin:
#8657 

Captura de Tela 2020-08-04 às 12 04 47
Captura de Tela 2020-08-04 às 12 04 55


bthaile (approved):
lgtm

PR #8893, opened 2020-08-05
address pending order persistence issues

### LINKED ISSUE #8840: Trading issues
- [x] open order not disappearing from order book when a user gets filled while offline / has tab closed
- [x] open order not disappearing from open orders if a user exits after getting bell alert filled but before UI removes the open order
- [x] self trade fill not working right in UI, UI says failed but really succeeded

nuevoalex:
@Mergifyio backport dev

This addresses the issues in #8840 

I did notice a react error when pending orders are removed but it doesn't seem to actually break anything.


bthaile (approved):
Lgtm

Issue #8843, opened 2020-08-03
Use max(eth_gasPrice, recommended) Gas when submitting transactions


joeykrug:
- [x] should use max(eth_gasPrice, recommended) in UI as default vs solely one from provider otherwise will be too slow

Split this into two parts
#8841 - Upon first trade there should be a popup modal that says if you want to set your gas price to something different than the default, you should do so in account settings at the top right if you want to (with an image of it) and *not* to do so in your wallet otherwise it’ll cause your trade to fail.

Chwy5:
@matt-bullock need modal

pgebheim:
This is merged into 2.1.9

PR #8895, opened 2020-08-05
EthersProvider queries gas station for gas price

robert-e-davidson3:
#8843


packages/augur-contract-dependencies-ethers/src/ContractDependenciesEthers.ts
```diff
@@ -268,18 +272,14 @@ export class ContractDependenciesEthers implements Dependencies<BigNumber> {
     tx: Transaction<ethers.utils.BigNumber>,
     txMetadata: TransactionMetadata
   ): Promise<ethers.providers.TransactionReceipt> {
-    const gasLimit = tx.gasLimit || (await this.provider.estimateGas(tx));
-    const gasPrice = await this.provider.getGasPrice();
-    tx.gasPrice = gasPrice;
+    tx.gasLimit = tx.gasLimit || await this.provider.estimateGas(tx);
+    tx.gasPrice = tx.gasPrice || await this.provider.getGasPrice();
```
robert-e-davidson3, line 276: This behavior is new and might(?) break something: we can now specify gas price per-tx and it will pass through to the signer. The signer can then do whatever it wants of course. But if this does break anything then it's likely due to revealing a hidden bug.

packages/ethersjs-provider/src/EthersProvider.ts
```diff
@@ -228,13 +230,30 @@ export class EthersProvider extends ethers.providers.BaseProvider
     return super.getBlockNumber();
   }
 
-  async getGasPrice(): Promise<ethers.utils.BigNumber> {
+  async getGasPrice(networkId?: NetworkId): Promise<ethers.utils.BigNumber> {
```
robert-e-davidson3, line 233: heart of the new logic

packages/augur-contract-dependencies-ethers/src/ContractDependenciesEthers.ts
nuevoalex: Yea this should be reasonable and if not something else was wrong ;p

nuevoalex (approved):
👍

Issue #8901, opened 2020-08-05
Design QA - Hover Categories


yanelbottini:
Please, add hover style for categories:

<img width="298" alt="Screen Shot 2020-08-05 at 06 38 49" src="https://user-images.githubusercontent.com/49915900/89397641-79a68600-d6e6-11ea-967d-7d17d4f2cf81.png">

**Figma Reference:**
https://www.figma.com/file/wogMuUDylgOHgTEYhZ3jgp/Augur-Bet-UI?node-id=7733%3A13

yanelbottini:
Done!

PR #8908, opened 2020-08-05
update-category-hover

JohnDanz:
https://github.com/AugurProject/augur/issues/8901

updated hover behavior for the category rows in sportsbook to match the mocks.

### DIFF
--- a/packages/augur-ui/src/modules/common/form.styles.less
+++ b/packages/augur-ui/src/modules/common/form.styles.less
@@ -2462,6 +2462,18 @@ div.error > div > div > div > div > input {
         border: @size-1 solid var(--color-primary-text);
       }
     }
+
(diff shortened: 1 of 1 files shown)


packages/augur-ui/src/modules/common/form.styles.less
```diff
@@ -2462,6 +2462,18 @@ div.error > div > div > div > div > input {
         border: @size-1 solid var(--color-primary-text);
       }
     }
+
+    &:hover:not(.loading) {
```
bconfortin, line 2466: Consider adding `cursor: pointer` to that hover. The cursor only changes when there is something in the category. When it's 0, even though it's clickable, the cursor visual is not the clickable one.

bconfortin (approved):

Issue #8844, opened 2020-08-03
Approvals fixes


joeykrug:
- [x] The 4 approvals should note that they’re “Before your first trade only” then you’re done with approvals and you don’t have to do before every trade
- [x] It makes people do all approvals if they've already done n successfully already sometimes, does it lag to get the data that it was indeed confirmed? Need to fix this somehow so people don't do like 6-7 approvals

matt-bullock:
Suggestions are:

**Approval required to trade**
Your first trade requires 4 signings to approve using your funds. Once confirmed you can place your order.

**Approval required to add initial orders**
Your first order requires 4 signings to approve using your funds. Once confirmed you can place your order.

**Approval required to create market**
Your first market requires 1 signing to approve using your funds. Once confirmed you can create your market.

Can remove the header if we feel it's unnecessary 
![image](https://user-images.githubusercontent.com/39496225/89318869-abd0cd00-d67f-11ea-9a49-2187e95b43ba.png)

Try to maintain 8px padding between elements. Current version looks bunched up
![image](https://user-images.githubusercontent.com/39496225/89319261-28fc4200-d680-11ea-8cb4-a98ec1273318.png)

**Should include processing and confirmed states**

![image](https://user-images.githubusercontent.com/39496225/89321247-f869d780-d682-11ea-9dbb-138268f792fb.png)

Chwy5:
![image](https://user-images.githubusercontent.com/41301809/89644646-59adc880-d886-11ea-9476-4b3509fd25ba.png)

1) This box is showing in red - s/b white
2) Fix spacing....should not be on 4 lines

bthaile:
It shows red when user doesnt have enough ETH for gas

Chwy5:
shrunkin up - closing

PR #8922, opened 2020-08-06
fixed disabling and showing process... on the button so users won't m…

bthaile:
https://github.com/AugurProject/augur/issues/8844

better processing button mgmt so user won't have an opp to mash it too many times. 

image


andrewdamelio (approved):
LGTM

Issue #8884, opened 2020-08-04
Trader should be able to understand Invalid outcome tooltip


pgebheim:
To make this more understandable: Use the normal tooltip with the once piece of information trades care about and tell them to click for more.

1. tooltip: 
"Invalid outcome pays $1.00 per share if the market resolves as Invalid. Click here to learn what makes a market Invalid"

use market's max price value instead of 1.00
for scalars use `resolves at xxx denomination`

2. Invalid rules put Augur Markets on top of invalid rules. 
3. change Augur Markets in Invalid Outcome text, change
to "Invalid outcome pays $1.00 per share for Yes/No and Categorical markets if the market resolves as Invalid. Scalar markets pay out the upper bound. A lower price indicates a lower probability of the market resolving as invalid."

Chwy5:
If we chagne the tooltip how is anyone supposed to know what makes up invalid.....I wouldnt change this...

Issue #8882, opened 2020-08-04
Remove 24hr Volume from market detail page


pgebheim:
![image](https://user-images.githubusercontent.com/86010/89317713-a1d4ad00-d632-11ea-83f9-b9790d6ac92e.png)

Chwy5:
removed - closing

PR #8932, opened 2020-08-06
Misc remove 24hr vol fixes

bthaile:
https://github.com/AugurProject/augur/issues/8884
https://github.com/AugurProject/augur/issues/8882

remove 24hr volume on trading page
update Invalid outcome tooltip to be trader minded.


pgebheim:
Leaving this open because I wanna make sure the text makes sense to chwy and the tooltip right now doesn't mention that you can find our about the invalid rules.

bthaile:
Closing until we can vet tickets

PR #8968, opened 2020-08-10
Readded 3box chat and comments to the UI

bconfortin:


packages/augur-ui/src/modules/global-chat/components/global-chat.tsx
```diff
@@ -1,19 +1,29 @@
...
+
+const ThreeBoxChat = lazy(() =>
+  import(/* webpackChunkName: '3box-chat' */ 'modules/global-chat/components/three-box-chat')
+);
 
 export const GlobalChat = () => {
-  let { env } = useAppStatusStore();
+  let { isLogged, loginAccount, env, initialized3box } = useAppStatusStore();
   const [show, setShow] = useState(false);
+  const signer = loginAccount.meta?.signer;
```
JohnDanz: for this we can actually just pull this out from the context so above `{ loginAccount: { meta: { signer } }, isLogged, etc...`

this way signer will just be the signer or it's undefined which satisfies the provider assignment below. since we never use loginAccount or Meta we should just pull directly to the important part.

packages/augur-ui/src/modules/market/components/common/comments/market-comments.tsx
```diff
@@ -1,22 +1,39 @@
...
 
 export const MarketComments = ({
   marketId,
   colorScheme = 'dark',
   numPosts = DEFAULT_NUM_POSTS,
 }) => {
-  let { isLogged, env } = useAppStatusStore();
+  let { isLogged, env, loginAccount } = useAppStatusStore();
+  const threeBoxAdminAccount = '0x913dA4198E6bE1D5f5E4a40D0667f70C0B5430Eb';
+  const signer = loginAccount.meta?.signer;
```
JohnDanz: same as the other comment

packages/augur-ui/src/modules/market/components/common/comments/three-box-comments.tsx
```diff
@@ -0,0 +1,52 @@
...
+  marketId: string;
+}
+
+const ThreeBoxComments = ({
+  adminEthAddr,
+  provider,
+  marketId,
+}: ThreeBoxCommentsProps) => {
+  const { initialized3box } = useAppStatusStore();
+  const { setInitialized3Box } = useAppStatus();
```
JohnDanz: this is not how we use useAppStatus. we should never use that outside of the appStatus component. instead we want to modify the line above to say `const { initialized3box, actions: { setInitialized3Box } } = useAppStatusStore();`

packages/augur-ui/src/modules/modal/components/modal-global-chat.tsx
```diff
@@ -0,0 +1,57 @@
...
+  provider: any;
+  whichChatPlugin: string;
+  isLogged: boolean;
+}
+
+export const ModalGlobalChat = () => {
+  let { isLogged, loginAccount, env, initialized3box, actions: { closeModal } } = useAppStatusStore();
+  const signer = loginAccount.meta?.signer;
+  const provider = signer ? signer.provider?._web3Provider : false;
+  let { setInitialized3Box } = useAppStatus();
```
JohnDanz: same thing as the previous file, never use `useAppStatus` outside of the time it's defined in appStatus component provider. add `setInitilized3Box` as part of the things imported in the `actions` section of `useAppStatusStore()`

packages/augur-ui/src/modules/modal/components/modal-global-chat.tsx
```diff
@@ -0,0 +1,57 @@
...
+export interface ModalGlobalChatProps {
+  closeModal: Function;
+  provider: any;
+  whichChatPlugin: string;
+  isLogged: boolean;
+}
+
+export const ModalGlobalChat = () => {
+  let { isLogged, loginAccount, env, initialized3box, actions: { closeModal } } = useAppStatusStore();
+  const signer = loginAccount.meta?.signer;
```
JohnDanz: same as the other comments, just destructure loginAccount in the import.

packages/augur-ui/src/modules/modal/components/modal-global-chat.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export interface ModalGlobalChatProps {
+  closeModal: Function;
+  provider: any;
+  whichChatPlugin: string;
+  isLogged: boolean;
+}
+
+export const ModalGlobalChat = () => {
+  let { isLogged, loginAccount, env, initialized3box, actions: { closeModal } } = useAppStatusStore();
```
JohnDanz: never use let. should be a const.

packages/augur-ui/src/modules/modal/components/modal-global-chat.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const signer = loginAccount.meta?.signer;
+  const provider = signer ? signer.provider?._web3Provider : false;
+  let { setInitialized3Box } = useAppStatus();
+
+  const whichChatPlugin = env.plugins?.chat;
```
JohnDanz: we should just destrcuture this on the useAppStatusStore() call

JohnDanz (changes requested):

packages/augur-ui/src/modules/market/components/common/comments/three-box-comments.tsx
bconfortin: Ohhhh. Now it makes sense xD

JohnDanz (approved):

PR #8975, opened 2020-08-10
ERC20 Proxies of ERC1155 Contract

### LINKED ISSUE #8982: ERC20 Proxy of ERC1155
Our share tokens are ERC1155. This is great except when we need an ERC20 contract.

Writing a set of contracts that lets us have erc20 contracts - each with their own address - that correspond to erc1155 tokens (ShareToken tokens).

My approach is to have a single "nexus" contract that deploys erc20 contracts that are 1) cloned proxies of each other, and 2) proxy all their token state to the nexus or target erc1155 contract. The only proxied information in the nexus is approvals since the nexus assumes the user gave them all the necessary approvals.

- [x]  Write contracts, compiled and passing Alex's review.
- [x]  Get python test infra to deploy the contracts.
- [x]  Write tests for the contracts.
- [x] Update artifacts to include nexus address.
- [x] Update deploy script and constants files with nexus and proxy.

robert-e-davidson3:


packages/augur-core/src/contracts/trading/ERC20Proxy.sol
```diff
@@ -0,0 +1,126 @@
...
+    // So we can trust its _caller to be the proxy's msg.sender.
+    function _gate(ERC20Proxy _proxy, uin256 _tokenId) internal {
+        require(msg.caller == _proxy); // only proxies may make calls
+        require(proxies[_tokenId] == _proxy); // no sneaky stuff!
+    }
+}
+
+contract ERC20Proxy is IERC20 {
+    uint8 constant public decimals = 18;
+    address nexus;
```
nuevoalex: This could just be type ERC20ProxyNexus and I think it'd be nicer throughout.

packages/augur-core/src/contracts/trading/ERC20Proxy.sol
```diff
@@ -0,0 +1,126 @@
...
+    // Guarantees that the calling function was invoked by a trusted proxy contract.
+    // So we can trust its _caller to be the proxy's msg.sender.
+    function _gate(ERC20Proxy _proxy, uin256 _tokenId) internal {
+        require(msg.caller == _proxy); // only proxies may make calls
+        require(proxies[_tokenId] == _proxy); // no sneaky stuff!
+    }
+}
+
+contract ERC20Proxy is IERC20 {
+    uint8 constant public decimals = 18;
```
nuevoalex: In general we just use uint256 for all integer types. Easier to audit / some auditors have recommended this practice.

packages/augur-core/src/contracts/trading/ERC20Proxy.sol
```diff
@@ -0,0 +1,126 @@
...
+    // The spender is whoever called the proxy contract, which is msg.sender in the proxy and _caller here in the nexus.
+    mapping (uint256 => mapping (address => mapping (address => uint256))) public allowances;
+    // tokenId -> proxy contract
+    mapping (uint256 => address) public proxies;
+
+    constructor(address _proxyToClone, address _target1155) public {
+        proxyToClone = _proxyToClone;
+        target1155 = _target1155;
+        // This contract will be able to do anything with the user's 1155 tokens.
+        IERC1155(target1155).setApprovalForAll(address(this), true);
```
nuevoalex: The way this actually has to work is each user will call `setApprovalForAll` for this contract.

packages/augur-core/src/contracts/trading/ERC20Proxy.sol
```diff
@@ -0,0 +1,126 @@
...
+
+    function newERC20(uint256 _tokenId) public returns (ERC20Proxy) {
+        require(proxies[_tokenId] == address(0)); // must not have been registered yet
+        _erc20 = ERC20Proxy(createClone(proxyToClone));
+        proxies[_tokenId] = _erc20; // register
+        _erc20.initialize(this, _tokenId);
+        return _erc20;
+    }
+
+    function transfer(address _proxy, address _caller, address _to, uint256 _amount) public returns (bool) {
```
nuevoalex: This and the methods below need to validate that the msg.sender is a registered ERC20 proxy

packages/augur-core/src/contracts/trading/ERC20Proxy.sol
```diff
@@ -0,0 +1,126 @@
...
+    mapping (uint256 => address) public proxies;
+
+    constructor(address _proxyToClone, address _target1155) public {
+        proxyToClone = _proxyToClone;
+        target1155 = _target1155;
+        // This contract will be able to do anything with the user's 1155 tokens.
+        IERC1155(target1155).setApprovalForAll(address(this), true);
+    }
+
+    function newERC20(uint256 _tokenId) public returns (ERC20Proxy) {
```
nuevoalex: It's probably fine to have this contract do the ERC20 generation, though the pattern we have elsewhere is to have a specific factory contract which is called by contracts like this to generate them. Just easier to know where things happen I guess

packages/augur-core/src/contracts/trading/ERC20Proxy.sol
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    function transfer(address _proxy, address _caller, address _to, uint256 _amount) public returns (bool) {
+        _transfer(_proxy, _caller, _to, _amount);
+        return true;
+    }
+
+    function transferFrom(address _proxy, address _caller, address _from, address _to, uint256 _amount) public returns (bool) {
+        _transfer(_proxy, _from, _to, _amount);
+        _approve(_proxy, _from, _caller, allowances[_tokenId][_from][_caller].sub(_amount));
```
nuevoalex: this should happen first so we can fail earlier if the allowance isn't sufficient.

packages/augur-core/src/contracts/trading/ERC20Proxy.sol
nuevoalex: Ah I see, this is done via the `_gate` calls in the internal methods. NM

packages/augur-core/src/contracts/trading/ERC20Proxy.sol
robert-e-davidson3: Our ERC20.sol contract actually does it in this order so I just copied it. `packages/augur-core/src/contracts/libraries/token/ERC20.sol:93`. I can reverse it here though.

packages/augur-core/src/contracts/trading/ERC20Proxy.sol
robert-e-davidson3: Our ERC20 contract uses uint8 here.

packages/augur-core/src/contracts/trading/ERC20Proxy.sol
robert-e-davidson3: OH duh!

robert-e-davidson3: Looking at the OZ ERC1155 implementation, it seems that the user will have to call the ShareToken contract's setApprovalForAll themselves. So there's nothing to be done here. It'll just fail if the user tries to do anything requiring approval.

packages/augur-core/tests/libraries/token/test_erc20_proxy.py
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env python
```
robert-e-davidson3, line 1: Ignore this file. Added by accident.

packages/augur-core/src/contracts/trading/ERC20Proxy.sol
robert-e-davidson3: Seems to work so I'm going to leave this as-is.

packages/augur-core/src/contracts/trading/erc20proxy1155/ERC20Proxy1155.sol
```diff
@@ -0,0 +1,53 @@
+import 'ROOT/libraries/CloneFactory.sol';
```
nuevoalex: missing pragma at top

packages/augur-core/src/contracts/trading/erc20proxy1155/IERC20Proxy1155Nexus.sol
```diff
@@ -0,0 +1,40 @@
...
+import 'ROOT/libraries/token/IERC1155.sol';
+import 'ROOT/libraries/math/SafeMathUint256.sol';
+
+import 'ROOT/trading/erc20proxy1155/IERC20Proxy1155.sol';
+
+// To use these ERC20s, the user will have to call
+// setApprovalForAll(ERC20Proxy1155Nexus, true)
+// on the ShareToken contract themselves.
+
+contract IERC20Proxy1155Nexus is CloneFactory {
```
nuevoalex: The actual ERC20Proxy1155Nexus should inherit from CloneFactory instead of this interface. That'll keep imports of the interface trim and implementation out of an interface

robert-e-davidson3: oh right!

packages/augur-core/src/contracts/trading/erc20proxy1155/ERC20Proxy1155.sol
robert-e-davidson3: derp ty

packages/augur-core/src/contracts/trading/erc20proxy1155/ERC20Proxy1155Nexus.sol
```diff
@@ -0,0 +1,84 @@
...
+    mapping (uint256 => mapping (address => mapping (address => uint256))) public allowances;
+    // tokenId -> proxy contract
+    mapping (uint256 => IERC20Proxy1155) public proxies;
+
+    constructor(address _proxyToClone, IERC1155 _target1155) public {
+        proxyToClone = IERC20Proxy1155(_proxyToClone);
+        target1155 = _target1155;
+    }
+
+    function newERC20(uint256 _tokenId) public returns (IERC20Proxy1155) {
```
nuevoalex, line 28: Would be nice to have a convenience function to create all the ERC20s for a given market too. I suspect that's what will normally be used since just one cant really be used effectively

robert-e-davidson3: That's true. I'll make a bulk version.

packages/augur-core/src/contracts/trading/erc20proxy1155/ERC20Proxy1155Nexus.sol
```diff
@@ -0,0 +1,84 @@
...
+        uint256 _tokenId = _proxy.tokenId();
+        return allowances[_tokenId][_owner][_spender];
+    }
+
+    function _transfer(IERC20Proxy1155 _proxy, address _from, address _to, uint256 _amount) internal {
+        uint256 _tokenId = _proxy.tokenId();
+        _gate(_proxy, _tokenId);
+
+        bytes memory _nothing = new bytes(0);
+        target1155.safeTransferFrom(_from, _to, _tokenId, _amount, _nothing);
```
nuevoalex: We should actually use `unsafeTransferFrom` which doesn't even accept a data argument. The problem with `safeTransferFrom` is that while it prevents people form accidentally depositing into contracts not willing to accept them that way, it also raises a bunch of ways a malicious user can lock up a trading contract and puts a requirement on contract integrations to support a very specific standard. `unsafeTransferFrom` is ironically probably way safer

packages/augur-core/src/contracts/trading/erc20proxy1155/ERC20Proxy1155.sol
```diff
@@ -0,0 +1,53 @@
...
+        return nexus.approve(this, msg.sender, _spender, _amount);
+    }
+
+    function allowance(address _owner, address _spender) public view returns (uint256) {
+        return nexus.allowance(this, _owner, _spender);
+    }
+
+    function logApproval(address _owner, address _spender, uint256 _amount) public {
+        require(msg.sender == address(nexus));
+        emit Approval(_owner, _spender, _amount);
```
nuevoalex, line 41: These logs should also come from the nexus along with the tokenId. This way only one contract needs to be listened to in order to get these logs.

nuevoalex: Actually nevermind. This is just to sate the ERC20 standard. We dont even need to listen to these and technically noone does

packages/augur-core/src/contracts/trading/erc20proxy1155/ERC20Proxy1155Nexus.sol
robert-e-davidson3: I can do that. `unsafeTransferFrom` is just on ShareToken though so I'd need to integrate against it specifically instead of a generic 20-1155 proxy. Which is fine I just liked that it was generic.

packages/augur-core/src/contracts/trading/erc20proxy1155/ERC20Proxy1155.sol
robert-e-davidson3: You can already get these logs from the ERC1155 contract though.

packages/augur-core/src/contracts/trading/erc20proxy1155/IERC20Proxy1155Nexus.sol
```diff
@@ -0,0 +1,40 @@
...
+    IERC1155 public target1155;
+
+    // tokenId -> owner -> spender -> allowance
+    // The spender is whoever called the proxy contract, which is msg.sender in the proxy and _caller here in the nexus.
+    mapping (uint256 => mapping (address => mapping (address => uint256))) public allowances;
```
nuevoalex: we're double defining this. Fine to put it on the interface contract IMO if its to help access via interface import

robert-e-davidson3: of the two, proxies is useful elsewhere but allowances isn't. the allowance method provides that info.

packages/augur-core/src/contracts/trading/erc20proxy1155/ERC20Proxy1155Nexus.sol
robert-e-davidson3: done

nuevoalex (commented):
Looks good! Just need to mod artifacts / addresses in the config and the dpeloyment script now

robert-e-davidson3:
> Looks good! Just need to mod artifacts / addresses in the config and the dpeloyment script now

Added!

nuevoalex (approved):
Yay Create2!

robert-e-davidson3:
@Mergifyio backport para_deploys

PR #8979, opened 2020-08-11
Consolidate db read/write tasks into one queue

justinbarry:


packages/augur-sdk/src/state/db/RollbackTable.ts
```diff
@@ -55,7 +55,7 @@ export class RollbackTable extends AbstractTable {
             const maxBlock =_.get(_.maxBy(documents, 'blockNumber'), 'blockNumber');
             await this.prune(maxBlock);
         }
-        await super.bulkPutDocumentsInternal(documents);
+        await super.bulkPutDocuments(documents);
```
nuevoalex, line 58: This will put this back in the queue. Its already gone through the queue though so seems it should still be internal

PR #8987, opened 2020-08-11
Updates to sportsgroups

JohnDanz:
updates to sportsgroup objects, first steps toward adding charts, still not styled 100% correctly but want to get something up. should only be working on futures pages.

### DIFF
--- a/packages/augur-ui/src/modules/market-charts/components/price-history/price-history.styles.less
+++ b/packages/augur-ui/src/modules/market-charts/components/price-history/price-history.styles.less
@@ -1,4 +1,4 @@
-@import (reference) "~assets/styles/shared";
+@import (reference) '~assets/styles/shared';
 
 // Used in Betting UI
 .PriceHistory {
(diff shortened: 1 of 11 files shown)


packages/augur-ui/src/modules/market-charts/components/price-history/price-history.styles.less
```diff
@@ -83,17 +84,34 @@
...
 
-      g.highcharts-yaxis-labels > g.highcharts-crosshair-label > text {
-        .mono-10;
+        > text {
+          .mono-10;
 
-        fill: var(--color-primary-text);
-        transform: translateX(38px);
+          fill: var(--color-primary-text);
+          transform: translateX(38px);
```
phoebemirman: should this be a constant?

phoebemirman (changes requested):
just one comment about css constant

phoebemirman (approved):

PR #8990, opened 2020-08-11
split approve and migrate into two actions

andrewdamelio:


bthaile:
seems we'd want to use the existing approval component so that the UI is consistent.

bthaile (approved):
lgtm, I would think the easiest approach would have been to use the existing approval component. but this looks good for now.

Issue #8883, opened 2020-08-04
Hotloaded markets should have clearer UX


pgebheim:
How about showing loading spinners in the spots that are still loading info. That way the info like market title loads immediately and nothing jumps around when the rest of the info loads. This informs the user that something is happening rather than just leaving it blank

### Prototype (can take a while to load!): https://framer.com/share/Trading-Hot-Loading-Animations-5xmNPfVviezZpCuMbPy2/iIgD0q6Wi?fullscreen=1&highlights=0

![image](https://user-images.githubusercontent.com/39496225/89523413-16e6e480-d7e3-11ea-9335-3dce15f6af10.png)

### Progress Indicator

Lets update to this style:
https://material.io/components/progress-indicators#circular-progress-indicators

Colour for progress indicator #9599A5

matt-bullock:
I'd already added this to 8489:

How about showing loading spinners in the spots that are still loading info. That way the info like market title loads immediately and nothing jumps around when the rest of the info loads. This informs the user that something is happening rather than just leaving it blank

Prototype (can take a while to load!): https://framer.com/share/Trading-Hot-Loading-Animations-5xmNPfVviezZpCuMbPy2/iIgD0q6Wi?fullscreen=1&highlights=0

![image](https://user-images.githubusercontent.com/39496225/89164851-ebac8d00-d577-11ea-849b-0d5325d4c9eb.png)

matt-bullock:
I also suggest we use the style of spinners used in the prototype. They're a bit nicer than what we currently have

PR #9001, opened 2020-08-12
Add hotloading loading market card

bthaile:
https://github.com/AugurProject/augur/issues/8883

filter out warp sync market
filter out non open markets
populate data in place when sdk is synced
categories nav will need to be populated when user manually refreshes or navigates around.

after markets loaded from warp sync file
image


andrewdamelio (approved):
Looks good!

Issue #8755, opened 2020-07-30
Add version number for UI


KillariDev:
Because there's now two versions of the UI's going around in the internet, it's confusing which version you are on.

Add version number somewhere on the UI (to the top or to the bottom).

Chwy5:
Can look to add it to the Acct Summary pg either at the bottom or within Augur Status

matt-bullock:
We already have it in the design

https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=2704%3A981

![image](https://user-images.githubusercontent.com/39496225/88930886-645fd080-d27c-11ea-9557-218fc32d25df.png)

veox:
IMO this should be available even prior to login, so Account Summary is not a good location.

EDIT: Bottom of page as shown above is pretty good.

matt-bullock:
@veox good point. I'll see if I can work it into somewhere obvious where you don't need to be logged in

matt-bullock:
Could  also go at the bottom of the market list page. That way you can see it whether logged in or not

![image](https://user-images.githubusercontent.com/39496225/88942414-87917c80-d28a-11ea-9cf4-f4a24043d9e8.png)

experiencedft:
I think something at the top besides the Augur logo would work better. As a user, I don't even want to have to scroll to see which version of the UI I'm on

matt-bullock:
Agree, but once we add in the app switcher (switch between trading and betting UIs) that top area will be too cramped at smaller sizes. Lets add it to the footer for now as this requires minimal changes. Both on market list page and account summary (we already have the footer in account summary anyway, just need to add version number)

[Account Summary design](https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=2704%3A981)

[Market list page design](https://www.figma.com/file/Do4qnkCNM8P7lwBmCdYZCS4X/Market-List-Page?node-id=3808%3A99422)

experiencedft:
I am of the opinion that we can worry about the app switcher design when the betting UI is actually deployed. Right now it is urgent to make it very easy for users to immediatly identify which version they are on without having to interact with the page at all, and the header is the best place for that.

matt-bullock:
See what others say but if people are in agreement then it can go here if wanted in the header until we need the app switcher

[...]

PR #9003, opened 2020-08-12
add current version to the bottom of pages

bthaile:
https://github.com/AugurProject/augur/issues/8755

uses `process.env.CURRENT_VERSION` to set version on bottom of pages.


andrewdamelio (approved):
Excellent!!

Issue #7403, opened 2020-04-21
Sports Betting - My Account - Page Layout Responsive


yanelbottini:
Mobile layout:

[Figma link](https://www.figma.com/file/wogMuUDylgOHgTEYhZ3jgp/Augur-Bet-UI?node-id=4153%3A196694)

Responsive references:

[Figma link](https://www.figma.com/file/wogMuUDylgOHgTEYhZ3jgp/Augur-Bet-UI?node-id=3272%3A1612)

JohnDanz:
https://github.com/AugurProject/augur/issues/8873

yanelbottini:
In general looks fine. Some comments:

- I'm seeing a double scroll bar on tablet breakpoint:

<img width="672" alt="Screen Shot 2020-09-16 at 09 56 05" src="https://user-images.githubusercontent.com/49915900/93340761-a9779c00-f803-11ea-96fe-e3d7c9e3361d.png">

- Can we check if notifications are showing up on mobile? Because right now seems empty to me, but I'm seeing the notifications on desktop.

![WhatsApp Image 2020-09-16 at 10 00 35 (2)](https://user-images.githubusercontent.com/49915900/93341567-a6c97680-f804-11ea-9d69-25f97ffae46e.jpeg)
<img width="521" alt="Screen Shot 2020-09-16 at 10 09 39" src="https://user-images.githubusercontent.com/49915900/93341645-c19beb00-f804-11ea-8115-947309d0be7a.png">

- Can we reduce the size we use to show page title and tabs on mobile?

<img width="875" alt="Screen Shot 2020-09-16 at 10 13 24" src="https://user-images.githubusercontent.com/49915900/93342083-46870480-f805-11ea-91a5-772d08044845.png">

- Filter icon seems different (First one design, second dev)

<img width="155" alt="Screen Shot 2020-09-16 at 10 16 56" src="https://user-images.githubusercontent.com/49915900/93342476-c57c3d00-f805-11ea-8580-f42b21399767.png">

<img width="237" alt="Screen Shot 2020-09-16 at 10 17 13" src="https://user-images.githubusercontent.com/49915900/93342515-cf9e3b80-f805-11ea-918d-1aa19af7b55e.png">


- Please, update to match design

<img width="1139" alt="Screen Shot 2020-09-16 at 10 16 04" src="https://user-images.githubusercontent.com/49915900/93342405-b1384000-f805-11ea-8729-3e5b7093c5ce.png">

[...]

PR #9163, opened 2020-09-01
Bruno/quadbox refactor

bconfortin:
**It might be better to get inside this branch and check it in the IDE**
Big PR since it has the quadbox/filterbox/filterswitchbox refactor.

#7403


phoebemirman (changes requested):
filled and open orders are missing their dropdown. also when you expand a collapsed box, I think it is supposed to make the two boxes share the space, not have it take up the whole space. similar to how trading page works.

phoebemirman:
seeing this as well 
image

phoebemirman:
also dropdown should be hidden when search is on 
image

also this market is missing my unsigned orders alert for some reason

also when you click on the search box X, the text is supposed to be cleared

phoebemirman:
can we add market counts here 
image

favorites have an arrow that doesn't do anything
image

JohnDanz (changes requested):
Screen Shot 2020-09-04 at 8 08 14 AM


need to have some space between these on mobile

JohnDanz:
Screen Shot 2020-09-04 at 8 10 49 AM

the search bar should fill the rest of the space, so we have only a small ~8px gap between filter and search, should grow when clicked to fill the entire top area and cover the filters until unfocused.

JohnDanz:
Screen Shot 2020-09-04 at 8 11 03 AM

similar issue here as the above, in this case, the search bar should cover the entire top section since there is no filter option and it should be aligned with the spacing of the content below.

phoebemirman:
toggling is very broken in portfolio 
looks like this:
Screen Shot 2020-09-04 at 10 12 44 AM

then when I click the top quad expand it becomes this:
Screen Shot 2020-09-04 at 10 12 49 AM
so there is no way to even out the space

phoebemirman:
this is broken as well. shows in reporting as open
Screen Shot 2020-09-04 at 10 16 19 AM

phoebemirman (changes requested):
spacing not matching original:
Screen Shot 2020-09-04 at 10 38 00 AM

looking off:
Screen Shot 2020-09-04 at 10 36 27 AM

JohnDanz (approved):

phoebemirman (approved):

Issue #9090, opened 2020-08-21
Refactor Trading


JohnDanz:
Alright, been side stepping this for a bit. We need to refactor the data flow on the trading page for trading ui so that we can not have so much input lag on this page. It's not optimized at all outside of redux and even then it hangs. It needs to be simplified and re-written.

Chwy5:
closing

PR #9179, opened 2020-09-03
Refactor marketview

JohnDanz:
Still probably some smaller pieces we will run into, but this is already substantial.

https://github.com/AugurProject/augur/issues/9090


phoebemirman (changes requested):
outcome dropdown in the form is not usable for multiple choice markets

packages/augur-ui/src/modules/common/buttons.tsx
```diff
@@ -217,22 +217,34 @@ export const ChatButton = ({ action, disabled }: DefaultButtonProps) => (
...
+  ...props,
+}: ProcessingButtonProps) => {
   const { pendingQueue } = useAppStatusStore();
   let disabled = false;
-
   const pendingData =
-    pendingQueue[props.queueName] &&
-    pendingQueue[props.queueName][props.queueId];
+    pendingQueue[queueName] &&
+    pendingQueue[queueName][queueId];
```
bconfortin, line 238: Is it possible to do like `pendingQueue[queueName]?[queueId]`? I really don't know, just got me wondering.

packages/augur-ui/src/modules/common/labels.tsx
```diff
@@ -721,79 +722,70 @@ export const TextLabel = ({ text, keyId }: TextLabelProps) => {
...
-          </span>
-        )}
-      </span>
-    );
-  }
+export const HoverValueLabel = ({
+    value = null,
+    showDenomination = false,
+    useFull = false
+  }) => {
```
bconfortin: I think it would be wise to add an interface here so we know what types can be passed as value.

packages/augur-ui/src/modules/common/buttons.tsx
JohnDanz: it's not unfortunately, it has to be `?.` for the check to work

phoebemirman (approved):

packages/augur-ui/src/modules/market-charts/components/price-history/price-history.tsx
```diff
@@ -249,7 +249,7 @@ const handleSeries = (
         id === '0' &&
         !isTrading &&
         priceTimeSeries[id][priceTimeSeries[id].length - 1].price === '0';
-      const isSelected = selectedOutcomeId && selectedOutcomeId == id;
+      const isSelected = selectedOutcomeId == id;
```
bconfortin: Shouldn't `==` be `===`?

bconfortin: Unless you are comparing different types, like string and integer
`"0" == 0`

packages/augur-ui/src/modules/market/components/market-header/market-header.tsx
```diff
@@ -17,94 +16,92 @@ import {
...
-  }
+      reportingState === CROWDSOURCING_DISPUTE ||
+      reportingState === OPEN_REPORTING ||
+      (reportingState === DESIGNATED_REPORTING &&
+        isSameAddress(designatedReporter, address))
+  );
 
-  const description = marketSelected.description || '';
-  let details = marketSelected.details || detailsText || '';
+  const description = marketDesctipion || '';
```
bconfortin: Typo: marketDesctipion

packages/augur-ui/src/modules/market-charts/components/price-history/price-history.tsx
JohnDanz: this is intentional, one is a number, the other is a string.

packages/augur-ui/src/modules/market/components/market-header/market-header.tsx
JohnDanz: good catch, will fix. stupid new keyboard ;)

bconfortin (approved):

PR #9181, opened 2020-09-03
AMM

robert-e-davidson3:


packages/augur-core/src/contracts/para/AMMExchange.sol
```diff
@@ -8,236 +8,251 @@ import 'ROOT/ICash.sol';
...
+    using SafeMathUint256 for uint256;
+
+    ICash public cash;
+    IParaShareToken public shareToken;
+    IMarket public augurMarket;
+    uint256 public numTicks;
+    uint256 public INVALID;
+    uint256 public NO;
+    uint256 public YES;
+    uint256 public fee; // [0-1000] how many thousandths of swaps should be kept as fees
```
MicahZoltu, line 22: These can all be immutable, which will save a significant amount of gas.

packages/augur-core/src/contracts/para/AMMExchange.sol
```diff
@@ -8,236 +8,251 @@ import 'ROOT/ICash.sol';
...
+    // Adds shares to the liquidity pool by minting complete sets.
+    function addLiquidity(uint256 _sharesToBuy) external {
+        cash.transferFrom(msg.sender, address(this), _sharesToBuy.mul(numTicks));
+        shareToken.publicBuyCompleteSets(augurMarket, _sharesToBuy);
+
+        uint256 _newLiquidityConstant = calculateLiquidityConstant();
+
+        if (liquidityConstant == 0) {
+            // The formula for issuing LP tokens breaks down when the supply (and therefore liquidityConstant) are zero.
+            _mint(msg.sender, _newLiquidityConstant);
```
MicahZoltu: There is a known bug in Uniswap v1 where someone can permanently break an exchange by setting it up with a tiny bit of liquidity.  This may not matter for this project like it does with Uniswap, especially if there can be multiple exchanges per market.

packages/augur-core/src/contracts/para/AMMExchange.sol
```diff
@@ -8,236 +8,251 @@ import 'ROOT/ICash.sol';
...
+        } else {
+            uint256 _totalSupply = totalSupply;
+            // User gains LP tokens relative to how many LP tokens were already issued, and the change in the ratio of shares in the pool.
+            // User gains more LP tokens if they are an earlier
+            _mint(msg.sender, _totalSupply.mul(_newLiquidityConstant).div(liquidityConstant).sub(_totalSupply));
```
MicahZoltu: This is notably different from the Uniswap function.  I'm curious how you came to it?

/me prays that you don't say "I copied it from your code Micah" because he has no idea where he got it...

packages/augur-core/src/contracts/para/AMMExchange.sol
```diff
@@ -8,236 +8,251 @@ import 'ROOT/ICash.sol';
...
+    // Removes shares from the liquidity pool; does not redeem complete sets for cash.
+    function removeLiquidity(uint256 _poolTokensToSell) external {
+        uint256 _poolSupply = totalSupply;
+        (uint256 _poolInvalid, uint256 _poolNo, uint256 _poolYes) = shareBalances(address(this));
+        uint256 _poolCash = cash.balanceOf(address(this));
+        uint256 _invalidShare = _poolInvalid.mul(_poolTokensToSell).div(_poolSupply);
+        uint256 _noShare = _poolNo.mul(_poolTokensToSell).div(_poolSupply);
+        uint256 _yesShare = _poolYes.mul(_poolTokensToSell).div(_poolSupply);
+        uint256 _cashShare = _poolCash.mul(_poolTokensToSell).div(_poolSupply);
+        _burn(msg.sender, _poolTokensToSell);
```
MicahZoltu, line 65: Consider doing the burn first, just as a precautionary step (so we can think less about reentrancy).

packages/augur-core/src/contracts/para/AMMExchange.sol
```diff
@@ -8,236 +8,251 @@ import 'ROOT/ICash.sol';
...
+        shareTransfer(address(this), msg.sender, _invalidShare, _noShare, _yesShare);
+        cash.transfer(msg.sender, _cashShare);
+        liquidityConstant = calculateLiquidityConstant();
```
MicahZoltu: Why is this a storage variable?  Seems like it can be a local calculated from pool balance.  Perhaps it is an attempt at gas savings?

packages/augur-core/src/contracts/para/AMMExchange.sol
```diff
@@ -8,236 +8,251 @@ import 'ROOT/ICash.sol';
...
+            // yesToUser + poolYes - poolConstant / (poolNo + _setsToBuy)
+            return _setsToBuy.add(_poolYes.sub(_poolConstant.div(_poolNo.add(_setsToBuy))));
+        } else {
+            return _setsToBuy.add(_poolNo.sub(_poolConstant.div(_poolYes.add(_setsToBuy))));
+        }
+    }
+
+    // If you do not have complete sets then you must have more shares than _setsToSell because you will be swapping
+    // some of them to build complete sets.
+    function exitPosition(uint256 _setsToSell) external {
```
MicahZoltu: This should have a `minCashReceived` parameter to protect users from frontrunners.

packages/augur-core/src/contracts/para/AMMExchange.sol
```diff
@@ -8,236 +8,251 @@ import 'ROOT/ICash.sol';
...
+            uint256 _noToPool = _poolConstant.div(_poolYes.sub(_yesToUser)).sub(_poolNo);
+            require(_noToPool <= _userNo.sub(_setsToSell), "AugurCP: You don't have enough NO tokens to close out for this amount.");
+            _yesFromUser = _userYes;
+            _noFromUser = _noToPool + _setsToSell;
+        }
+
+        return (_noFromUser, _yesFromUser);
+    }
+
+    function swap(uint256 _inputShares, bool _inputYes) external returns (uint256) {
```
MicahZoltu: Will need a parameter to prevent front running here too.  Something like `minimumOutputShares` or something.

packages/augur-core/src/contracts/para/AMMExchange.sol
```diff
@@ -8,236 +8,251 @@ import 'ROOT/ICash.sol';
...
+        } else { // gain yesses, lose nos
+            shareToken.unsafeTransferFrom(address(this), msg.sender, YES, _outputShares);
+            shareToken.unsafeTransferFrom(msg.sender, address(this), NO, _inputShares);
+        }
+
+        return _outputShares;
+    }
+
+    // How many of the other shares you would get for your shares.
+    function rateSwap(uint256 _inputShares, bool _inputYes) public returns (uint256) {
```
MicahZoltu: Why isn't this `view`?

packages/augur-core/src/contracts/para/AMMExchange.sol
robert-e-davidson3: It is. I figured it would be cheaper if adding liquidity were a bit more common than removing liquidity. But today I learned storage will become significantly more costly. I'll switch back to calculating each time.

packages/augur-core/src/contracts/para/AMMExchange.sol
robert-e-davidson3: Eek prays unanswered. I have the uniswap code so I'll look there.

packages/augur-core/src/contracts/para/AMMExchange.sol
MicahZoltu: Storage *reads* are going to be come more expensive, so this optimization may still be valuable if it minimizes the number of reads.

packages/augur-core/src/contracts/para/AMMExchange.sol
MicahZoltu: This is the most confusing part of Uniswap to me.  😬
Maybe I got it from Uniswap v1 where I think it is a bit simpler.  May be worth checking there as well as the v2 contracts.

packages/augur-core/src/contracts/para/AMMExchange.sol
robert-e-davidson3: fixed

packages/augur-core/src/contracts/para/AMMExchange.sol
robert-e-davidson3: It adds a storage read actually so it's just worse. Switching to calculating each time.

packages/augur-core/src/contracts/para/AMMExchange.sol
robert-e-davidson3: You'll always receive the same amount of cash for a given amount of complete sets. However, the risk here is in how many shares you swap away to build complete sets. I'll add `_maxSharesSwappedAway`.

Looks like we're going to switch to doing the quadratic equation on-chain though so this is probably moot.

packages/augur-core/src/contracts/para/AMMExchange.sol
robert-e-davidson3: added

packages/augur-core/src/contracts/para/AMMExchange.sol
robert-e-davidson3: I think it will still matter. I assume it's this? https://github.com/Uniswap/uniswap-v2-core/blob/master/contracts/UniswapV2Pair.sol#L121

packages/augur-core/src/contracts/para/AMMExchange.sol
robert-e-davidson3: Yeah it's confusing. Reserve is everywhere but I'm not sure of its purpose. Checking out v1.

For reference:
v1: https://github.com/Uniswap/uniswap-v1/blob/master/contracts/uniswap_exchange.vy#L56
v2: https://github.com/Uniswap/uniswap-v2-core/blob/master/contracts/UniswapV2Pair.sol#L123

packages/augur-core/src/contracts/para/AMMExchange.sol
MicahZoltu: Yeah, that is the one.  It doesn't matter if you can have multiple exchanges per pair, since if someone creates a broken pair you can just create another one to use instead.

packages/augur-core/src/contracts/para/AMMExchange.sol
MicahZoltu: `reserve` just means the amount of that token that the contract currently holds (or in the case of Uniswap v2, the amount of a token that Uniswap holds and is aware of).

packages/augur-core/src/contracts/libraries/math/SafeMathInt256.sol
```diff
@@ -71,4 +71,17 @@ library SafeMathInt256 {
...
+            int256 x = (y + 1) / 2;
+            z = y;
+            while (x < z) {
+                z = x;
+                x = (y / x + x) / 2;
+            }
+        } else if (y != 0) {
+            z = 1;
+        }
+    }
```
MicahZoltu, line 86: Recommend a test suite for this function specifically.  My confidence in it being totally correct is very low.  Should be easy to generate a whole bunch of test cases, with a focus around boundary conditions (e.g., right before, after, and on a rounding condition like `sqrt(15)`, `sqrt(16)`, `sqrt(17)`.  Also would be good to text whatever the largest int256 square is.

Note: I don't think we need negative numbers here, and that adds complexity so I would recommend just having this be `uint256` instead.

packages/augur-core/src/contracts/para/AMMExchange.sol
```diff
@@ -1,243 +1,325 @@
...
 
-		// short circuit if user is closing out their own complete sets
-		if (_userInvalid >= _setsToSell && _userNo >= _setsToSell && _userYes >= _setsToSell) {
-			shareTransfer(msg.sender, address(this), _setsToSell, _setsToSell, _setsToSell);
-			cash.transfer(msg.sender, _cashToBuy);
-			return;
+		// Figure out how many shares we're buying in our synthetic swap and use that to figure out the final balance of Yes/No (setsToSell)
+		if (_yesShares > _noShares) {
+			uint256 _delta = _yesShares.sub(_noShares);
+			uint256 _noSharesToBuy = quadratic(1, -int256(_delta.add(_poolYes).add(_poolNo)), int256(_delta.mul(_poolNo)), _yesShares);
```
MicahZoltu: Seeing as how all call sites have the second parameter as negative, consider instead keeping these as positive `uint256` variables and just having the `quadratic` function subtract instead of add.  I believe this will actually clean up the quadratic formula quite a bit if you do a bit of algebra to extract out the sign.

packages/augur-core/src/contracts/para/AMMExchange.sol
```diff
@@ -1,243 +1,325 @@
...
-			uint256 _noFromUser = _inputShares;
-			uint256 _yesToUser = _poolYes.sub(_poolConstant.div(_poolNo.add(_noFromUser)));
-			shareToken.unsafeTransferFrom(msg.sender, address(this), NO, _noFromUser);
-			shareToken.unsafeTransferFrom(address(this), msg.sender, YES, _yesToUser);
-			return _yesToUser;
-		}
+	function quadratic(int256 _a, int256 _b, int256 _c, uint256 _maximum) public returns (uint256) {
+		int256 _piece = SafeMathInt256.sqrt(_b*_b - (_a.mul(_b).mul(4)));
+		int256 _resultPlus = (-_b + _piece) / (2 * _a);
+		int256 _resultMinus = (-_b - _piece) / (2 * _a);
```
MicahZoltu:
```suggestion
		int256 _piece = SafeMathInt256.sqrt(_b*_b + (_a.mul(_b).mul(4)));
		int256 _resultPlus = (_b + _piece) / (2 * _a);
		int256 _resultMinus = (_b - _piece) / (2 * _a);
```
Per suggestion above, I think by passing in positive values as `b` instead of negative values, we can not only make this function `uint256` instead of `int256`, but we get rid of a bunch of awkward negatives as well.

packages/augur-core/src/contracts/para/AMMExchange.sol
```diff
@@ -1,243 +1,325 @@
...
-			return;
+		// Figure out how many shares we're buying in our synthetic swap and use that to figure out the final balance of Yes/No (setsToSell)
+		if (_yesShares > _noShares) {
+			uint256 _delta = _yesShares.sub(_noShares);
+			uint256 _noSharesToBuy = quadratic(1, -int256(_delta.add(_poolYes).add(_poolNo)), int256(_delta.mul(_poolNo)), _yesShares);
+			_setsToSell = _noShares.add(_noSharesToBuy);
+		}
+		else if (_noShares > _yesShares) {
+			uint256 _delta = _noShares.sub(_yesShares);
+			uint256 _yesSharesToBuy = quadratic(1, -int256(_delta.add(_poolYes).add(_poolNo)), int256(_delta.mul(_poolYes)), _noShares);
```
MicahZoltu: Same comment as above, get rid of this negative and simplify all the things and stop using `int256`.

packages/augur-core/src/contracts/para/AMMExchange.sol
```diff
@@ -1,243 +1,325 @@
...
+			_setsToSell = _invalidFromUser;
 		}
 
-		// materialize the complete set sale for cash
+		uint256 _cashPayout = _setsToSell.mul(numTicks);
+		require(_cashPayout >= _minCashPayout, "Proceeds were less than the required payout");
+
 		shareTransfer(msg.sender, address(this), _invalidFromUser, _noFromUser, _yesFromUser);
-		cash.transfer(msg.sender, _cashToBuy);
+        cash.transfer(msg.sender, _cashPayout);
```
MicahZoltu: Tabs vs spaces on this line.

MicahZoltu (commented):
Glanced over the new quadratic stuff, at a glance seems fine though I did have a few comments, particularly around the choice of `int256`.

packages/augur-core/src/contracts/libraries/math/SafeMathInt256.sol
robert-e-davidson3: Wow yeah this and the SafeMathUint256 sqrt are untested. Will add!

packages/augur-core/src/contracts/para/AMMFactory.sol
```diff
@@ -0,0 +1,34 @@
...
+    // market -> para -> amm
+    AMMExchange internal proxyToClone;
+
+    constructor(address _proxyToClone) public {
+        proxyToClone = AMMExchange(_proxyToClone);
+    }
+
+    function addAMM(IMarket _market, IParaShareToken _para, uint256 _fee) external returns (AMMExchange) {
+        address _amm = AMMExchange(createClone2(address(proxyToClone), salt(_market, _para)));
+        _amm.initialize(_market, _para, _fee);
```
MicahZoltu: I recommend running the numbers on deployment cost vs usage costs.  By deploying like this, you can't have any immutable constants which means *every* call is going to be more expensive, and with the upcoming changes that increase storage reads to something like 2500 gas this is very likely to add up to a big number.

The AMM contract is small enough, that the increased gas cost in a direct deploy may be worth it since runtime costs will be substantially lower.

PR #9203, opened 2020-09-08
Replaced "REP bond" with "Creation bond"

CrystalBallBe:


pgebheim:
This is never called the creator bond, in the UI its called the "no-show" bond.

CrystalBallBe:
@pgebheim 
Thank you for checking this pull request.

I still think the bond should be called ”creation bond”.

My understanding is that the bond, which is paid by REP when creating a market, is the maximum of the no-show bond and the designated reporter bond. 
That is mentioned in this block of the whitepaper.
image

So, I don't think that "no-show bond" correctly express the meaning of the bond.
And the reason why I call the bond "**creation** bond" (not "**creator** bond"), because I heard that devs go with "creation bond".(Reference: https://discordapp.com/channels/378030344374583298/617739423920029779/752789093297160262)

pgebheim:
You know, you're actually correct. I'll merge this but first I need to rebase this against our new production branch.

Issue #9088, opened 2020-08-21
Template Display Bug


Chwy5:
The W needs to be in ( ) like the (M) below see screenshot (this was a Women's tennis template for SB

![image](https://user-images.githubusercontent.com/41301809/90890272-a2d24200-e387-11ea-8eda-d74390ed09b2.png)

phoebemirman:
updated but won't work for templates that have been created already

PR #9301, opened 2020-09-23
add parantheses to gender templates

phoebemirman:
https://github.com/AugurProject/augur/issues/9088


bthaile (changes requested):
second thought. 
don't change templates-source.ts

and add a post processing step when placeholder titles are generated using abbreviations to change (( to ( 
and 
)) to )

then we don't have to retire templates

bthaile (approved):
lgtm

PR #9330, opened 2020-09-28
update build configs

andrewdamelio:


bthaile (approved):
Lgtm, more changes needed to update-assets to call helper contract to get balances. Can be on another PR

PR #9332, opened 2020-09-28
Add ParaAugurDeployFinished event to deployer

justinbarry:


pgebheim:
We will also need the ParaAugur and ParaAugurTrade addresses as that will power the rest of the augur state syncing.

pgebheim:
@justinbarry -- this needs to be rebased and the contracts re-built

packages/augur-core/src/contracts/para/ParaDeployer.sol
```diff
@@ -199,6 +209,7 @@ contract ParaDeployer is Ownable {
     function deployTrade(address _token) private {
         address _trade = factories.tradeFactory.createTrade();
         paraAugurTradings[_token].registerContract("Trade", _trade);
+        IParaAugur _paraAugur =  paraAugurs[_token];
```
SwerveFinance, line 212: Can be removed I think

pgebheim:
@justinbarry are you combining this PR with #9319 ?

Issue #9292, opened 2020-09-22
Go to trading button to create a market should link to the start of create a market flow


matt-bullock:
![image](https://user-images.githubusercontent.com/39496225/93909589-3341d080-fd00-11ea-8a04-2928a656d898.png)

PR #9341, opened 2020-09-29
Added a redirect when clicking the button

bconfortin:
#9292

### DIFF
--- a/packages/augur-ui/src/modules/portfolio/components/markets/markets.tsx
+++ b/packages/augur-ui/src/modules/portfolio/components/markets/markets.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import classNames from 'classnames';
-import { Link } from 'react-router-dom';
+import { Link, Redirect } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


jimsmithson111:
why does there need to be a redirect?

packages/augur-ui/src/modules/portfolio/components/markets/markets.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import classNames from 'classnames';
-import { Link } from 'react-router-dom';
+import { Link, Redirect } from 'react-router-dom';
```
JohnDanz: get rid oft his import since I don't think it's used.

JohnDanz (changes requested):

packages/augur-ui/src/modules/portfolio/components/markets/markets.tsx
bconfortin: Done

JohnDanz (approved):

PR #9395, opened 2020-10-06
onboarding updates

andrewdamelio:


packages/augur-ui/src/modules/modal/components/modal-view.tsx
```diff
@@ -23,6 +22,11 @@ import { Loading as ModalLoading } from 'modules/modal/loading';
 import { ModalUniverseSelector } from 'modules/modal/components/modal-universe-selector';
 import { ModalTestBet } from 'modules/modal/onboarding-modals';
 import { ModalAugurUsesDai } from 'modules/modal/onboarding-modals';
+import { ModalApprovals } from 'modules/modal/onboarding-modals';
+import { ModalEthDeposit } from 'modules/modal/onboarding-modals';
+import { ModalSwap } from 'modules/modal/onboarding-modals';
+import { ModalBankroll } from 'modules/modal/onboarding-modals';
+import { ModalTokenSelect } from 'modules/modal/onboarding-modals';
 import { ModalTutorialOutro } from 'modules/modal/onboarding-modals';
 import { ModalTutorialIntro } from 'modules/modal/onboarding-modals';
```
JohnDanz: can we just combine all of these into one import statement? there's no reason for us to do this as one liners when they are all coming from `modules/modal/onboarding-modals`

JohnDanz (changes requested):

JohnDanz (approved):

PR #9449, opened 2020-10-12
Fixed dropdown styles, footer, filters and missing buttons

bconfortin:
#7403

### DIFF
--- a/packages/augur-ui/src/modules/portfolio/components/common/sidebar-nav.styles.less
+++ b/packages/augur-ui/src/modules/portfolio/components/common/sidebar-nav.styles.less
@@ -62,6 +62,7 @@
   > div:nth-of-type(2) {
     padding: @size-24 @size-12;
     background-color: var(--color-page-background);
+    min-height: calc(100vh - @size-64);
 
     > div {
       > div,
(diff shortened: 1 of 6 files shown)


packages/augur-ui/src/modules/portfolio/components/common/sidebar-nav.styles.less
```diff
@@ -62,6 +62,7 @@
   > div:nth-of-type(2) {
     padding: @size-24 @size-12;
     background-color: var(--color-page-background);
+    min-height: calc(100vh - @size-64);
```
JohnDanz, line 65: shouldn't this be @top-bar-height-mobile?

JohnDanz (changes requested):

packages/augur-ui/src/modules/portfolio/components/common/sidebar-nav.styles.less
bconfortin: Not really. That's the height of the filter's header. I thought about using that variable but if we ever change the height for the top bar, we'll be messing around with the filter's header.

image

JohnDanz (approved):

Issue #9391, opened 2020-10-06
Mobile betslip transition (open and close)


matt-bullock:
![Kapture 2020-10-06 at 16 18 07](https://user-images.githubusercontent.com/39496225/95213957-aa429300-07ef-11eb-9833-57022bbc15ab.gif)

Prototype: https://framer.com/share/SportsBook-Mobile-Prototype-place-multiple-bets--8ditECPvRjfXSlIUfGJJ

Transition code

// On Tap - Navigation
type: "tween",
delay: 0,
duration: 0.3,
ease: [0.44, 0, 0.56, 1],

PR #9517, opened 2020-10-21
Fixed style and mobile animation for bestlip

bconfortin:
#9391 

Regarding the removal of the button to open the Betslip from inside the aside element:
To get a smooth animation I had to use `transform: translateY()`, and if I use that with the open Betslip button inside the aside element, the button also transforms and go to the bottom of the screen, even when using things like `position: fixed/absolute/sticky`.

Overall I think it was a good compromise. Test it out, see what you think.

### DIFF
--- a/packages/augur-ui/src/modules/trading/betslip.styles.less
+++ b/packages/augur-ui/src/modules/trading/betslip.styles.less
@@ -7,145 +7,30 @@
 :root[theme="BETTING"],
 :root[theme="SPORTS"] {
   .Betslip {
-    display: flex;
-    flex-flow: column nowrap;
     align-items: center;
+    background: var(--color-module-background);
+    display: flex;
+    flex-flow: row nowrap;
     justify-content: flex-start;
+    padding: 0;
 
     > div {
       display: none;
-      transition: background .3s ease;
-
-      > button:first-of-type {
-        .text-12-bold;
-
-        align-content: center;
-        align-items: center;
-        background: var(--color-primary-text);
-        color: var(--color-module-background);
-        cursor: pointer;
-        display: flex;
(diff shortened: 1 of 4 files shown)


packages/augur-ui/src/modules/trading/betslip.styles.less
```diff
@@ -166,6 +141,64 @@
...
+  }
+
+  .OpenBetslipButton {
+    display: none;
+
+    @media @breakpoint-mobile-tablet {
+      &.Minimized {
+        display: flex;
+        position: fixed;
+        bottom: 0.5rem;
```
phoebemirman: can you use constants

packages/augur-ui/src/modules/trading/betslip.styles.less
```diff
@@ -177,13 +210,13 @@
     box-shadow: 0 0 @size-4 var(--color-primary-text-15);
     position: relative;
     top: 0;
-    transition: all .3s cubic-bezier(0.44, 0, 0.56, 1);
+    transition: top .3s cubic-bezier(0.44, 0, 0.56, 1);
```
phoebemirman: we need it to be all for tablet/ this breaks tablet animation

phoebemirman (changes requested):
not seeing it animate at and around 870px wide

packages/augur-ui/src/modules/trading/betslip.styles.less
bconfortin: Oops

phoebemirman (changes requested):
tablet is missing the semi-transparent background behind the betslip when it is open

phoebemirman (approved):

Issue #9419, opened 2020-10-08
Mobile notifications visual fixes


matt-bullock:
- placement of reporting ends
- lack of space between reporting ends label and countdown
- white spacing between notifications needs removing
- all buttons should be full width

![image](https://user-images.githubusercontent.com/39496225/95441168-16d9a100-095a-11eb-9761-3b7607ddd42e.png)

matt-bullock:
![image](https://user-images.githubusercontent.com/39496225/96993628-dc602e80-152b-11eb-9cd7-23222be80583.png)
could you remove the white space between notifications. Rest looks good

PR #9524, opened 2020-10-22
Fixed mobile notifications' styles

bconfortin:
#9419

### DIFF
--- a/packages/augur-ui/src/modules/account/components/notifications.tsx
+++ b/packages/augur-ui/src/modules/account/components/notifications.tsx
@@ -20,6 +20,7 @@ import {
 } from 'modules/account/components/notifications-templates';
 import { Notification, QueryEndpoints, } from 'modules/types';
 import {
+  MARKET_IS_MOST_LIKELY_INVALID_TITLE,
   MODAL_CLAIM_FEES,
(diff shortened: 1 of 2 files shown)


packages/augur-ui/src/modules/account/components/notifications.tsx
```diff
@@ -368,7 +374,7 @@ const Notifications = ({ toggle }: NotificationsProps) => {
       toggle={toggle}
       customClass={classNames({
         [Styles.DarkBackgroundMobile]: notificationCount !== 0,
-        [Styles.NoBorderAndBgOnMobile]: notificationCount === 0,
+        [Styles.NoBorderAndBgOnMobile]: notificationCount !== 0,
```
JohnDanz: if these are both going to get applied at the same time every-time (same condition to apply it) then why don't we just drop one of these classes and combine their definitions?

JohnDanz (changes requested):

packages/augur-ui/src/modules/account/components/notifications.tsx
bconfortin: Good one. Done.

JohnDanz (approved):

PR #9588, opened 2020-10-30
Fix tablet transition

phoebemirman:


bconfortin (approved):
Awesome work :D lgtm!

Issue #9578, opened 2020-10-29
Mobile Account menu QA


matt-bullock:
- remove lines down left and right sides in background
- match design for add funds section
- font size and weight of balance to the top right (under metamask)
- metama... getting truncated too soon
- green bar at very top is missing
- Keep refer a friend section as we may keep this
- wallet provider section is missing from mobile
- logout icon looks small
- transaction fee section - space too large between texts
- remove this gap:
![image](https://user-images.githubusercontent.com/39496225/97612430-b816bd00-1a17-11eb-96d4-2d7a68656d78.png)

![image](https://user-images.githubusercontent.com/39496225/97611691-c87a6800-1a16-11eb-90ed-afa4b85695ea.png)

matt-bullock:
@bconfortin still a couple of things:

Still seeing white lines down left and ride sides - is the background layer not wide enough or something?

![image](https://user-images.githubusercontent.com/39496225/98106470-aad75380-1e99-11eb-86ce-03c191fb9c04.png)

---
Wallet provider section still missing from mobile

---
These look like they need to go up a font weight - 600 for both
![image](https://user-images.githubusercontent.com/39496225/98107288-99db1200-1e9a-11eb-80a3-0466070f7c0f.png)

---

Can you increase the size of the icons to 24px 24px

Link to design if needed (at the bottom of the linked frame) https://www.figma.com/file/42QoTLQcbObu7esoTPjWRv/Augur-Bet-Design-System?node-id=283%3A145
![image](https://user-images.githubusercontent.com/39496225/98107648-166df080-1e9b-11eb-90e4-765e55132dc7.png)

---
Button should be slightly smaller at 36px and space above and below 10px
![image](https://user-images.githubusercontent.com/39496225/98107960-87150d00-1e9b-11eb-8b8d-bcd2333acea5.png)

matt-bullock:
@bconfortin this looks good. Couple of requests:

- can you turn off background scrolling when the menu is open. If it's a big task, leave it for a later date.

- Can you centre the dropdown arrow. I've just updated the designs as it looks odd being so high

![image](https://user-images.githubusercontent.com/39496225/99519838-a8452580-2992-11eb-8e73-40f65cc17e52.png)

yanelbottini:
Hi @bconfortin 
I'm still seeing the arrow on the Account section top-aligned. Should be centered based on previous comment:

**Dev:**
[...]

PR #9631, opened 2020-11-03
Adjusted styles for mobile account dropdown

bconfortin:
#9578

### DIFF
--- a/packages/augur-ui/src/modules/common/icons.tsx
+++ b/packages/augur-ui/src/modules/common/icons.tsx
@@ -592,8 +592,8 @@ export const AddIcon = (
 
 export const BettingAddIcon = (
   <svg viewBox="0 0 12 12" fill="none">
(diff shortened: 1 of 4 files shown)


packages/augur-ui/src/modules/common/icons.tsx
```diff
@@ -590,6 +590,14 @@ export const AddIcon = (
   </svg>
 );
 
+export const BettingAddIcon = (
+  <svg viewBox="0 0 12 12" fill="none">
+    <path d="M6 3V9" stroke-width="2" stroke-miterlimit="10" stroke-linecap="square"/>
```
JohnDanz: these should be `strokeWidth` `strokeMiterlimit` and `strokeLinecap` respectively, same with the next line.

JohnDanz (changes requested):

JohnDanz (approved):

Issue #9574, opened 2020-10-29
Mobile: if opening rules modal and scrolling the background will scroll and the betslip opens


matt-bullock:
Looks odd

![image](https://user-images.githubusercontent.com/39496225/97594329-dc1bd380-1a02-11eb-816d-f7885079ab41.png)

matt-bullock:
Rules button isn't working on mobile

PR #9635, opened 2020-11-03
update scrolling issues

phoebemirman:
https://github.com/AugurProject/augur/issues/9574
https://github.com/AugurProject/augur/issues/9574

### DIFF
--- a/packages/augur-ui/src/modules/market/components/market-view/betting-market-view.tsx
+++ b/packages/augur-ui/src/modules/market/components/market-view/betting-market-view.tsx
@@ -88,6 +88,11 @@ const BettingMarketView = () => {
   }, [showCopied]);
 
   useEffect(() => {
+    document.getElementById("mainContent")?.scrollTo(0, 0);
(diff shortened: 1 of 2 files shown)


packages/augur-ui/src/modules/market/components/market-view/betting-market-view.tsx
```diff
@@ -87,6 +87,11 @@ const BettingMarketView = () => {
     return () => (isMounted = false);
   }, [showCopied]);
 
+  useEffect(() => {
+    document.getElementById("mainContent")?.scrollTo(0, 0);
```
JohnDanz, line 91: so for this one, I think we need to find a solution because we are removing functionality. the idea here is we make sure to scroll back to the top of the page when we first hit market page cuz sometimes you can end up clicking into a market and being scrolled down the page etc.

JohnDanz (changes requested):

JohnDanz (approved):

Issue #9609, opened 2020-11-02
Strong Typing - Market Charts


JohnDanz:
update typing within the Market Charts folder.

PR #9645, opened 2020-11-04
Market charts typing

JohnDanz:
https://github.com/AugurProject/augur/issues/9609


packages/augur-ui/src/modules/market-charts/components/price-history/price-history.tsx
```diff
@@ -230,7 +237,7 @@ const getOptions = ({
     line: {
       dataGrouping: {
         forced: true,
-        units: [['minute', [1]]],
+        // units: [['minute', [1]]],
```
JohnDanz, line 240: purposefully left as a comment. might end up reverting it in a future pr.

phoebemirman (approved):

PR #9706, opened 2020-11-12
add raw positions tab, add getter to return user's raw positions base…

bthaile:
…d on share token logs

### DIFF
--- a/packages/augur-sdk/src/Augur.ts
+++ b/packages/augur-sdk/src/Augur.ts
@@ -482,6 +482,12 @@ export class Augur<TProvider extends Provider = Provider> {
     return this.bindTo(Users.getUserTradingPositions)(params);
   };
 
+  getRawUserTradingPositions = (
+    params: Parameters<typeof Users.getRawUserTradingPositions>[2]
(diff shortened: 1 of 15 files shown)


enolan:
I'm running this from a git checkout, and it's only working for one market out of four I have positions in. In the one market, the raw positions tab is populated and looks correct, in the others it's blank. The one it works in I've made trades only using the core Augur UI, but two of the markets it doesn't work in I've also only used the core UI. One of the non-working ones is the Catnip market, where I've made transfers in and out of Foundry.

Hope this is helpful.

pgebheim:
> I'm running this from a git checkout, and it's only working for one market out of four I have positions in. In the one market, the raw positions tab is populated and looks correct, in the others it's blank. The one it works in I've made trades only using the core Augur UI, but two of the markets it doesn't work in I've also only used the core UI. One of the non-working ones is the Catnip market, where I've made transfers in and out of Foundry.
> 
> Hope this is helpful.

Thanks for this information. I'm gonna merge this as is and do a build and put it out for more feedback and then see if we can fix those issues.

PR #9722, opened 2020-11-16
Add pagination to graphql provider

justinbarry:


pgebheim:
Is it much more difficult to do the more efficient mechanism of using block/log index to page vs using skip?

Issue #9671, opened 2020-11-06
Mobile: my active markets clean up and errors


matt-bullock:
- missing 24hr percentage change at top

- remove custom market tags

- Duplicate of percentage change in each market

- allowing 3 lines of text would be better for market title

- check spacing above and below market title text - looks too tight

https://www.figma.com/file/bZh7tOiLm6Ls9Hl4RL6qMC5X/Augur-Redesign?node-id=15745%3A34

![image](https://user-images.githubusercontent.com/39496225/98366671-a6936d80-2034-11eb-8a24-5e999f6841c9.png)

matt-bullock:
![image](https://user-images.githubusercontent.com/39496225/99996195-b395b800-2dbb-11eb-95d8-cdb3d9fb1418.png)
why does this part look so different to the design?

PR #9790, opened 2020-11-25
Fixed issues with product switcher, account summary and account dropdown

bconfortin:
#9578 see last comment - locked the scroll on mobile when account dropdown is open
#9671 see last comment - fixed the borders of the 24h thing, kept search on top

Made a few improvements to account switcher (no ticket, just a talk with JD):
- Trading button shouldn't appear when the user is in Trading UI, same for Sportsbook.
- Dropdown now closes when a link that is not an inner dropdown is clicked.

### DIFF
--- a/packages/augur-ui/src/modules/app/components/inner-nav/base-inner-nav-pure.tsx
+++ b/packages/augur-ui/src/modules/app/components/inner-nav/base-inner-nav-pure.tsx
@@ -50,9 +50,9 @@ const BaseInnerNavPure = () => {
   const sortProps = {
     setFilterSortState: updates => setFilterSortState({ ...filterSortState, ...updates }),
   };
-  const showMainMenu = mobileMenuState >= MOBILE_MENU_STATES.FIRSTMENU_OPEN;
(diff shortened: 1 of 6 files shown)


packages/augur-ui/src/modules/app/components/inner-nav/base-inner-nav-pure.tsx
```diff
@@ -50,9 +50,9 @@ const BaseInnerNavPure = () => {
   const sortProps = {
     setFilterSortState: updates => setFilterSortState({ ...filterSortState, ...updates }),
   };
-  const showMainMenu = mobileMenuState >= MOBILE_MENU_STATES.FIRSTMENU_OPEN;
+  const showMainMenu = mobileMenuState >= MOBILE_MENU_STATES.FIRSTMENU_OPEN && mobileMenuState <= MOBILE_MENU_STATES.SUBMENU_OPEN;
```
bconfortin, line 53: I'm confused regarding the logic of "showMainMenu". When I tried to set a mobile menu state, it was opening what I wanted + filters [???] so I just limited that condition to the amount of options that we had and added a new mobileMenuState.

JohnDanz (approved):

PR #9835, opened 2020-12-10
switched AMM to use long-short instead of invalid protection

robert-e-davidson3:


justinbarry (approved):
lgtm

bthaile (approved):
lgtm, let's get it out there so we can test it.

PR #9861, opened 2020-12-17
Updates for market page

JohnDanz:
data shapes and fitting stuff into market page. added dumb heartbeat (15s interval for getMarkets)


bthaile (approved):
lgtm

PR #9863, opened 2020-12-17
Unwrap eth when claiming proceeds via WethWrapperForAMMExchange

justinbarry:


packages/augur-core/src/contracts/para/WethWrapperForAMMExchange.sol
```diff
@@ -137,4 +137,37 @@ contract WethWrapperForAMMExchange {
         }
         shareToken.unsafeBatchTransferFrom(_from, _to, _tokenIds, _amounts);
     }
+
+    function claimTradingProceeds(IMarket _market, address payable _shareHolder, bytes32 _fingerprint) external {
+        require(shareToken.isApprovedForAll(_shareHolder, address(this)), "Weth wrapper contract must be approved to do share token transfers on behalf of the share holder.");
```
robert-e-davidson3, line 142: I've been letting the share token contract throw an error when lacking approval. It's a bit harder to track though, so either way.

packages/augur-core/src/contracts/para/WethWrapperForAMMExchange.sol
```diff
@@ -137,4 +137,37 @@ contract WethWrapperForAMMExchange {
...
+
+        uint256 _INVALID = shareToken.getTokenId(_market, 0);
```
robert-e-davidson3, line 145: I'd also add a comment explaining why you're duplicating code. It's easy to miss the crucial detail and end up trying to refactor, breaking some functionality.

PR #9888, opened 2020-12-23
market-links

JohnDanz:
added market links and updated some types/warnings, added currency label for top of market page.


bthaile (approved):
lgtm

PR #9892, opened 2020-12-24
Multicall user balances

bthaile:
added tweaked ethereum-multicall, I think more changes will be needed to support web3-react web3 provider that simplified has access to.
get user's LP token balances and market share balances.


packages/augur-simplified/src/modules/App.tsx
```diff
@@ -25,9 +27,13 @@ const AppBody = () => {
   const {
     sidebarType,
     showTradingForm,
-    actions: { setIsMobile, updateGraphData, updateProcessed },
+    processed,
+    loginAccount,
+    actions: { setIsMobile, updateGraphData, updateProcessed, updateUserBalances },
   } = useAppStatusStore();
 
+  const { library } = useActiveWeb3React()
```
andrewdamelio: You don't need to use `useActiveWeb3React()`, all that is stored inside the main app context
```
const { chainId, library, active } = loginAccount;
```

bthaile: removed refer to useActiveWeb3React in app.tsx

andrewdamelio (approved):

PR #9906, opened 2020-12-29
settings-updates

JohnDanz:
added input for slippage, added selected styles, added some generic testing for inputs on the slippage. probably needs a bit more work, but we need to circle back for the tooltips so will finish up then after some testing


bthaile (approved):
lgtm

PR #9958, opened 2021-01-05
Fix weth wrapper graph event sender issue

justinbarry:
* Add events to WethWrapperForAmmExchange
* Add WethWrapperForAmmExchange events to graph


bthaile (approved):
Lgtm

Issue #9913, opened 2020-12-30
Toast Setup


JohnDanz:
<img width="662" alt="Screen Shot 2020-12-30 at 9 50 41 AM" src="https://user-images.githubusercontent.com/10524630/103364920-84dac380-4a84-11eb-8409-0ee97ed40adc.png">

We need to setup a toast system in the state (or derived from state) and add it to the activity area of the portfolio. Add any missing components as well.

https://www.figma.com/file/6y4nvjfeVZwzwKcoXB0neq/Simplified-UI?node-id=320%3A56291

PR #9963, opened 2021-01-05
Toast setup

JohnDanz:
mainly focused on the visual here. in this mock case, we have 2 toasts, a confirmed and a failed. they will appear in order of when we received them based on `lastUpdated` then display one at a time. after 3 seconds, it's automatically dismissed.


JohnDanz:
https://github.com/AugurProject/augur/issues/9913

bthaile (approved):
lgtm

Issue #9974, opened 2021-01-06
Some USDT add liquidity graph transactions have contract instead of actual sender


bthaile:
Graph results for AMM `0x07f45f1f4d1e1f3644f1c5ae2c88acce5fbd2589`
![Screen Shot 2021-01-06 at 06 57 10](https://user-images.githubusercontent.com/3970376/103770930-6d538b80-4fec-11eb-9b8a-f99f02dd6e51.png)

<https://kovan.etherscan.io/tx/0x038f87d3f1f633442fccfec674eed3149bc316f72de201c8aecae186cec1bf22>

![Screen Shot 2021-01-06 at 06 57 38](https://user-images.githubusercontent.com/3970376/103770971-84927900-4fec-11eb-8d81-a64aed2ec214.png)

this is USDT, it's weird because there are a few USDT add liquidity transactions the sender is correct but this one the sender is the ` 0x78977faabb7df506d2699b864ebcf081102a36c5` contract in the graph transaction.

PR #9975, opened 2021-01-06
stab at calc init value and current value of user's LP tokens

bthaile:
values are off, graph transactions `sender` value is inconsistent. So I can't calculate initial value consistently. create ticket <https://github.com/AugurProject/augur/issues/9974> for middleware

image


JohnDanz (approved):

bthaile:
currently blocked <https://github.com/AugurProject/augur/issues/9974>

Issue #9981, opened 2021-01-06
Remove dropdown from orderticket 


matt-bullock:
Should just look like this

https://www.figma.com/file/6y4nvjfeVZwzwKcoXB0neq/Simplified-UI?node-id=319%3A4398

![image](https://user-images.githubusercontent.com/39496225/103789236-111a5700-5040-11eb-9dba-e145b4eadf8c.png)

Issue #9980, opened 2021-01-06
Remove invalid from graph. Make sure Yes is first checkbox followed by no


matt-bullock:
same as this:

![image](https://user-images.githubusercontent.com/39496225/103785583-abc46700-503b-11eb-824d-85e38efd71e1.png)

matt-bullock:
Still seeing no first

PR #9984, opened 2021-01-06
More chart fun

JohnDanz:
updated invalid hidden and reversed order of YES/NO checkboxes for chart. Updated order ticket to remove currency dropdown in favor of a currency label. label is now actively using market's amm data to determine what label to show or if it should show at all in the case of an amm-less market.

https://github.com/AugurProject/augur/issues/9980
https://github.com/AugurProject/augur/issues/9981


bthaile (approved):
lgtm

PR #10029, opened 2021-01-08
add in error states for trading for, add error message to button

bthaile:
disable button if error state
show error message in button


packages/augur-simplified/src/modules/market/trading-form.tsx
```diff
@@ -305,26 +311,27 @@ const TradingForm = ({
...
   const [tradeEstimates, setTradeEstimates] = useState<TradeEstimates>({})
+  const [tradeError, setTradeError] = useState<string>(null);
   const ammCash = amm?.cash;
   const outcomes = amm?.ammOutcomes || [];
   const userCashBalance = amm?.cash?.name ? balances[amm?.cash?.name]?.balance : "0";
 
   useEffect(() => {
     let isMounted = true;
     const getEstimate = async () => {
+      setTradeError(null);
```
phoebemirman: if you move this line, the button won't flash. can you only put it after you know there are no errors?

phoebemirman (changes requested):
see request

phoebemirman (approved):

PR #10042, opened 2021-01-11
<ApprovalButton /> updates

andrewdamelio:
- Added localStorage for transaction store so we can restore pending states
- Updated approval calls to update and remove (when confirmed) transactions items in our transactions store
- Adde Add/Remove Liquidity action hooks

- TODO: doing ENTER/EXIT trade now
- TODO: approval store still needs to be finished for better UX


bthaile (approved):
lgtm, let's merge up and start testing

Issue #10020, opened 2021-01-08
Update to icons and category tags on market cards and market page


matt-bullock:
Have icons based on secondary category and text for the sub category, we now use the icon of the primary category and and the text tag is also the primary category.

So in the example below the icon is US Politics and the tag is also US Politics (the text is just there to reinforce the icon)

![image](https://user-images.githubusercontent.com/39496225/104026062-bf033e00-51c5-11eb-91fa-9f9584159bf8.png)

Below is the link to all the new icons we are using. These will all require updating.

https://www.figma.com/file/6y4nvjfeVZwzwKcoXB0neq/Simplified-UI?node-id=319%3A3202

![image](https://user-images.githubusercontent.com/39496225/104025589-2b317200-51c5-11eb-8ebc-591103478750.png)

PR #10087, opened 2021-01-12
Oh, the ole Cat in the Wall, ay?

JohnDanz:
https://github.com/AugurProject/augur/issues/10020
added more category icons and updated category icon map.


bthaile (approved):
Lgtm

PR #10091, opened 2021-01-12
add Enter/Exit for Approval button

andrewdamelio:
-  add Enter/Exit trade flows
- fixes some bugs where user state wasn't getting cleared


bthaile (approved):
lgtm

Issue #10024, opened 2021-01-08
Filter/Sort Accuracy


JohnDanz:
It appears like the filter/sorting options on the markets list page are currently hooked up and filtering/sorting market cards. However we are unsure if it's accurate/using the right values. It's very possible it's using the wrong volume value to sort or something like that. This ticket is basically a "make sure this is actually working like someone would expect it to"

`Total Volume`: AmmExchange.volumeTotalUSD
`24 hr Volume`: AmmExchange.volume24hrTotalUSD
`Liquidity`: AmmExchange.liquidityUSD
`Ending Soon` : AmmExchange.market.endTimestamp

Issue #9973, opened 2021-01-06
On mobile show when filters have been activated by displaying the number of active filters in the button


matt-bullock:
Or users may forget they have filters set

![image](https://user-images.githubusercontent.com/39496225/103768532-f71e4b80-5022-11eb-9bea-b3edab7d2b0f.png)

PR #10101, opened 2021-01-12
Fix sort and add filter number to mobile

phoebemirman:
https://github.com/AugurProject/augur/issues/10024
https://github.com/AugurProject/augur/issues/9973


bthaile (approved):
lgtm

Issue #10033, opened 2021-01-08
pagination sits 48px from bottom and 48px from cards


matt-bullock:
![image](https://user-images.githubusercontent.com/39496225/104059255-503dd900-51f5-11eb-89de-604857bdd792.png)

matt-bullock:
@phoebemirman Can you do the same for mobile

PR #10124, opened 2021-01-13
add right padding to pagination, update markets layout, add tooltip to exit trading form, fix tooltip on trading form

phoebemirman:
https://github.com/AugurProject/augur/issues/10033


bthaile (approved):
lgtm

Issue #10119, opened 2021-01-13
show min balance of Invalid and No shares


bthaile:
for user positions balances show min of Invalid shares and no shares.

bthaile:
we'll need to figure out how to test this, prob trade No shares in staging and verify that user's balance is the lesser of No and Invalid shares.

PR #10145, opened 2021-01-14
Normalize no invalid shares

bthaile:
<https://github.com/AugurProject/augur/issues/10119>

take min of "No Shares" and "Invalid shares" balances when populating user's No position


bconfortin (approved):
lgtm

andrewdamelio (approved):

PR #10164, opened 2021-01-15
fix issue where wrong user balance was used to estimate remove liquid…

bthaile:
…ity that cause errors in console

fix console error when user opens modal to remove liquidity, user's USDC balance was sent to estimate remove liquidity.


bconfortin (approved):
lgtm

PR #10210, opened 2021-01-18
Check for Cash approval on enter/exit trade also

### LINKED ISSUE #10209: user not prompted to approve enter position for USDC
I had a user fail to enter position (buy trade) user didn't get approval button. 
user account did get approve button to add liquidity, once it went through my account could enter position (buy)

andrewdamelio:
fixes #10209

Added cash approval to trade form


bthaile (approved):
approving
I'd like the conditional logic to not be in the trading form. 
only ever one approval is needed to trade, what needs to be approved differs depending on cash or buy or sell. This crazy logic should not be in the trading form, It needed to be abstracted out. imo

PR #10219, opened 2021-01-19
cash now impacts LP token calculation

### LINKED ISSUE #10177: Lp token amounts more than suppose to be when user adds liquidity
User is adding USDC to existing market that has trades and users have added and removed liquidity. 
user is getting more LP tokens than expected. 

example:
user adds 1000 USDC and ends up with 1000.83 LP tokens

robert-e-davidson3:
AMM LP token calculation was ignoring cash. Treating cash like sets to be minted, adding the numticks-adjusted cash held by the AMM to long and short sides.

closes #10177


bthaile (approved):
lets get it in and start testing. looks like a new deployment is needed.

PR #10227, opened 2021-01-19
trading-form-refactoring

JohnDanz:
refactored trading-form to reduce calls to get breakdown, eliminate unneeded state.

### DIFF
--- a/packages/augur-simplified/src/modules/market/trading-form.tsx
+++ b/packages/augur-simplified/src/modules/market/trading-form.tsx
@@ -418,19 +418,20 @@ const TradingForm = ({
   const [selectedOutcome, setSelectedOutcome] = useState(
     initialSelectedOutcome
   );
-  const [breakdown, setBreakdown] = useState<EstimateTradeResult>(null);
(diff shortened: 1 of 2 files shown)


bthaile (approved):
lgtm

packages/augur-simplified/src/modules/market/trading-form.tsx
```diff
@@ -381,90 +418,78 @@ const TradingForm = ({
...
+      if (isMounted) setBreakdown(breakdown);
     };
-    if (orderType && selectedOutcome.id && amount && Number(amount) > 0) {
+
+    if (amount && Number(amount) > 0) {
       getEstimate();
-    } else {
-      orderType === BUY ? setBreakdown(undefined) : setBreakdown(undefined);
+    } else if (breakdown !== null) {
+      setBreakdown(null);
```
bthaile: setting Breakdown to null, means insufficient liquidity. for default set to undefined.

JohnDanz: reverting but we should move away from a pattern that relies on null vs undefined vs some object all meaning different things.

PR #10230, opened 2021-01-19
more clear setting of breakdown rather than using null and undefined …

bthaile:
…to define states


bthaile:
going in a different direction

PR #10236, opened 2021-01-19
fix-breakdown-regression

JohnDanz:
removed more unneeded state and corrected a regression, added the word shares to the rate. need to further update this in the future


bthaile (approved):
Lgtm

PR #10256, opened 2021-01-21
use para deploys config and not hardcoded cash addresses in UI

bthaile:
use para deploys config and not hardcoded cash addresses in UI


robert-e-davidson3 (approved):
monsta mash

pgebheim (approved):
:potato:

PR #10271, opened 2021-01-21
use new method that takes multiple markets to claim either single or …

bthaile:
…multiple market winnings per para share token


JohnDanz (approved):
lgtm

Issue #10269, opened 2021-01-21
Order ticket numbers (average price, est shares, max profits, est fees) should become dark text when user enters an amount


matt-bullock:
![image](https://user-images.githubusercontent.com/39496225/105370343-c97cf900-5c03-11eb-992c-bdd4a18df52f.png)

Issue #10226, opened 2021-01-19
display decimal consistently


bthaile:
comprehensive list
- price: 2 decimals
- USDC values 2 decimals
- ETH values 4 decimals
- Share values 4 decimals

JohnDanz:
should update formatPrice/formatShares/formatEth/formatUSDC in simplified to default to these -- then make sure whenever we display one of the above values we always put it through one of these on render.

Ideally we just come up with a formatCash that takes in (value, cashName, options) where value and cashName are required and options are optional override for the default options (like ever other formatNumber)

JohnDanz:
going to close this, any individual examples we find going forward should just get it's own ticket.

PR #10282, opened 2021-01-21
Numbers n ticket

JohnDanz:
https://github.com/AugurProject/augur/issues/10269
https://github.com/AugurProject/augur/issues/10226 -- most of this leaving the ticket open so it can keep being updated if we missed a spot.


bthaile (approved):
reviewed

phoebemirman (approved):

Issue #10191, opened 2021-01-18
Update remove liquidity 


matt-bullock:
Updating this so the user can only remove all. Remove section 'How much do you want to remove' and disabled button in initial state will say 'Remove all Liquidity' rather than 'Enter Amount'. And Change 'Remove' button to 'Remove all Liquidity'

https://www.figma.com/file/6y4nvjfeVZwzwKcoXB0neq/Simplified-UI?node-id=1208%3A2271

![image](https://user-images.githubusercontent.com/39496225/104930718-d3092580-59a5-11eb-9cfb-c7ba83b7d8cc.png)

Chwy5:
This is what I see when trying to remove liquidity...think it is missing something (It also will not allow me to enter an amount)

![image](https://user-images.githubusercontent.com/41301809/105417738-077d1b80-5c0a-11eb-9389-e776c80f825a.png)

PR #10295, opened 2021-01-21
fix remove liquidity

phoebemirman:
https://github.com/AugurProject/augur/issues/10191#issuecomment-764968543


bthaile (approved):
Lgtm, only remove liq will get user’s max balance

PR #10314, opened 2021-01-22
Claim winnings UI

JohnDanz:
added claim winnings transactions to transactions. refactored a little bit. cleaned up the style for claim individual markets.


bthaile (approved):
lgtm

Issue #10335, opened 2021-01-23
don't abbr number in trading form


bthaile:
Numbers are super important in trading form. Don't abbr numbers when having to do with trading.

PR #10344, opened 2021-01-25
remove abbr

JohnDanz:
dropped abbr from shares. was accidently defaulting to it for shares specifically

https://github.com/AugurProject/augur/issues/10335


bthaile (approved):
lgtm

PR #10348, opened 2021-01-25
move cash conversion after cash compare to shares

bthaile:
estimation for exit position give same numbers as graph tx

Screen Shot 2021-01-25 at 13 41 27

image


robert-e-davidson3 (approved):
definitely has a bug but this unblocks testing. need a new PR soon for the real fix

Issue #10338, opened 2021-01-25
Buying shares with ETH, user doesn't receive shares but sends ETH


bthaile:
User tries to buy shares with ETH but user doesn't receive any shares. Here is a tx.

https://kovan.etherscan.io/tx/0x6bdd33ffe248cc0d55d50625f3c8ea5ade937bd5facdf5856dcdaa03cc3213f9

justinbarry:
@bthaile Looks like the balances are there on contract.

Chwy5:
fixed - closing

PR #10353, opened 2021-01-25
multicall collection needs a unique key, needed to add share token address.

bthaile:
contract calls were getting overridden because key wasn't unique in collection

<https://github.com/AugurProject/augur/issues/10338>


bconfortin (approved):
Lgtm

Issue #10357, opened 2021-01-26
currency icons look broken on mobile


matt-bullock:
![image](https://user-images.githubusercontent.com/39496225/105846661-c8b9dd80-5fdc-11eb-8f97-441390a24efa.png)

![image](https://user-images.githubusercontent.com/39496225/105846704-d7a09000-5fdc-11eb-941f-448d0470f1dc.png)

![image](https://user-images.githubusercontent.com/39496225/105846719-de2f0780-5fdc-11eb-96e7-8313ee43d258.png)

Issue #10356, opened 2021-01-26
Augur logo gets partially hidden when liquidity modals are open


matt-bullock:
![image](https://user-images.githubusercontent.com/39496225/105845314-dff7cb80-5fda-11eb-854d-adcdd46420f2.png)

PR #10365, opened 2021-01-26
fix svgs

phoebemirman:
https://github.com/AugurProject/augur/issues/10356
https://github.com/AugurProject/augur/issues/10357


bconfortin (approved):
Lgtm

PR #10367, opened 2021-01-26
Enable kovan for portis/fortmatic

andrewdamelio:
- enable kovan for portis/fortmatic
- remove extra portis button in Connect button


bconfortin (approved):
Tested and was able to trade with Portis and Fortmatic. Nice work. LGTM

Issue #10324, opened 2021-01-22
Add pending, success and failed states to account button


matt-bullock:
Design for buttons: https://www.figma.com/file/6y4nvjfeVZwzwKcoXB0neq/Simplified-UI?node-id=309%3A2852

If you place an order on the first market of this prototype you'll get an idea of how it works: https://www.figma.com/proto/6y4nvjfeVZwzwKcoXB0neq/Simplified-UI?node-id=93%3A9478&viewport=64%2C436%2C0.08909575641155243&scaling=min-zoom

![image](https://user-images.githubusercontent.com/39496225/105525626-a3c62180-5ce1-11eb-9fcc-ecb38cdc6f80.png)
![image](https://user-images.githubusercontent.com/39496225/105525948-10412080-5ce2-11eb-943a-6d09598d1045.png)

PR #10374, opened 2021-01-26
Add pending account button

andrewdamelio:
relates #10324 

- Added Pending status to ConnectAccount button
- Updated localStorage as followed:
   - `lastUser` to keep track of lastLogin for auto login (MM)
   - `WALLET-ADDRESS` to track the key'd address transactions, approvals, and userSettings. (approvals and userSettings in next PR)
- Check status of PENDING transactions on each block.

### DIFF
--- a/packages/augur-simplified/src/assets/styles/_colors.less
+++ b/packages/augur-simplified/src/assets/styles/_colors.less
@@ -23,6 +23,7 @@
   --color-primary-text-30: fade(#D7DDE0, 30%);
   --color-primary-text-40: fade(#D7DDE0, 40%);
   --color-primary-text-50: fade(#D7DDE0, 50%);
+  --color-primary-text-80: fade(#D7DDE0, 70%);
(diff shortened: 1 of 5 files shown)


packages/augur-simplified/src/assets/styles/_colors.less
```diff
@@ -23,6 +23,7 @@
   --color-primary-text-30: fade(#D7DDE0, 30%);
   --color-primary-text-40: fade(#D7DDE0, 40%);
   --color-primary-text-50: fade(#D7DDE0, 50%);
+  --color-primary-text-80: fade(#D7DDE0, 70%);
```
JohnDanz: Is this even used? if so, rename it to --color-primary-text-70 (already have an 80) if it's not in use then just remove it.

JohnDanz (changes requested):

JohnDanz (approved):

PR #10394, opened 2021-01-27
State optimization

JohnDanz:
removed updateProcessed/GraphData/Blocknumber in favor of updateGraphHeartbeat which will handle all 3 updates at once.

### DIFF
--- a/packages/augur-simplified/src/modules/stores/app-status-hooks.ts
+++ b/packages/augur-simplified/src/modules/stores/app-status-hooks.ts
@@ -47,6 +47,7 @@ const {
   BLOCKNUMBER,
   MODAL,
   IS_LOGGED,
+  SHOW_TRADING_FORM,
 } = APP_STATE_KEYS;
 
 const isAsync = (obj) => {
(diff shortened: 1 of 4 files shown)


packages/augur-simplified/src/modules/stores/app-status-hooks.ts
```diff
@@ -234,6 +225,12 @@ export function AppStatusReducer(state, action) {
           activity,
         };
       }
+      updatedState[GRAPH_DATA] = {
+        markets: arrayToKeyedObject(action[GRAPH_DATA].markets),
+        past: arrayToKeyedObject(action[GRAPH_DATA].past),
+        paraShareTokens: arrayToKeyedObject(action[GRAPH_DATA].paraShareTokens),
+      };
+      updatedState[BLOCKNUMBER] = action.blocknumber;
```
bthaile, line 233: Is blocknumber on state needed for future? I don't see it used anywhere

bthaile (approved):
works and I don't see any issues.

Issue #9986, opened 2021-01-06
Add search bar


matt-bullock:
### Desktop design: https://www.figma.com/file/6y4nvjfeVZwzwKcoXB0neq/Simplified-UI?node-id=842%3A122

![image](https://user-images.githubusercontent.com/39496225/103796598-b684f900-5047-11eb-8881-0e7c7ee1a71f.png)

![Kapture 2021-01-06 at 16 31 56](https://user-images.githubusercontent.com/39496225/103794650-44abb000-5045-11eb-96b8-9dd37e64b392.gif)

Clickable prototype: https://www.figma.com/proto/6y4nvjfeVZwzwKcoXB0neq/Simplified-UI?node-id=818%3A3488&viewport=672%2C534%2C0.28564971685409546&scaling=min-zoom

**Animation guide:**
animation-timing-function: ease-in-out;
animation-duration: 100ms;

---

### Mobile Design

https://www.figma.com/file/6y4nvjfeVZwzwKcoXB0neq/Simplified-UI?node-id=842%3A2431

![image](https://user-images.githubusercontent.com/39496225/103797009-2f845080-5048-11eb-8143-41a49272b1d4.png)

---

Add empty states

![image](https://user-images.githubusercontent.com/39496225/104056089-2fbf5000-51f0-11eb-984f-5fcc1a00c188.png)

PR #10397, opened 2021-01-27
add search to market view

phoebemirman:
https://github.com/AugurProject/augur/issues/9986


bthaile (approved):
works well and is fast

Issue #10439, opened 2021-01-30
total value is off, should be USD value of transaction


bthaile:
maybe graph data changed, have to look into it. 

![Screen Shot 2021-01-30 at 13 31 13](https://user-images.githubusercontent.com/3970376/106366141-83105400-62ff-11eb-9e90-ff54eb963f1a.png)

PR #10440, opened 2021-01-30
calc usd value of transaction

bthaile:
<https://github.com/AugurProject/augur/issues/10439>

image


packages/augur-simplified/src/utils/process-data.ts
```diff
@@ -416,6 +420,7 @@ const shapeAddLiquidityTransactions = (
   transactions: GraphAddLiquidity[],
   cash: Cash
 ): AmmTransaction[] => {
+  console.log('cash.usdPrice', cash.usdPrice)
```
bconfortin: Did you leave this here intentionally?

bconfortin (approved):
lgtm

Issue #10442, opened 2021-02-01
When user doesn't have a custom value show the placeholder


bthaile:
![Screen Shot 2021-02-01 at 08 19 46](https://user-images.githubusercontent.com/3970376/106470600-4c088280-6466-11eb-9a09-3c1fab1eb8c4.png)

JohnDanz:
closing -- Matt just had me change this to say Custom last week.

PR #10444, opened 2021-02-01
custom slippage label

JohnDanz:
https://github.com/AugurProject/augur/issues/10442
corrected slippage to use the custom label when not selected


packages/augur-simplified/src/modules/common/top-nav.tsx
```diff
@@ -133,7 +145,7 @@ export const SettingsButton = () => {
                         }
                       }
                     }}
-                    placeholder={slippage}
+                    placeholder="custom"
                     max="100"
```
bthaile: max slippage is 1000%, need to update value in custom input. can be done in later PR or this one.

bthaile (approved):
lgtm

Issue #10447, opened 2021-02-01
Need message for when the graph can't fetch data


matt-bullock:
Can use the same banner as when connected to the wrong network.

Copy: 
Unable to retrieve market data

![image](https://user-images.githubusercontent.com/39496225/106489047-a3cdda80-64b4-11eb-89d9-19ff0719c93c.png)

PR #10448, opened 2021-02-01
graph-down-banner

JohnDanz:
https://github.com/AugurProject/augur/issues/10447

updated network mismatch banner to also handle no graph data, updated updateHeartbeat to include errors. updated state to include errors to processed.


bthaile (approved):
lgtm

Issue #10449, opened 2021-02-01
custom slippage value not retained after refresh


bthaile:
- cleared local storage to get to a know state. 
- verified 2% is the default
- refreshed and noticed that no slippage value was retained. 
- typed in 75 for custom slippage value
- refreshed
- no slippage value was retained in Settings again. See below.

![2021-02-01 11 31 21](https://user-images.githubusercontent.com/3970376/106495373-36ed1d00-6481-11eb-978c-e71100b5cc3a.gif)

PR #10450, opened 2021-02-01
add-slippage-check

JohnDanz:
added a custom slippage check to update custom value if its loaded from storage and is custom
https://github.com/AugurProject/augur/issues/10449


bthaile (approved):
lgtm

Issue #10451, opened 2021-02-01
update page layout for two tone pages


phoebemirman:
![image](https://user-images.githubusercontent.com/6775839/106499467-7ca3e880-6475-11eb-9032-860c4df1d189.png)

the lighter color on the right should bleed/continue in that space

PR #10465, opened 2021-02-01
update-two-tone-display

JohnDanz:
https://github.com/AugurProject/augur/issues/10451
added bleeding effect for twotone, aka white background should now bleed all the way to the right of the screen on wide screen displays on two tone pages.


phoebemirman (changes requested):
Screen Shot 2021-02-01 at 1 42 30 PM
i don't think this top bar is supposed to be transparent

phoebemirman (approved):

PR #10468, opened 2021-02-01
fix enterposition. better amm bames

robert-e-davidson3:


pgebheim (approved):
This looks good to me but would love @bthaile eyes on the Middleware.

robert-e-davidson3:
One small additiona change: use outputLong for all middleware swap methods not just one.

bthaile (approved):
Lets do this! Merge in so we can test

PR #10498, opened 2021-02-02
Add balancer contracts and kovan factory address

### LINKED ISSUE #10422: Add Support for Invalid Pricing via Balancer Pool
## Goal
Provide a pool that can price invalid in any market against collateral while remaining relatively capital efficient.

## Structure

- [x] During AMM Creation, also create a corresponding Balancer pool.
- [x] During AMM Creation, make the Invalid outcome ERC-20 using the ERC-20 wrapper.
- [ ] Also create Yes/No wrappers at the same time. (ref #10618 )
- [x] We need to name these token. UI should pass base symbol name and prepend yes/no/invalid on contract (Need #10558)
- This pool should have a weight of 90% for the collateral, and 10% for the Invalid Wrapper.
- When a LP adds liquidity.
  - [x] Split collateral into two chunks, 90% of the collateral gets deposited in the AMM pool as normal. 10% of it is used to fill the balancer pool.
  - [x] Of the 10%, the pool should be populated as follows: 50% of the collateral is used as the deposit into the invalid pool. The other 50% is used to mint complete sets. The invalid shares from those complete sets are transferred into the Balancer pool as liquidity. The remaining shares are returned to the LP as well as the BPT from adding to the pool.
- When LP removes liquidity
  - [x] Should also remove from the Balancer pool via BPT.
  - This means it will need approval to spend BPT.
  - TBD: Can BPT support `permit` ? If so, we should support the signature style approvals for this.

I think all of the above percentages and stuff should be calculated on the front and and passed in, so the contracts just deal with the various mechanics of moving the tokens around and creating the right things.

- [ ] Is invalid flag on the graph. Should be true when the pool is at a certain price. (ref #10691 )

justinbarry:
close #10422


packages/augur-core/src/contracts/para/AMMFactory.sol
```diff
@@ -7,15 +7,18 @@ import 'ROOT/para/ParaShareToken.sol';
...
     using SafeMathUint256 for uint256;
 
     IAMMExchange internal proxyToClone;
+    BPool internal bPool;
 
     event AMMCreated(IAMMExchange amm, IMarket market, ParaShareToken shareToken, uint256 fee);
 
-    constructor(address _proxyToClone) public {
+    constructor(address _proxyToClone, BPool _bPool) public {
+        bPool = _bPool;
```
bthaile, line 21: Is there a pool per AMM? or does factory create pools through this `bPool`?

Issue #10509, opened 2021-02-03
graph erros but no banner shown to user


bthaile:
![image](https://user-images.githubusercontent.com/3970376/106682300-d0065b80-6587-11eb-9a3e-544897144503.png)

Don't see a banner telling user data could not be retrieved. 
![image](https://user-images.githubusercontent.com/3970376/106682311-d5fc3c80-6587-11eb-8f52-993cd3eb388f.png)

PR #10512, opened 2021-02-03
remove extra try catches

JohnDanz:
removed extra try/catch blocks and updated searchMarkets to handle graph errors
https://github.com/AugurProject/augur/issues/10509


bthaile (approved):
lgtm

PR #10518, opened 2021-02-03
workaround for MM switch issue, add logout

andrewdamelio:
- Fix issue where provider is incorrect on MM account switching (temp solution to unbrick MM account switch) 
- Add  *Disconnect* for MM on AccountDetails


bthaile (approved):
lgtm

JohnDanz (approved):

Issue #10504, opened 2021-02-02
multi-winnings calc is incorrect


bthaile:
user's winnings calc is incorrect

![image](https://user-images.githubusercontent.com/3970376/106657298-768b3600-6561-11eb-98e9-399de8d98e7b.png)

bthaile:
USDC is calc correctly. Have to test ETH winnings

![image](https://user-images.githubusercontent.com/3970376/106782361-b01a7a80-660f-11eb-846a-ac882e5d95e8.png)

Chwy5:
fixed - closing

PR #10526, opened 2021-02-03
calculate winnings based on share balances instead of using usd value

bthaile:
claim winnings was in USD value for ETH and USDC, needed to be in cash.

Screen Shot 2021-02-03 at 13 24 40

<https://github.com/AugurProject/augur/issues/10504>


andrewdamelio (approved):
Lgtm

Issue #10522, opened 2021-02-03
Add invalid flag to portfolio positions and liquidity 


matt-bullock:
Same size as the currency icon. 8px to the left of the currency icon.

Use the same tooltip that's used for invalid on the market cards

![image](https://user-images.githubusercontent.com/39496225/106792791-e8917700-6656-11eb-9a1a-d0e3f3b09393.png)

Issue #10521, opened 2021-02-03
Add invalid tag to market pages when market is within the invalid filter


matt-bullock:
https://www.figma.com/file/6y4nvjfeVZwzwKcoXB0neq/Simplified-UI?node-id=1762%3A102965

![image](https://user-images.githubusercontent.com/39496225/106791963-e549bb80-6655-11eb-8451-a2fb3ef8a936.png)

Issue #10519, opened 2021-02-03
Add invalid tag + tooltip to market card (when market is classed as invalid)


matt-bullock:
https://www.figma.com/file/6y4nvjfeVZwzwKcoXB0neq/Simplified-UI?node-id=309%3A90

![image](https://user-images.githubusercontent.com/39496225/106790872-75870100-6654-11eb-809e-6f23f83bb1b8.png)

PR #10535, opened 2021-02-04
Update invalid flags

JohnDanz:
left it commented for now, once we have a flag on the markets for trending invalid we can update the common component to check for it on the market object it takes in and then simply uncomment the commented out components and we should be good to go.
https://github.com/AugurProject/augur/issues/10519
https://github.com/AugurProject/augur/issues/10521
https://github.com/AugurProject/augur/issues/10522


bthaile (approved):
lgtm

PR #10543, opened 2021-02-04
Setup show invalid markets

JohnDanz:
updated filtering of markets in markets list to handle showInvalidMarkets flag from settings and update when flag changes


bthaile (approved):
lgtm

Issue #10541, opened 2021-02-04
Make market card clickable anywhere under top bar if liquidity.


matt-bullock:
Currently only the market title takes you to the market page. It would be good if the entire card is clickable

PR #10545, opened 2021-02-04
card-click-surface

JohnDanz:
https://github.com/AugurProject/augur/issues/10541

updated market cards to make most of the area clickable. basically if it has an AMM then the whole thing is clickable except the top bar area.


bthaile (approved):
lgtm

Issue #10534, opened 2021-02-04
When creating the initial AMM the user will need to give a custom token name (ui part added, waiting on middleware)


matt-bullock:
We need an input which displays the token names below with 'y', 'n' and 'i' as default at the start of the names.

The token names will then be displayed on the confirmation screen

![image](https://user-images.githubusercontent.com/39496225/106911395-63fc3280-6702-11eb-97aa-1fe2569b7254.png)

Link to entire flow: https://www.figma.com/file/6y4nvjfeVZwzwKcoXB0neq/Simplified-UI?node-id=555%3A3269

![image](https://user-images.githubusercontent.com/39496225/106911355-5646ad00-6702-11eb-9d6d-914d14a4d24e.png)

matt-bullock:
@phoebemirman Can you reduce the font size in the input to 16px Bold

Chwy5:
added - closing

PR #10557, opened 2021-02-05
add field for custom tokens

phoebemirman:
https://github.com/AugurProject/augur/issues/10534


bthaile (approved):
lgtm. only shows custom token on amm create.

Issue #10565, opened 2021-02-06
Remove Negatie sign from 0 pnl balance


Chwy5:
![image](https://user-images.githubusercontent.com/41301809/107121696-e64f3880-6861-11eb-944b-7c73c819e1f9.png)

PR #10578, opened 2021-02-08
update-formatDai-and-movementlabel

JohnDanz:
updated the logic for formatDai denomination to only add a - sign if the formatted value would be less than 0.00. updated movement label logic to only add a + sign if the value is greater than or equal to 0.01 instead of greater than 0 to prevent display of `+$0.00` for example if the value was a positive fraction of a penny.

https://github.com/AugurProject/augur/issues/10565


bthaile (approved):
lgtm

Issue #10529, opened 2021-02-03
Claim trading proceeds data processing issue. 


bthaile:
![image](https://user-images.githubusercontent.com/3970376/106810915-34cac000-6633-11eb-8c39-69af6bd40f22.png)

PR #10595, opened 2021-02-09
fix-double-claim-activity

JohnDanz:
updated process data to correctly only sort out claims if they match the currency of the exchange.

https://github.com/AugurProject/augur/issues/10529


bthaile (approved):
lgtm

PR #10596, opened 2021-02-09
add Kovan for walletConnect

andrewdamelio:


bconfortin (approved):
Already tested and it's working beautifully. Good job :D

PR #10599, opened 2021-02-09
fix-buffer-overrun

JohnDanz:
fixed a bad token pass on claim all for checking eth markets approval prevents a buffer override app blowing up style error


bthaile (approved):
lgtm

PR #10600, opened 2021-02-09
fix-bad-input

JohnDanz:
removed margin on amount input with currency dropdown as it was unintentionally shortening the input display.


bthaile (approved):
lgtm

PR #10604, opened 2021-02-10
update filenames in docs

jeremyschlatter:
Renamed source to src, to reflect the actual directory name.

Deleted some references to files that no longer exist.


pgebheim (approved):
Nice!

Docs were generated from a previous branch. Good find.

There's some other issues with this too that we need to look into, like sometimes it redirects to augur.net.

Issue #10641, opened 2021-02-12
User needs approval to claim ETH winnings


bthaile:
On market page user needs to be approved to claim ETH winnings. This behavior has changed slightly all claims need to go through middleware even single ETH AMM market claims. 

need to use same approve than claim button behavior as Portfolio page.

Chwy5:
works - closing

PR #10646, opened 2021-02-12
upgrade-position-footer

JohnDanz:
added checks for if we are approved to exit position on eth claims instead of assuming we are
https://github.com/AugurProject/augur/issues/10641


bthaile (approved):
lgtm

PR #10648, opened 2021-02-12
logged-out-trading-form-state

JohnDanz:
updated trading form to always be disabled and say Connect Wallet on the buy/sell button if user is not connected.

https://github.com/AugurProject/augur/issues/10647


bthaile (approved):
lgtm

PR #10659, opened 2021-02-12
fix provider error on wallet switch

andrewdamelio:
- fixes issue where switching between two accounts (MM / Portis) would result in the SDK still using the first wallet provider

- also updated `useCanExitCashPosition/useCanEnterCashPosition` to always check for approval -- switching from a wallet that is `APPROVED` to a wallet that is `NOT APPROVED` would always return true since we don't reset the hook state between wallet switching.


bthaile (approved):
Lgtm

bthaile (approved):
lgtm

Issue #10652, opened 2021-02-12
betslip pop up animation


phoebemirman:
https://www.figma.com/proto/6y4nvjfeVZwzwKcoXB0neq/Simplified-UI?node-id=68%3A1144&viewport=-2579%2C612%2C0.2975539267063141&scaling=scale-down

// On click
// Open overlay: "Order Ticket / Mobile / Yes No";
// Animate: Move in Bottom;
animation-timing-function: ease-in-out;
animation-duration: 200ms;

PR #10664, opened 2021-02-15
trading form animation, and other small UI updates

phoebemirman:
https://github.com/AugurProject/augur/issues/10652


bconfortin (approved):
Looks good. I don't see any hickups/performance issues.

Issue #10677, opened 2021-02-18
sort market positions on portfolio page


bthaile:
sort positions on portfolio page so that winning positions are at the top.

PR #10679, opened 2021-02-18
update-positions-order

JohnDanz:
updated portfolio view to put claimable positions first on the list of positions.
https://github.com/AugurProject/augur/issues/10677


bthaile (approved):
lgtm

PR #10683, opened 2021-02-18
add call to get gas prices

bthaile:
image


JohnDanz (approved):
in the future, we should decouple as much as possible. technically gas price isn't coming from the graph, we should probably have it split out into it's own thing.

Issue #9977, opened 2021-01-06
Animation for redrawing points when switching date ranges


matt-bullock:
Can we do something similar to this example:

https://www.indexcoop.com/dpi

Issue #9976, opened 2021-01-06
Round points on graph to improve the visual


matt-bullock:
recommend rounding by 4px

![image](https://user-images.githubusercontent.com/39496225/103778924-47050e80-5033-11eb-8b94-df737e3071da.png)

PR #10704, opened 2021-02-22
chart updates

JohnDanz:
updated chart style to be even more rounded and have less sharp edges, re-introduced animation when switching date ranges.

https://github.com/AugurProject/augur/issues/9976
https://github.com/AugurProject/augur/issues/9977


bthaile (approved):
lgtm

Issue #10573, opened 2021-02-08
negative sign should be before the dollar sign.


bthaile:
![image](https://user-images.githubusercontent.com/3970376/107247599-0ebc6b80-69f7-11eb-997a-891763c5c27a.png)

PR #10741, opened 2021-03-02
negative-before-denom

JohnDanz:
https://github.com/AugurProject/augur/issues/10573

changed formatDai to put the negative sign prior to the $ sign


bthaile (approved):
lgtm

PR #10762, opened 2021-03-04
More comps integration

JohnDanz:
removed graphData, Market Card, and marketCardStyles from simplified and import them from comps now. WindowRef was also removed in favor of comps version.


bthaile (approved):
lgtm

PR #10777, opened 2021-03-08
More simplified integration

JohnDanz:
removed utils from simplified, updated comps, updated references in simplified to use comps. removed apollo folder from comps.


bthaile (approved):
lgtm

PR #10831, opened 2021-03-10
Comps integration part4

JohnDanz:
replaced Selection.tsx in simplified.
made comps appStatus
use comp's appStatus and UserStore in migrate
removed unused files/code in simp and migrate.


phoebemirman (changes requested):
theres a bug here with transactions not being finalized in the migration app

packages/augur-migration/src/modules/stores/user-hooks.ts
```diff
@@ -1,179 +0,0 @@
...
-          (tx) => tx.hash !== action.hash
-        );
-      }
-      break;
-    }
-    case FINALIZE_TRANSACTION: {
-      updatedState[TRANSACTIONS].forEach((tx) => {
-        if (tx.hash === action.hash) {
-          tx.confirmedTime = now;
-          tx.status = TX_STATUS.CONFIRMED;
```
phoebemirman, line 116: we should add this line to the user hooks in comps to fix the bug

phoebemirman (approved):

PR #10844, opened 2021-03-11
add console log for remove liquidity

bthaile:


justinbarry (approved):
This is the greatest PR I have ever seen.

Issue #10845, opened 2021-03-12
Trading form doesn't clear out when user goes from Buy to Sell or vice versa


bthaile:
Notice that user inputs value and clicks on "Sell" and the estimation is negative.
The order ticket should clear the amount input when order type changes (Buy | Sell)

![2021-03-12 08 21 33](https://user-images.githubusercontent.com/3970376/110952581-0617c780-830c-11eb-8c58-85c8f7fe5fb2.gif)

PR #10848, opened 2021-03-12
clear amount when clicking between buy and sell

phoebemirman:
https://github.com/AugurProject/augur/issues/10845


bthaile (approved):
lgtm

Issue #10847, opened 2021-03-12
All time chart should crop into the time the market started


matt-bullock:
For example this market is a day old but the all time chart goes back months

![image](https://user-images.githubusercontent.com/39496225/110968020-4fd9cf80-8357-11eb-8a09-9f4b29b8c17a.png)

PR #10856, opened 2021-03-15
update-alltime-chart-logic

JohnDanz:
updated logic for charts to tighten up all time to be no more time back than start time - 1 tick, ticks calculated based on hours passed since market creation.

https://github.com/AugurProject/augur/issues/10847


bthaile (approved):
Lgtm

Issue #10846, opened 2021-03-12
Dont show "winning outcome" until market is finalized


bthaile:
only show winning outcome when market "isFinalized"

![image](https://user-images.githubusercontent.com/3970376/110967467-028c3c80-831c-11eb-88aa-68a233d0d528.png)

Chwy5:
fixed - closing

PR #10859, opened 2021-03-15
winning-for-finalized-only

JohnDanz:
updated winning outcome label on market view and on market cards (in comps) to only show on finalized/resolved markets.

https://github.com/AugurProject/augur/issues/10846


bthaile (approved):
Lgtm

PR #10872, opened 2021-03-16
More refactors

JohnDanz:
bum bum bum....

Big one. removes inputs and commonized up the labels outside of a few that are simp specific. made simp store, use appstatus store from comps. use comps where we weren't before. cleanup imports/minor refactors sprinkled in on simple components that wrapped a return for example instead of just returning.


phoebemirman (changes requested):
getting a crash/error in portfolio page when logged in

bthaile (approved):
lgtm

phoebemirman (approved):

Issue #10909, opened 2021-03-22
Update background border and fill colour of ETH icon as it looks too washed out


matt-bullock:
Please update border outline to 8B959E and fill to F9FAFB

![image](https://user-images.githubusercontent.com/39496225/111986540-5bbf5180-8b0e-11eb-982e-42bd1dc7579e.png)

PR #10910, opened 2021-03-22
update eth logo

JohnDanz:
updated the border color for eth logo icon to be a bit darker to contrast a bit better
https://github.com/AugurProject/augur/issues/10909


bthaile (approved):
lgtm

PR #10934, opened 2021-03-29
fix-approval-hook

JohnDanz:
fixed an oversight that allowed for the useApproval hook to use other approval states for the current approval option. now correctly calls again whenever we change inputs that may require a new check.


Chwy5 (approved):

bconfortin (approved):
Lgtm

PR #10935, opened 2021-03-29
reduce-login-updates

JohnDanz:
improve useEffect for ConnectAccountButton to be a bit more specific about when it should updateLoginAccount again


bthaile (approved):
super sweet

